PR #21, opened 2015-06-11
Do not truncate refund amount when printing debug data to the user

maaku:
Since '%d' was used, refund values are being truncated from display. This fix uses the floating point '%.16g' format flag instead, the same flag used for displaying the value.


gmaxwell:
Seems reasonable! utACK

sipa:
utACK

PR #22, opened 2015-06-12
Guard wallet code in pow.cpp with ENABLE_WALLET

pstratem:
Include missing ENABLE_WALLET guards to avoid building wallet/including bdb if --disable-wallet is specified.


jtimon:
ut ACK

PR #33, opened 2015-06-23
Add support for Merkle tree scripts (1-of-n multisig, with n up to 2^32).

sipa:
This adds support for a new standard script type, one which contains a merkle root of a tree of public keys, and verifies a signature with one of these public keys, using a merkle path to prove it belongs to the tree.

The practical effect is the ability to encode a 1-of-N multisig in Elements' scripting language, using very moderate resources (1 CHECKSIG, log2(N) SHA256, 136 + 6*log2(N) bytes scriptSig).


sipa:
Addressed @gmaxwell's comments.

sipa:
Working on a shiny and improved version which is incompatible.

PR #34, opened 2015-06-24
Support creating blinded inputs without blinded output

sipa:
Whenever some inputs are blinded, but no outputs are, forcibly add a 0-value OP_RETURN output.

Also fix a bug to not complain when a blinded change output was already added in this case.


sipa:
The 0-value OP_RETURN output could have a reduced size range proof. That isn't implemented yet.

sipa:
@gmaxwell ping

sipa:
@gmaxwell Bi-weekly ping.

PR #39, opened 2015-07-23
Update alpha-README to fix line formatting

erykwalder:


wtogami:
How is adding spaces fixing the formatting?

erykwalder:
Markdown requires some indication for when to add line breaks, which is space following the end of a line. This way step 2 begins on its own line.

sipa:
ACK

PR #48, opened 2015-08-21
Key tree signatures

sipa:
Adds:
* Generic threshold tree code, and ability to iterate over their allowed combinations and count them.
* Keytree layer that adds serializable key tree objects (which are threshold trees over public key leaves), and ability to compute their Merkle root and branches in logspace.
* Human-readable format/parse code for key trees.
* Signing/ismine code that can cope with key tree outputs.
* Unit tests for all the above.
* Wallet/walletdb code to store key trees.
* RPCs addtreesigaddress and createtreesig.


gmaxwell:
I'm not sure about this. It has tests. Is that allowed in elements? :P

sipa:
I tried to remain consistent with the existing style which includes some rudimentary tests. Perhaps these were just a leftover from an older codebase and should be removed?

src/core_read.cpp
```diff
@@ -155,3 +156,86 @@ vector<unsigned char> ParseHexUV(const UniValue& v, const string& strName)
...
+        }
+        children.push_back(KeyTreeNode());
+        if (!ParseKeyTreeNode(s, pos, children.back())) return false;
+        count++;
+    }
+}
+
+static bool ParseKeyTreeNode(const std::string &s, size_t &pos, KeyTreeNode& tree)
+{
+    while (pos < s.size() && isspace(s[pos])) pos++;
```
apoelstra: I'd like to have this line at the top of `ParseKeyTreeCall` and at the top of the `while` loop in that function. That allows to have whitespace before `(` and `,`.

src/core_read.cpp
```diff
@@ -155,3 +156,86 @@ vector<unsigned char> ParseHexUV(const UniValue& v, const string& strName)
...
+    if (s.size() >= pos + 66 && IsHex(s.substr(pos, 66))) {
+        std::vector<unsigned char> data = ParseHex(s.substr(pos, 66));
+        tree.leaf.Set(data.begin(), data.end());
+        pos += 66;
+        return tree.leaf.IsFullyValid();
+    }
+    if (s.size() >= pos + 2 && s.substr(pos, 2) == "OR") {
+        pos += 2;
+        if (!ParseKeyTreeCall(s, pos, NULL, tree.children)) return false;
+        if (tree.children.size() < 2) return false;
```
apoelstra: Using `< 2` rather than `< 1` forbids constructs such as `AND(key)` or `OR(key)` (which would both be synonyms for `key`). I take it this is deliberate?

I think we want better error reporting here, even an integer code.

sipa: The human representation is intentionally deterministic, allowing only a
single representation for every tree (using the simplest construct
available). I guess we could theoretically allow either OR(one) or
AND(one), though.

Maybe that rule is overkill, and we can allow thresholds of 1 or n, single
arguments, whitespace, lowercase commands, ...

apoelstra: I like having the rule, -if- we can report it to the user. Otherwise I think it'll result in some "mysterious" parse failures, which in the case of script-generated tree descriptions may also be intermittent.

src/thresholdtree.h
```diff
@@ -0,0 +1,195 @@
...
+ *  An iterator is also parametrized in the type of an accumulator. This
+ *  accumulator is an object kept inside the iterator, and matching leaves are
+ *  added and removed from it. When a iterator reaches a stable combination,
+ *  the accumulator can be queried to retrieve information about the matching
+ *  subset of leaves. Using an accumulator means that early checks can be
+ *  performed, and not all combinations are necessarily actually ever stored.
+ */
+template<typename Node, typename Accumulator>
+class ThresholdTreeIterator {
+    // InnerIterators are creating for iterating over the combinations of one
```
apoelstra: creating -> created

src/thresholdtree.h
```diff
@@ -0,0 +1,195 @@
...
+        }
+        // If not, move to the next combinations of children.
+        for (int childnum = iter->children.size() - 1; ; childnum--) {
+            if (iter->children[childnum].position + iter->children.size() != childnum + iter->node->Children()) {
+                iter->children[childnum].position++;
+                for (int childnum2 = childnum + 1; childnum2 < iter->children.size(); childnum2++) {
+                    iter->children[childnum2].position = iter->children[childnum2 - 1].position + 1;
+                }
+                for (int childnum2 = 0; childnum2 < iter->children.size(); childnum2++) {
+                    iter->children[childnum2].inner_iterator = BuildInnerIterator(iter->node->Child(iter->children[childnum2].position));
```
apoelstra: Can you release the old `iter->children[childnum2].inner_iterator` before making a new one?

src/keytree.cpp
```diff
@@ -0,0 +1,328 @@
...
+
+namespace {
+
+/* This implements a constant-space merkle root/branch calculator, limited to 2^63 leaves. */
+template<typename LeafSource>
+void MerkleComputation(LeafSource& source, uint256* proot, std::vector<uint256>* pbranch) {
+    if (pbranch) pbranch->clear();
+    if (!source) {
+        if (proot) *proot = uint256();
+        return;
```
apoelstra: I'd prefer this check be an assertation than "return an error code using a hash type".

src/key.cpp
```diff
@@ -144,6 +144,81 @@ bool CKey::Derive(CKey& keyChild, unsigned char ccChild[32], unsigned int nChild
...
 }
 
+bool CKey::PartialSigningNonce(const uint256& hash, std::vector<unsigned char>& pubnonceout) const {
+    if (!fValid)
+        return false;
+    secp256k1_pubkey_t pubnonce;
+    unsigned char secnonce[32];
+    LockObject(secnonce);
+    int ret = secp256k1_schnorr_generate_nonce_pair(secp256k1_context, hash.begin(), begin(), secp256k1_nonce_function_rfc6979, NULL, &pubnonce, secnonce);
+    UnlockObject(secnonce);
```
apoelstra: Should we be zeroing out `secnonce` before unlocking it?

src/thresholdtree.h
sipa: At this point all subiterators should already be released in the loop above.

src/key.cpp
sipa: UnlockObject wipes the associated memory.

src/keytree.cpp
sipa: Indeed. Do you think a method would be clearer?

src/core_read.cpp
sipa: So... introducing an enum with potential parse errors etc? Meh.

src/core_read.cpp
sipa: Whitespace is not very deterministic...

src/thresholdtree.h
sipa: Done.

src/keytree.cpp
apoelstra: I'd much prefer a method with a name like exhausted(), rather than a cast to bool.

src/thresholdtree.h
apoelstra: Yep, you're right, I see it now. Though I think a comment to this effect should be added.

src/keytree.cpp
sipa: Done, I have changed the bool operator and the increment operator into methods everywhere.

src/thresholdtree.h
sipa: Added a comment to clarify above.

src/script/script.h
```diff
@@ -558,6 +558,20 @@ class CScript : public std::vector<unsigned char>
...
+    static bool DecodeInt(opcodetype opcode, const std::vector<unsigned char>& data, int64_t* ret)
+    {
+        if (opcode == OP_0) {
+            *ret = 0;
+            return true;
+        } else if (opcode >= OP_1 && opcode <= OP_16) {
+            *ret = (int)opcode - (int)(OP_1 - 1);
+            return true;
+        } else if (data.size() < 5) {
+           *ret = CScriptNum(data, false).getint64();
```
apoelstra: Are you missing a `return true` here?

sipa:
Closing; I'd like to use a different implementation, once bitcoin/secp256k1#322 is merged.

Anarky990:
> Adds:
> 
>     * Generic threshold tree code, and ability to iterate over their allowed combinations and count them.
> 
>     * Keytree layer that adds serializable key tree objects (which are threshold trees over public key leaves), and ability to compute their Merkle root and branches in logspace.
> 
>     * Human-readable format/parse code for key trees.
> 
>     * Signing/ismine code that can cope with key tree outputs.
> 
>     * Unit tests for all the above.
> 
>     * Wallet/walletdb code to store key trees.
> 
>     * RPCs addtreesigaddress and createtreesig.

PR #49, opened 2015-08-26
Use Script-based signing for messages

sipa:


jtimon:
ut ACK.

luke-jr:
Would be nice to support HD derivation somehow to avoid signatures leaking the public key. Is this feasible?

jtimon:
This is also a good excuse to PR script/generic.hpp to Bitcoin Core...

gmaxwell:
Soft-forks make this fail open.

jtimon:
> Soft-forks make this fail open.

Can you elaborate on that? I don't think I understand.

gmaxwell:
A soft fork adds a new script feature.  I create a new address that uses it and publish that.

You accept my pubkey.

In the future someone gives you an invalid signature with my key and you verify it with old software, the signature will pass without warning.

jtimon:
Thanks. Mhmm, that seems intrinsically unsolvable...I guess you mean to solve it using versions (maybe the validation flags?) in the signed messages or something of the sort.

gmaxwell:
Yes.

sipa:
@gmaxwell We could restrict future changes to message signing to identifiable things (specifically: NOP opcodes and segwit script versions).

maaku:
Well message validation is not consensus code. It's acceptable for the signature validator to throw an exception if unrecognized soft-fork features are used. The API gets a little ugly as the result becomes a triple-valued boolean of Valid/Invalid/UpgradeRequired, but I don't see how that is avoidable. The UpgradeRequired exception is thrown if an unrecognized NOP or segwit script version is used.

jtimon:
utACK ddf13c5b48a34125681a313f61306f1b13cea39e

I like the idea of the upgrade-required-to-verify-this-scriptSig exception.
Perhaps we want just want a script version at the beginning of the script ala segwit2 (meaning the newer softfork version)?
Then we could show a warning if the verification passes (but still reject the message if it fails even with the old known rules).

sipa:
@maaku No need for an exception. Just evaluate once with SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_NOPS and once without. If the result differs, return UpgradeRequired.

jtimon:
ack @sipa's outlined final design.

jtimon:
Perhaps this can be replaced by a PR to https://github.com/ElementsProject/elements/commits/elements-0.13.1 or just be closed?

jtimon:

PR #50, opened 2015-08-26
Disable proof checking for all blocks on load

sipa:
We still verify the headers of the last -checkblocks blocks.


gmaxwell:
I believe this should go upstream as well-- though obviously it'll be much less of a benefit there. :)

apoelstra:
Do we lose some checksumming benefit by doing this?

ACK anyway, certainly we shouldn't be verifying multisignatures as a checksum :)

PR #52, opened 2015-09-11
Add `report_error` function for peer errors which can be overridden

apoelstra:
I'm not sure exactly which errors should go through this function rather than being `print`ed directly; the idea is that these messages are ones that indicate network/peer problems that should be addressed immediately.

### DIFF
--- a/contrib/fedpeg/blocksign.py
+++ b/contrib/fedpeg/blocksign.py
@@ -52,6 +52,9 @@ class WatchPeerController(RotatingConsensus):
 		self.round_local_block_hex = ""
 		return
 
+	def report_error(self, msg):
+		settings.report_error(msg)
+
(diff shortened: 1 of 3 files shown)


maaku:
Maybe have a command line parameter that specifies a shell command to execute?

apoelstra:
@maaku I think rather than a shell command I should move the function into `constants.py` with the other user-configurable stuff.

jtimon:
Does this still make sense in the rebased https://github.com/ElementsProject/elements/commits/elements-0.13.1 ? do we still want this for alpha?

instagibbs:
This doesn't make any sense for 0.13.1 since it's not using the python
federation scripts.

jtimon:
Merge or close?

jtimon:

PR #55, opened 2015-09-30
Add commandline flag to set genesis block script destination

### LINKED ISSUE #54: Regtest needs ability to change the blocksigning scriptpubkey from the command line

apoelstra:
Fixes #54

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -80,7 +80,7 @@ static const Checkpoints::CCheckpointData dataRegtest = {
 
 class CMainParams : public CChainParams {
 public:
-    CMainParams() {
+    CMainParams(CScript scriptDestination) {
         networkID = CBaseChainParams::MAIN;
         strNetworkID = "main";
         /** 
@@ -127,7 +127,9 @@ public:
         genesis.hashMerkleRoot = genesis.BuildMerkleTree();
         genesis.nVersion = 1;
         genesis.nTime    = 1231006505;
-        CScript scriptDestination(CScript() << OP_5 << ParseHex("027d5d62861df77fc9a37dbe901a579d686d1423be5f56d6fc50bb9de3480871d1") << ParseHex("03b41ea6ba73b94c901fdd43e782aaf70016cc124b72a086e77f6e9f4f942ca9bb") << ParseHex("02be643c3350bade7c96f6f28d1750af2ef507bc1f08dd38f82749214ab90d9037") << ParseHex("021df31471281d4478df85bfce08a10aab82601dca949a79950f8ddf7002bd915a") << ParseHex("0320ea4fcf77b63e89094e681a5bd50355900bf961c10c9c82876cb3238979c0ed") << ParseHex("021c4c92c8380659eb567b497b936b274424662909e1ffebc603672ed8433f4aa1") << ParseHex("027841250cfadc06c603da8bc58f6cd91e62f369826c8718eb6bd114601dd0c5ac") << OP_7 << OP_CHECKMULTISIG);
+        if (scriptDestination.empty()) {
+            scriptDestination = CScript() << OP_5 << ParseHex("027d5d62861df77fc9a37dbe901a579d686d1423be5f56d6fc50bb9de3480871d1") << ParseHex("03b41ea6ba73b94c901fdd43e782aaf70016cc124b72a086e77f6e9f4f942ca9bb") << ParseHex("02be643c3350bade7c96f6f28d1750af2ef507bc1f08dd38f82749214ab90d9037") << ParseHex("021df31471281d4478df85bfce08a10aab82601dca949a79950f8ddf7002bd915a") << ParseHex("0320ea4fcf77b63e89094e681a5bd50355900bf961c10c9c82876cb3238979c0ed") << ParseHex("021c4c92c8380659eb567b497b936b274424662909e1ffebc603672ed8433f4aa1") << ParseHex("027841250cfadc06c603da8bc58f6cd91e62f369826c8718eb6bd114601dd0c5ac") << OP_7 << OP_CHECKMULTISIG;
+        }
         genesis.proof = CProof(scriptDestination, CScript()); // genesis block gets a PoW pass
 
         hashGenesisBlock = genesis.GetHash();
@@ -163,14 +165,13 @@ public:
         return data;
     }
 };
-static CMainParams mainParams;
 
 /**
  * Testnet (v3)
  */
 class CTestNetParams : public CMainParams {
 public:
-    CTestNetParams() {
+    CTestNetParams(CScript scriptDestination) : CMainParams(scriptDestination) {
         networkID = CBaseChainParams::TESTNET;
         strNetworkID = "test";
         pchMessageStart[0] = 0xee;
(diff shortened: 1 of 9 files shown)


jtimon:
Why not call ScriptDestinationFromCommandLine etc directly in CRegTestParams instead of in SelectParamsFromCommandLine() ?
Or maybe CRegTestParams's constructor could take a CScript parameter that defaults to OP_TRUE. 
Certainly ChangeGenesisScriptDestination doesn't need to be a public method, all this can be a just a static function in chainparams.cpp
In fact, this PR is buggy since ChangeGenesisScriptDestination doesn't update hashGenesisBlock nor mapCheckpointsTestnet[0].

apoelstra:
@jtimon For "why did I structure the functions this way" I was trying to copy as best I could what is done for the `-testnet`/`-regtest` parameters. As for giving a parameter to the `CRegTestParams` that's tricky because the regtest paramaters are a static variable, and IIRC in C++ we have no control over what order constructors for static variables are run in.

As for the bugs, you're right, thanks.

jtimon:
re CRegTestParams::CRegTestParams() parameter: I was assuming some of https://github.com/bitcoin/bitcoin/pull/6382 had already been merged, but some of those changes were rejected for alpha so never mind.

Anyway, the solution is still simple: just put all the new code inside CRegTestParams::CRegTestParams(), before the hash of the genesis block is calculated.

See https://github.com/bitcoin/bitcoin/pull/6382 which also builds the genesis block dynamically from a command line parameter for the sizetest ```std::numeric_limits<uint64_t>::max()``` different chains.

src/chainparams.cpp
```diff
@@ -249,7 +249,7 @@ class CRegTestParams : public CTestNetParams {
         txNew.vout[1].scriptPubKey = CScript() << OP_TRUE;
         genesis.vtx[0] = CTransaction(txNew);
 
-        CScript scriptDestination(CScript() << OP_TRUE);
+        CScript scriptDestination(CScript() << OP_2 << ParseHex("023303dedc51b9d227b17c9fb4710f96b844e1ccdc2c776e1b7274bd4e246b6202") << ParseHex("03024c3b4f830854d6d26d6e34d92aff4c703bf57e85cd42abe328d928e01d4286") << ParseHex("03fe4e8c8d99b9dcbb529a87d54f606bae0149a34018325547fa0c2239e038a1c9") << OP_3 << OP_CHECKMULTISIG);
```
apoelstra: Oops, this should not have made it into the PR.

apoelstra:
Thanks for https://github.com/bitcoin/bitcoin/pull/6382/files @jtimon, there are a lot of good ideas in there. I think I'll copy your factory pattern.

jtimon:
Awesome! Maybe you want to utACK one of the following open little steps in bitcoin/bitcoin?

 Chainparams: Translations: DRY: options and error strings #6235
 Don't check the genesis block #6597

apoelstra:
Rebased using a factory pattern as in @jtimon's code. Need to add some unit tests before this is usable.

Like bitcoin/6382, there is an argument that this is only useful for QA and may not make sense to be merged. But in this case, I think it's reasonable to expect users to play with their own network with their own functionaries, and this makes this much easier.

apoelstra:
Added unit tests that confirms the genesis hash is computed correctly (for testnet and regtest, that the output of RPC `getblockhash 0` matches the result of directly passes `CScript`s into `SetParams`; for mainnet and unittest I assumed the result was correct and added a check for it to catch regressions).

apoelstra:
@jtimon Do you have further thoughts on this PR? I know you have overlapping PRs in Core.

@gmaxwell You commented on https://github.com/bitcoin/bitcoin/pull/6907 that you feel making this a factory is a bad idea. Keeping them static would mean moving the constructor into some sort of `init` function, have the constructor just call this, and have `SelectParams` call that rather than calling the factory. Would you prefer this?

In any case I'd like to move forward with this PR.

src/chainparams.h
```diff
@@ -80,6 +80,14 @@ class CChainParams
     const std::vector<unsigned char>& Base58Prefix(Base58Type type) const { return base58Prefixes[type]; }
     const std::vector<CAddress>& FixedSeeds() const { return vFixedSeeds; }
     virtual const Checkpoints::CCheckpointData& Checkpoints() const = 0;
+    /** Change the script in the genesis hash */
+    void ChangeGenesisScriptDestination(CScript s) { genesis.proof = CProof(s, CScript()); }
```
jtimon: Unused method.

src/chainparams.cpp
```diff
@@ -80,7 +80,7 @@ static const Checkpoints::CCheckpointData dataRegtest = {
 
 class CMainParams : public CChainParams {
 public:
-    CMainParams() {
+    CMainParams(CScript scriptDestination) {
```
jtimon: Is this ever used for MAIN ?

src/chainparams.cpp
```diff
@@ -308,16 +309,14 @@ class CUnitTestParams : public CMainParams, public CModifiableParams {
     virtual void setAllowMinDifficultyBlocks(bool afAllowMinDifficultyBlocks) {  fAllowMinDifficultyBlocks=afAllowMinDifficultyBlocks; }
     virtual void setSkipProofOfWorkCheck(bool afSkipProofOfWorkCheck) { fSkipProofOfWorkCheck = afSkipProofOfWorkCheck; }
 };
-static CUnitTestParams unitTestParams;
 
-
-static CChainParams *pCurrentParams = 0;
+static CChainParams *pCurrentParams;
 
 CModifiableParams *ModifiableParams()
```
jtimon: Is this function still being used anywhere?

jtimon:
If I understood it correctly, @gmaxwell 's nit was that I was making globalChainParams and globalSwitchingChainParams extern instead of static and it's already solved (but it would be nice if he can clarify whether his nit has been solved or not in https://github.com/bitcoin/bitcoin/pull/6907 ).

As said, I don't like the script being passed to SeelectParams(), the factory, nor the constructors and I would prefer ScriptDestinationFromCommandLine to be called directly within the constructors, but just a preference.
I added a couple of minor nits though.

The overlapping PR/commits in Core are contained in https://github.com/jtimon/bitcoin/tree/alpha-base-0.12 which I will maintain rebased until 0.12 is forked and is what I would l would like @TheBlueMatt to use as a base when rebasing.

src/chainparams.cpp
apoelstra: Is MAIN ever used period? I think this should be there for symmetry's sake in any case.

src/chainparams.cpp
apoelstra: Apparently not. I'll remove it.

apoelstra:
Addressed @jtimon's nits

src/chainparams.h
apoelstra: Removed

src/chainparams.cpp
jtimon: Yes, in all unit tests, but fair enough.

jtimon:
I'm going to merge.

src/chainparams.h
```diff
@@ -139,13 +145,15 @@ const CChainParams &Params();
 /** Return parameters for the given network. */
 CChainParams &Params(CBaseChainParams::Network network);
```
jtimon: Wait a minute...where is this function defined now?

apoelstra: chainparams.cpp:315

jtimon: No, I mean the parametrized version.

apoelstra: Oh, hmm, this is not defined anywhere anymore (nor is it called anywhere, of course, else we would have compilation errors).

jtimon: mhmm I would like to eventually remove it from core too.I guess when rebasing the remaining uses can be replaced with the factory i elements. It's just something I expected to happen in bitcoin core first.

jtimon: Of course another option is to maintain it in elements until it removed in core.

PR #57, opened 2015-10-23
fixes comment in script_(in)valid.json to match content.

joshrabinowitz:
fix comment to describe tests correctly


gwillen:
My understanding after talking to @joshrabinowitz on IRC is that this was fixed upstream between 0.10 and 0.11 so we should probably just catch this when we rebase onto master.

sipa:
Indeed, this is fixed in Bitcoin. Going to close this, and wait for Alpha to get rebased to the 0.12 upstream version.

Issue #59, opened 2015-11-08
send-to-sidechain Aborted (core dumped)


Christewart:
Hi everyone, I'm trying to follow the [alpha-README.md](https://github.com/ElementsProject/elements/blob/alpha/alpha-README.md). I'm on step 6 trying to execute this line

>chris@chris-870Z5E-880Z5E-680Z5E:~/dev/elements$ ./contrib/sidechain-manipulation.py send-to-sidechain 2N9N5MUVcieJysWJqEUa3CXdPnGQARexHXA 1

and I am getting the following error: 

chris@chris-870Z5E-880Z5E-680Z5E:~/dev/elements$ ./contrib/sidechain-manipulation.py send-to-sidechain 2N9N5MUVcieJysWJqEUa3CXdPnGQARexHXA 1

[libsecp256k1] illegal argument: *outputlen >= ((flags & SECP256K1_FLAGS_BIT_COMPRESSION) ? 33 : 65) 

Aborted (core dumped)

==================================================
<open file '/home/chris/dev/elements/contrib/../../contracthashtool/contracthashtool -t -g -r 55210269992fb441ae56968e5b77d46a3e53b69f136444ae65a94041fc937bdb28d93321021df31471281d4478df85bfce08a10aab82601dca949a79950f8ddf7002bd915a2102174c82021492c2c6dfcbfa4187d10d38bed06afb7fdcd72c880179fddd641ea121033f96e43d72c33327b6a4631ccaa6ea07f0b106c88b9dc71c9000bb6044d5e88a210313d8748790f2a86fb524579b46ce3c68fedd58d2a738716249a9f7d5458a15c221030b632eeb079eb83648886122a04c7bf6d98ab5dfb94cf353ee3e9382a4c2fab02102fb54a7fcaa73c307cfd70f3fa66a2e4247a71858ca731396343ad30c7c4009ce57ae -d 2N9N5MUVcieJysWJqEUa3CXdPnGQARexHXA', mode 'r' at 0x7ff3cd570ed0>
==================================================
['Using testnet!', '']
Traceback (most recent call last):
  File "./contrib/sidechain-manipulation.py", line 115, in <module>
    full_contract = cht_read.split("\n")[1 + is_testnet][26:]
IndexError: list index out of range

I've supplemented `sidechains-manipulations.py` to add a few extra print statements to help debug. It now looks like this 
```python
104         elif sys.argv[1] == "send-to-sidechain":
105                 if len(sys.argv) != 4:
[...]

PR #60, opened 2015-11-09
Add CL flag support to sidechain manipulation script

droark:
- Add support for command line arguments to the sidechain manipulation script. argparse requires Python 2.7 or 3.2+.
- Add support for unlocking mainchain wallet when running send-to-sidechain.
- Add various bits of help text, some of which may not be quite right.

### DIFF
--- a/contrib/sidechain-manipulation.py
+++ b/contrib/sidechain-manipulation.py
@@ -1,9 +1,53 @@
 #!/usr/bin/env python2
 
-import sys, os, traceback
+import sys, os, traceback, argparse
 sys.path.append(os.path.join(os.path.dirname(os.path.abspath(__file__)), "../../python-bitcoinrpc"))
 from bitcoinrpc.authproxy import AuthServiceProxy, JSONRPCException
 from decimal import *
+
+# Command line arguments
+parser = argparse.ArgumentParser(description = 'Script that accommodates the ' \
+    'movement of coins in and out of sidechains. Optional arguments must ' \
+    'come after the name of the action being performed and any related ' \
+    'arguments.')
+subparsers = parser.add_subparsers(help = 'Sidechain manipulation commands',
+                                   dest = 'command')
+
+parser_generate_1of1_MS = subparsers.add_parser('generate-one-of-one-multisig')
+parser_generate_1of1_MS.add_argument('wallet', help = 'Determines if ' \
+    'sidechain or mainchain wallet will generate an address', \
+    choices = ('sidechain-wallet', 'mainchain-wallet'))
+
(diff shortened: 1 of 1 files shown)


droark:
This is a more thorough (and tested) replacement for PR #44. I ended up replacing the sys.argv usage with argparse. This way, the code adds a more useful help system alongside efficient argument parsing. The one and only time that CL flags right now are to unlock Bitcoin wallets when using send-to-sidechain. I discovered that the code, when Armory is running, will use an Armory wallet. That's how I tested this.

Note that, for now, the script won't work due to libsecp256k1 API breakage. See issue #59 for details regarding where to rewind libsecp256k1 & contracthashtool in order to test this patch.

Thanks.

droark:
Paging @gwillen. Am I on the right path? Did I go too far with the changes?

gwillen:
No way, you didn't go too far at all, this is awesome. Let me test it out and then figure out what the process is for us to merge it. Thanks!

droark:
Thanks, and sorry about the extra commits that just popped up! I added a commit and tried to squash it. My squash mojo just isn't right today, unfortunately. I'll clean it up one way or another.

gwillen:
No problem. I think what you want is 'git rebase -i foo' where foo is the _parent_ of the first commit you want to squash; then replace 'pick' with 'squash' on all lines but the first: http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

And then 'git push -f [something appropriate]' to update the pullreq. (Which will nuke all inline comments on the previous commits unfortunately, but in this case there aren't any.)

droark:
Thanks! That did the trick. I had basically done that. I didn't force the initial push, though, so something went sideways. Oh well. All's well now.

gwillen:
Other than the comment above about the behavior in case of failure -- which I think should emulate the old behavior as much as possible if the issue is not a locked wallet -- I think this is ready to be merged.

droark:
Oh no, it's totally fine! I'm happy to make any and all changes as needed. I'll look at everything and update in a little bit.

droark:
@gwillen - I believe I've addressed your feedback. Forgot to run a sanity check but I'll do that right now. Anything else you'd like for me to do? Should I squash the new commit?

gwillen:
Looks good, thanks! Go ahead and squash now that I've looked at it.

droark:
*squash*

gwillen:
Ok, after a bit of a battle with the commit-signing script, it is done! Thanks for your patience and thanks for your contribution!

droark:
Thanks! I'll keep an eye out for any more contributions I can make.

PR #61, opened 2015-11-17
Give a nice error for "make bitcoind", instead of a misleading one.

gwillen:
Right now, running 'make src/bitcoind' in alpha gives a very misleading error message about include files. What's actually happening is, there's no rule for 'bitcoind' in the Makefile, but because there's a source file named bitcoind.cpp, make is inventing a default rule, which doesn't work.

So let's make an overriding rule that displays a helpful error. (I did this for the three binaries I have run into this building; we can add more if there are more I forgot about.)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -396,6 +396,9 @@ clean-local:
 	rm -f leveldb/*/*.gcno leveldb/helpers/memenv/*.gcno
 	-rm -f config.h
 
+bitcoind bitcoin-cli bitcoin-tx:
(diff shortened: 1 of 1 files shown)


jtimon:
utACK

gwillen:
Thanks @jtimon! I don't know what the protocol is at this point -- I did test this, can I merge it now on my own judgement or should I wait for someone else to do it?

sipa:
utACK

jtimon:
I'm not sure what the protocol is, but you got two utACK, you tested and this is quite trivial, so I'm just going to merge it.

gwillen:
Thanks!

PR #64, opened 2015-11-23
Fix GUI build (broken by PR #55)

apoelstra:
I broke the GUI build with #55. This is a fix.

### DIFF
--- a/src/qt/paymentserver.cpp
+++ b/src/qt/paymentserver.cpp
@@ -211,13 +211,23 @@ void PaymentServer::ipcParseCommandLine(int argc, char* argv[])
             {
                 CBitcoinAddress address(r.address.toStdString());
 
-                if (address.IsValid(Params(CBaseChainParams::MAIN)))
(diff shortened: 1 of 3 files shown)


jtimon:
I think it will be easier for rebases to just reestablish the function until is removed from Bitcoin Core (I really plan to do that, but not before #6907).

src/qt/paymentserver.cpp
```diff
@@ -211,13 +211,23 @@ void PaymentServer::ipcParseCommandLine(int argc, char* argv[])
             {
                 CBitcoinAddress address(r.address.toStdString());
 
-                if (address.IsValid(Params(CBaseChainParams::MAIN)))
```
jtimon: This is, I think, the long term solution for paymentserver and chainparams: https://github.com/jtimon/bitcoin/commit/ae575131833790e464cc9d7b17651a882dfae1c0

jtimon:
Closed in favor of #65

PR #65, opened 2015-11-25
Chainparams: fix the factory the #6907 way

jtimon:
An alternative to #64.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -13,6 +13,7 @@
 #include <assert.h>
 
 #include <boost/assign/list_of.hpp>
+#include <boost/scoped_ptr.hpp>
 
 using namespace std;
 using namespace boost::assign;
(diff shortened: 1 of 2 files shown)


src/chainparams.cpp
```diff
@@ -334,9 +335,15 @@ CChainParams* CChainParams::Factory(CBaseChainParams::Network network, CScript s
     }
 }
 
+CChainParams &Params(CBaseChainParams::Network network)
+{
+    globalSwitchingChainParams.reset(CChainParams::Factory(network, CScript()));
+    return *globalSwitchingChainParams;
+}
+
```
apoelstra: I considered this, and decided it felt very weird for `Params()` to be `const` while `Params(network)` has global side-effects.

I think my PR is functionally identical.

apoelstra: Ah, but if your goal is easier rebases, I think that should take priority. Esp. since `Params()` will go away eventually.

Thoughts?

jtimon: Both can return a const ref though.
I mean, Params(network) is documented as deprecated here and bip70 is not implemented in the most ideal way with respect to chainparams. But those are things that should be fixed on bitcoin core and alpha-0.12/beta-0.12, not in alpha-0-10. Is anybody using the payment server with alpha anyway?
I really think just leaving Params(network) there until it's removed in bitcoin core is simpler for rebases. Maybe you or @TheBlueMatt disagree.

apoelstra:
Nah, it's fine. Agreed that my concerns should be fixed on bitcoin, not alpha.

PR #72, opened 2016-01-13
Relay everything and dont bother with priority

pstratem:


sipa:
I don't think this is safe for a public network.

pstratem:
@sipa it's not

PR #74, opened 2016-02-02
Use HMAC derivation for blinding keys

sipa:
Use a per-wallet secret to HMAC-SHA256 deriving blinding keys, rather than using a single static one.

No support for importing keys with alternate derivation keys yet.

Needs testing.


sipa:
Also included now:
* Bugfix where unblinded addresses were reported in transaction RPCs.
* Add RPCs `dumpblindingkey` and `importblindingkey` that operate on-address blinding key (not the full wallet master key)

jonasnick:
ACK when getaddressesbyaccount is fixed such that it includes blinding keys.
Tested with wallet.py and walletbackup.py (#77).

sipa:
Updated.

src/wallet.h
```diff
@@ -930,16 +944,38 @@ class CWalletTx : public CMerkleTx
...
+                res = UnblindOutput(blinding_key, vout[nOut], vAmountsOut[nOut], vBlindingFactors[nOut]);
+            }
+            if (!res && (blinding_key = pwallet->GetBlindingKey(&vout[nOut].scriptPubKey)).IsValid()) {
+                // For outputs using derived blinding.
+                res = UnblindOutput(blinding_key, vout[nOut], vAmountsOut[nOut], vBlindingFactors[nOut]);
+                if (res) {
+                    vBlindingKeys[nOut] = blinding_key.GetPubKey();
+                }
+            }
+            if (!res && (blinding_key = pwallet->GetBlindingKey(NULL)).IsValid()) {
```
instagibbs: We're having issues with none of these three cases firing with either unblinded or static blinded outputs. We get:

alphad: base58.cpp:240: CBitcoinAddress& CBitcoinAddress::AddBlindingKey(const CPubKey&): Assertion `pubkey.size() == 33' failed.

sipa: Can you create a backtrace?

instagibbs: http://pastebin.com/V8fjAUky  

It doesn't happen when the list is going to be empty.

sipa: I pushed a fix for the non-blinded case.

instagibbs:
https://github.com/instagibbs/elements/commit/ede50230faa14b73dc03f999f2a2376bb5b01147

My attempt at fixing sending, etc.

wtogami:
@sipa Please review @instagibbs patch and add to your PR branch if you like it?

sipa:
@warren Cherry picked.

PR #76, opened 2016-02-03
Improve README, Add Documentation

martindale:
The Elements repository is to be the starting point for many developers hoping to work on sidechain technology.  This proposal adds easily navigable links and messaging that guides those developers to further study.


sipa:
ACK

PR #77, opened 2016-02-05
Re-enable rpc tests wallet.py and walletbackup.py

jonasnick:
Depends on #74


instagibbs:
We should probably forward port this to `elements-0.13.1`.

jtimon:

PR #78, opened 2016-02-08
Multithreaded range proof checks

sipa:


src/main.cpp
```diff
@@ -992,6 +994,153 @@ bool CheckTransaction(const CTransaction& tx, CValidationState &state)
     return true;
 }
 
+extern secp256k1_context* secp256k1_bitcoin_verify_context;
```
jtimon: Nit: would like this in consensus/* rather than in main.cpp despite depneding on CCoinsViewCache/coins.

jtimon:
fast-read ack

jtimon:
Apart from my general dislike on adding more code to main.cpp (making things harder to read, rebase, blame, etc), it would be also nice that VerifyAmounts took a CValidationState as parameter to return different error strings in different cases.

maaku:
Merged in #143 .

instagibbs:
Included in `elements-0.13.1`, closing

PR #85, opened 2016-02-16
Deal better with unknown amounts in balances

sipa:


sipa:
Tested. This not only fixes incorrect account getbalances, but also the *amounts* listed in `listtransactions` for perfectly ordinary usage.

PR #86, opened 2016-02-26
Import settings from constants.py into sidechain-manipulation.py

TomMcCabe:
Avoids redundancies with changing the same settings in two separate files.

### DIFF
--- a/alpha-README.md
+++ b/alpha-README.md
@@ -46,7 +46,7 @@ for notes on dependencies that must be installed before beginning.
 ```bash
   git clone https://github.com/jgarzik/python-bitcoinrpc
 ```
-5\. Edit sidechain-manipulation.py (replace `user:pass` with your RPC username and password):
+5\. Edit contrib/fedpeg/constants.py (replace `user:pass` with your RPC username and password):
 ```python
   # VARIOUS SETTINGS...
   sidechain_url = "http://user:pass@127.0.0.1:4241"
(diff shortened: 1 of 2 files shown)


martindale:
ACK.

gwillen:
Oops, I was getting ready to merge this, but actually "secondScriptPubKey", removed from sidechain-manipulation.py is missing from constants.py. @TomMcCabe, can you fix, and double-check that nothing else is missing?

And can you please edit alpha-README.md to refer to editing config.py instead of sidechain-manipulation.py, to match the code change (and any other documentation you're aware of)?

Thanks!

(I think the reason this was not done this way to start with, was that 'fedpeg' was supposed to be a specialization of the general concept of sidechains, and so it doesn't make sense for this script to depend on something in 'fedpeg'. However, I think the removal of redundancy is more valuable than preserving that currently-theoretical separation, and we can always refactor it later.)

TomMcCabe:
@gwillen: Fixed as requested and squashed commits.

gwillen:
Awesome, thanks and sorry for the delay. Taking a look now.

gwillen:
Lightly-tested ACK.

I don't know if there are conflicting views on this, but as a matter of personal preference, I tend to suggest waiting until after final review to squash (because it makes it easy for the reviewer to see what changed since the last review.)

I have rebased and will push.

gwillen:
Heh, github still claims this can be merged, probably because I did a rebase-push without updating the code on the pullreq. But it's in! Thanks for doing this. :-)

PR #88, opened 2016-03-11
Assets: Attempts to fix regtest

greenaddress:
This PR includes my attempts to get alphad to work in regtest mode. What I'm trying to do is generate 100 blocks and then `createrawtransaction` to issue a new asset.

 1. I got `CheckTransaction: vTxFees bigger than vout` by just running it without any modifications. This seems due to `>` check being reversed in `CheckTransaction` in `main.cpp`.
 2. After the change above alphad still fails with `ConnectBlock: coinbase pays too much` on startup. I relaxed the checks in `CCoinsViewCache::VerifyAmounts` in `coins.cpp` so they now allow coinbase with null asset id. Not sure if it's correct, but at least it makes the server start.
 3. However it now SIGSEGVs on `setgenerate true` due to `vout` being too small in `miner.cpp`. Moving the `resize` after the change to `mBlockReward` makes `setgenerate` at least succeed
 4. Now I have two more issues - first one is `createrawtransaction` failing with more than two arguments, which I attempted to fix by changing the checks in `rpcrawtransaction.cpp`. However it still fails with `{"code":-3,"message":"Expected type obj, got str"}`, not sure why.
 5. The last issue I'm having seems more serious. While the `setgenerate` calls seem to succeed, restarting the server now fails with `Corrupted block database detected.`. More logs after restart below.

```
2016-03-11 16:01:06 Verifying last 101 blocks at level 3
2016-03-11 16:01:06 ERROR: DisconnectBlock() : added transaction mismatch? database corrupted
2016-03-11 16:01:06 ERROR: DisconnectBlock() : added transaction mismatch? database corrupted
(...)
2016-03-11 16:01:06 ERROR: DisconnectBlock() : added transaction mismatch? database corrupted
2016-03-11 16:01:06 ERROR: DisconnectBlock() : added transaction mismatch? database corrupted
2016-03-11 16:01:06 ERROR: VerifyDB() : *** coin database inconsistencies found (last 101 blocks, 0 good transactions before that)
```

Now I'm not sure if this is due to the other changes I made above, but it seems like too much to debug for me now.

### DIFF
--- a/src/rpcrawtransaction.cpp
+++ b/src/rpcrawtransaction.cpp
@@ -454,7 +454,7 @@ Value listunspent(const Array& params, bool fHelp)
 
 Value createrawtransaction(const Array& params, bool fHelp)
 {
-    if (fHelp || params.size() != 2)
+    if (fHelp || params.size() < 2)
         throw runtime_error(
             "createrawtransaction [{\"txid\":\"id\",\"vout\":n},...] {\"address\":amount,...}\n"
             "\nCreate a transaction spending the given inputs and sending to the given addresses.\n"
@@ -484,12 +484,6 @@ Value createrawtransaction(const Array& params, bool fHelp)
             "      \"address\": \"id\"  (string, optional) The key is the bitcoin address, the value is the asset id\n"
             "      ,...\n"
             "    }\n"
-            "3. \"output asset ids\"    (object, optional) a json object with addresses as keys and asset ids as values\n"
-            "                                              All addresses not contained here default to -feeasset\n"
-            "    {\n"
-            "      \"address\": \"id\"  (string, optional) The key is the bitcoin address, the value is the asset id\n"
-            "      ,...\n"
-            "    }\n"
             "4. Asset definition        (boolean, optional, default=false) Make it an asset definition transaction. Outputs with assetID = 0 will represent the newly issued assets.\n"
 
             "\nResult:\n"
@@ -500,7 +494,13 @@ Value createrawtransaction(const Array& params, bool fHelp)
             + HelpExampleRpc("createrawtransaction", "\"[{\\\"txid\\\":\\\"myid\\\",\\\"vout\\\":0}]\", \"{\\\"address\\\":0.01}\"")
         );
 
-    RPCTypeCheck(params, list_of(array_type)(obj_type));
+    if (params.size() == 2) {
+        RPCTypeCheck(params, list_of(array_type)(obj_type));
+    } else if (params.size() == 3) {
+        RPCTypeCheck(params, list_of(array_type)(obj_type)(obj_type));
+    } else if (params.size() == 4) {
+        RPCTypeCheck(params, list_of(array_type)(obj_type)(obj_type)(bool_type));
+    }
 
     Array inputs = params[0].get_array();
     Object sendTo = params[1].get_obj();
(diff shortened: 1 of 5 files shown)


src/rpcrawtransaction.cpp
```diff
@@ -484,12 +484,6 @@ Value createrawtransaction(const Array& params, bool fHelp)
             "      \"address\": \"id\"  (string, optional) The key is the bitcoin address, the value is the asset id\n"
             "      ,...\n"
             "    }\n"
-            "3. \"output asset ids\"    (object, optional) a json object with addresses as keys and asset ids as values\n"
```
jtimon: why remove this?

jtimon:
Besides the nit, utACK.

src/rpcrawtransaction.cpp
greenaddress: It was duplicated.

jtimon: Oops

instagibbs:
>{"code":-3,"message":"Expected type obj, got str"}, not sure why.  

This will most likely be due to some dumb json::spirit error. Can you get the line that's throwing that?

greenaddress:
```
#0  RPCTypeCheck (params=std::vector of length 3, capacity 3 = {...}, typesExpected=std::list, 
    fAllowNull=fAllowNull@entry=false) at rpcserver.cpp:64
#1  0x00007f54e04449cb in createrawtransaction (params=std::vector of length 3, capacity 3 = {...}, 
    fHelp=<optimized out>) at rpcrawtransaction.cpp:500
#2  0x00007f54e03c9fbd in CRPCTable::execute (this=this@entry=0x7f54e09c4f20 <tableRPC>, 
    strMethod="createrawtransaction", params=std::vector of length 3, capacity 3 = {...})
    at rpcserver.cpp:1018
#3  0x00007f54e03ce429 in HTTPReq_JSONRPC (fRun=false, mapHeaders=std::map with 7 elements, 
    strRequest="{\"method\":\"createrawtransaction\",\"params\":[[{\"txid\":\"391520a77b69186b518aea58ab6c0aba64fa9b167fa102b7971074011873c87a\",\"vout\":1,\"nValue\":10500000}],{\"2MwaKXZC8x2MoLGrZDdqToW1uLdFVMNHxVr\":1,\"22E8QKHaTi"..., conn=0x7f54b00008c0) at rpcserver.cpp:925

(...)

(gdb) p strRequest
$16 = "{\"method\":\"createrawtransaction\",\"params\":[[{\"txid\":\"391520a77b69186b518aea58ab6c0aba64fa9b167fa102b7971074011873c87a\",\"vout\":1,\"nValue\":10500000}],{\"2MwaKXZC8x2MoLGrZDdqToW1uLdFVMNHxVr\":1,\"22E8QKHaTijFemPDwKvAk9qoTgagPfp8nBQiry87MMU1h2gQR9PDGpGgKrxhBek5mbVTrqQ3wALzvFpbd\":10},\"{\\\"2MwaKXZC8x2MoLGrZDdqToW1uLdFVMNHxVr\\\": 0}\"],\"id\":1}\n"
```

So it was a client error, it wasn't converting the last CLI arg to an object. Fixed. :)

jtimon:
Ok, the "ConnectBlock: coinbase pays too much" issue still remains. I will work on a more permanent solution, that's why I asked for the revert on the changes to coins.cpp.

PR #89, opened 2016-03-11
Chainparams: Don't check the genesis block, but store its transaction…

jtimon:
…s asvalid (HF for bitcoin)

- Like in #6597 (bitcoin), the genesis block shouldn't be checked
because it is a consensus rule in itself

- On the other hand, some elements like deterministic peg (because the
  21M reserve of pegged-coins needs to be created at some point) and
  multi-assets (because one cannot define new assets without existing
  utxos to be used as input for the asset definition transactions)


jtimon:
Well, restarting the node is still not fixed and the PR ended up with a bunch of other unrelated fixes, but we can clean up the history when rebasing.

PR #97, opened 2016-04-28
Fix LSHIFT test

jl2012:
The test passed since EQUAL is considered as part of the data push, leaving a TRUE top stack item


jtimon:

PR #98, opened 2016-05-18
Fixed sidechain-manipulation print bug when there are 0 confirmations

dickbrouwer:
This logic is invalid: If the "confirmations" key does not exists you cannot get the value from this key from the dict, it needs to do a "get-or-default-to-0-if-null". Traceback:

Traceback (most recent call last):
  File "./contrib/sidechain-manipulation.py", line 249, in <module>
    print("You must wait for at least %s confirmations to claim this output (have %d)" % (prev_script[9], prev_tx.get("confirmations")))
KeyError: 'confirmations'


instagibbs:
utACK

jonasnick:
utAck

jtimon:

PR #107, opened 2017-01-18
Add Changes From Liquid, Bitcoin

martindale:
This pull request adds a wide body of changes from both Liquid and the upstream Bitcoin project, including most notably an import of the Segregated Witness code as included in Bitcoin 0.13.1, replacing the original implementation in Elements Alpha.  Special thanks to @instagibbs, @jonasnick, and @jtimon for working on this.

Some highlights:
- **Performance**: Confidential Transaction processing has been greatly improved via rangeproof and balance check caching.
- **Stability**: Validation of the sidechain is much more robust to exceptional behavior
- **UX**: Fewer steps to starting your Elements node, pegging in testnet coins as well as creating your own sidechain.
- **Consensus Fixes**: A number of consensus issues have been identified and fixed, including reports from the community.
- **Core codebase**: This pull request is based on a recent version of Bitcoin Core, 0.13.1, bringing all the performance enhancements and updates with it.

The new Elements codebase also continues to include support for creating your own sidechain in a single-signer test environment, convenient for local development and experimentation.

Several more updates to follow.


Christewart:
This is meant to represent what was previously the `mainchain` branch, correct? The project previously had two branches that were both necessary for setting up elements. But this merges into `master`, not `mainchain` or `alpha`.

instagibbs:
@Christewart `master` is the "replacement" for the `alpha` branch once this PR is merged. For running a validating sidechain, you can checkout almost any Core version from the Bitcoin Core repo.

instagibbs:
Pushed some changes to allow pegins into blocks without manually setting at the command-line.

src/txdb.cpp
```diff
@@ -199,7 +199,7 @@ bool CBlockTreeDB::LoadBlockIndexGuts(boost::function<CBlockIndex*(const uint256
                 pindexNew->nStatus        = diskindex.nStatus;
                 pindexNew->nTx            = diskindex.nTx;
 
-                if (!CheckProofOfWork(pindexNew->GetBlockHash(), pindexNew->nBits, Params().GetConsensus()))
+                if (!CheckProofOfWork(pindexNew->GetBlockHash(), pindexNew->nBits, consensusParams))
```
jtimon: It seems this commit needs s/consensusParams/Params().GetConsensus()/ in this line to compile.

src/wallet/rpcwallet.cpp
```diff
@@ -2567,6 +2652,401 @@ UniValue fundrawtransaction(const UniValue& params, bool fHelp)
...
+    CTxOut dummyTxOut(0, relock_spk);
+    CAmount lockDust(dummyTxOut.GetDustThreshold(withdrawLockTxFee));
+
+    LOCK(cs_main);
+
+    std::vector<std::pair<COutPoint, CAmount> > lockedUTXO;
+    if (!GetLockedOutputs(genesisBlockHash, value+lockDust, lockedUTXO))
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Failed to find inputs with sufficient value - are you sure bitcoinTx is valid?");
+
+    while (lockedUTXO.size() != 1) {
```
Christewart: This loop predicate seems rather strange to me. Why exactly `lockecUTXO.size != 1`? What happens if we are given 0 utxos? Maybe there is a guard against that in code leading up to this, but it seems like this is fragile. 

Also, by reading this piece of code I have a hard time deciphering what is actually going on here. Are we creating a chain of transactions that consolidates locked UTXOs to one output on the sidechain?

instagibbs: `GetLockedOutputs` will return true IFF enough unspent unlocked funds were found. This should pretty much never return false(?) as it will spend in-mempool locked utxos as well, as a last resort. If it returns true that guarantees a non-0 amount of utxos.

>Are we creating a chain of transactions that consolidates locked UTXOs to one output on the sidechain?
Indeed. There are two types of OP_WPV spends(simplification):
1) Claims, which consume a single locked utxo and output a normal utxo
2) reductions, which take 2 utxos and combine them into 1

So if the value returned is spread among 5 utxos it will make a chain of 4 reductions, then a pegin transaction of the correct amount.

Christewart: So why create this chain of transactions instead adding all the inputs to one tx and creating a single output?

instagibbs:
It's due to how the consolation transaction scripts work. You might have to
dive into the interpreter code to grok it.

droark:
FYI, when I try to compile the *refresh* branch on OSX 10.12.3, it won't compile. This is due to a protobuf issue. This issue was [fixed in 0.14](https://github.com/bitcoin/bitcoin/pull/9366) but doesn't seem to have been backported to 0.13. I'm not sure how this bug ended up in the refresh.

Anyway, the immediate workaround is to [install protobuf 2.6](https://github.com/bitcoin/bitcoin/pull/9467). Backporting the 0.14 fix wouldn't hurt, though. :) I'll backport and test to make sure it works.

instagibbs:
@droark thanks for the heads up. I've pushed the commit here and tests seem to be passing. Could you ACK the build on the last commit?

instagibbs:
Assuming we don't have any more major changes I'm going to squash and merge today so we have a stable base to work on.

jtimon:
I still haven't reviewed this enough for a full utACK, but I've spent quite some time reviewing it during the last two weeks, finding mostly only things related to the order in which things are done for history cleanness. I plan to continue that work after this is merged to review further and as a preparation for the next rebase to 0.14.

8736efe complies, passes both unitttests and rpc tests.
When I run it with -extended (with pruning.py commented) I got some errors:
```
bip9-softforks.py              | False  | 6 s
bip65-cltv.py                  | False  | 7 s
bip68-sequence.py              | False  | 9 s
bipdersig.py                   | False  | 9 s
invalidateblock.py             | False  | 10 s
rpcbind_test.py                | False  | 10 s
```
But I don't think that should slow down merging this, we can fix those tests later. There's already 3 open PRs based on this: #109 #111 #112

droark:
@instagibbs - Thanks. Posthumous ACK on the backport. OSX builds Elements with this patch set. Haven't run it yet, though. Will run it later.

PR #109, opened 2017-01-21
Adding script to generate element blocks for 1 of 1 regtest

Christewart:
Adds a simple python script to generate blocks for elements on regtest


cdecker:
You might want to wrap the function call into 

```
if __name__ == "__main__":
    generate_n_blocks(100)
```
to make the module importable :smiley:

Christewart:
I'll add this later tonight, I also justt placed the script in the top level directory, should it be moved somewhere else?

Christewart:
Added changes suggested by @cdecker , also adding command line arguments so we can now use it like so: 
```
python generate_element_blocks.py --num=100 --datadir=$BETADATADIR
```
where `num` is the number of blocks to generate

generate_element_blocks.py
```diff
@@ -0,0 +1,30 @@
...
+
+def generate_n_blocks(n,datadir): 
+    for i in range(0,n): 
+        sign_block(datadir)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--num')
+    parser.add_argument("--datadir")
+    datadir="~/.bitcoin/elements.conf"
```
jtimon: Not sure if you want to do
datadir="~/.bitcoin/elements/"
conf="~/.bitcoin/elements/elements.conf"

jtimon:
Needs rebase (just remove all the commits except yours and I think it should rebase without conflicts).

jtimon:
Actually, no, it needs to be reopened as a PR to master instead of refresh, sorry for the oversight.

PR #111, opened 2017-02-01
HF: Move block.proof.challenge to Consensus::Params::signblockScript

jtimon:
Based on #107.

Since the scriptPubKey for signing blocks never changes, there's no point in repeating it with every block header. This saves more space in memory the more signatures there are in a particular chain.

Also includes a hash of fedpegscript and signblockscript in the genesis block which is also a hf, but makes the genesis block depend on them.


instagibbs:
Concept ACK, but the one unknown I have with this is that it may effect the simplicity/design of HSM schemes or how we can swap signer sets more on the fly with future changes.

jtimon:
Closing, see https://github.com/ElementsProject/elements/pull/109#issuecomment-277015584

jtimon:
Reopened as  #111

PR #112, opened 2017-02-02
Migrate `elements-cli` to `elements`

martindale:
I think it'd be nice if `elements` was the primary command used to interact with your local daemon via the command line, rather than the redundant `elements-cli`.  Maybe upstream would like to do something similar at some point.


Christewart:
Ehh, if you I'm a slight nACK on this. If you feel strongly about it I would be fine changing it. I think it is important to be consistent across codebases though. I doubt it is going to get changed in bitcoin, so is it worth setting a precedences here?

instagibbs:
Personally light NACK as well. Mostly due to inconsistency with upstream and because people can always rename the binary.

If it is getting changed upstream we can sync up in this way definitely.

jtimon:
Needs rebase (just remove all the commits except yours and I think it should rebase without conflicts).

jtimon:
Closing, see https://github.com/ElementsProject/elements/pull/109#issuecomment-277015584

PR #114, opened 2017-02-02
HF: Move block.proof.challenge to Consensus::Params::signblockScript

jtimon:
Replaces  #111
Since the scriptPubKey for signing blocks never changes, there's no point in repeating it with every block header. This saves more space in memory the more signatures there are in a particular chain.

Also includes a hash of fedpegscript and signblockscript in the genesis block which is also a hf, but makes the genesis block depend on them.

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -21,21 +21,20 @@
 #include "wallet/wallet.h"
 #endif
 
-CScript CombineBlockSignatures(const CBlockHeader& header, const CScript& scriptSig1, const CScript& scriptSig2)
+CScript CombineBlockSignatures(const Consensus::Params& params, const CBlockHeader& header, const CScript& scriptSig1, const CScript& scriptSig2)
 {
     SignatureData sig1(scriptSig1);
     SignatureData sig2(scriptSig2);
-    return GenericCombineSignatures(header.proof.challenge, header, sig1, sig2).scriptSig;
+    return GenericCombineSignatures(params.signblockScript, header, sig1, sig2).scriptSig;
 }
 
 bool CheckChallenge(const CBlockHeader& block, const CBlockIndex& indexLast, const Consensus::Params& params)
 {
-    return block.proof.challenge == indexLast.proof.challenge;
+    return true;
 }
 
 void ResetChallenge(CBlockHeader& block, const CBlockIndex& indexLast, const Consensus::Params& params)
 {
-    block.proof.challenge = indexLast.proof.challenge;
 }
 
 bool CheckBitcoinProof(const Sidechain::Bitcoin::CBlockHeader& block)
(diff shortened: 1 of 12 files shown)


instagibbs:
I think these changes should be proposed separately or in sequence. One changes the header format which can effect how the HSMs operate and will need more review from more people.

jtimon:
As discussed with @instagibbs , since there's no hurry to do the hardfork, the first change can wait for review on the second one.

Christewart:
utACK 38bf14a

jtimon:
Needed rebase

jtimon:
Sorry, some unittests are failing
```
TEST                           | PASSED | DURATION

mempool_resurrect_test.py      | True   | 3 s
p2p-fullblocktest.py           | True   | 3 s
txn_doublespend.py --mineblock | True   | 5 s
txn_clone.py                   | True   | 6 s
rest.py                        | True   | 6 s
mempool_spendcoinbase.py       | True   | 4 s
mempool_reorg.py               | True   | 4 s
httpbasics.py                  | True   | 4 s
multi_rpc.py                   | True   | 4 s
wallet.py                      | True   | 12 s
proxy_test.py                  | True   | 6 s
merkle_blocks.py               | True   | 6 s
signrawtransactions.py         | True   | 3 s
fundrawtransaction.py          | True   | 5 s
getchaintips.py                | True   | 16 s
decodescript.py                | True   | 3 s
sendheaders.py                 | True   | 4 s
prioritise_transaction.py      | True   | 3 s
invalidblockrequest.py         | True   | 3 s
invalidtxrequest.py            | True   | 3 s
zapwallettxes.py               | True   | 16 s
keypool.py                     | True   | 7 s
abandonconflict.py             | True   | 4 s
segwit.py                      | True   | 4 s
signmessages.py                | True   | 3 s
reindex.py                     | True   | 14 s
nodehandling.py                | True   | 15 s
zmq_test.py                    | True   | 5 s
listtransactions.py            | False  | 68 s
receivedby.py                  | False  | 71 s
rawtransactions.py             | False  | 77 s
confidential_transactions.py   | False  | 84 s

ALL                            | False  | 471 s (accumulated)

Runtime: 104 s
```

jtimon:
Updated with some feedback from @jonasnick

instagibbs:
Could you share the feedback here?

jtimon:
Yes, sorry.
Basically we discussed the fields "bits" and "target" aren't needed in the existing places neither in rpc/blockchain nor in getBlockTemplate and new fields for the signblock script as a hex string and as assembly (as was done in the previous places) in getblockchaininfo can replace them.
The hex is necessary for a block signer. previously, it could simply extract the the challenge from the serialized block header.
And those are the changes I made to the last commit, the first commit wasn't modified, but I rebased.

jonasnick:
I wanted to be able to get the signblockscript as hex somehow from the RPC which is why @jtimon added `signblockhex` to the getblockchaininfo RPC. Without the change you can only get the signblockscript in asm string format with getblock.

jtimon:
Sorry, added one commit, potentially to squash to the previous one, or leave for #164

src/pow.cpp
```diff
@@ -21,21 +21,20 @@
...
     SignatureData sig1(scriptSig1);
     SignatureData sig2(scriptSig2);
-    return GenericCombineSignatures(header.proof.challenge, header, sig1, sig2).scriptSig;
+    return GenericCombineSignatures(params.signblockScript, header, sig1, sig2).scriptSig;
 }
 
 bool CheckChallenge(const CBlockHeader& block, const CBlockIndex& indexLast, const Consensus::Params& params)
 {
-    return block.proof.challenge == indexLast.proof.challenge;
+    return true;
```
instagibbs: What use is this stub?

jtimon: At first I thought we may want to change the challenges, for example, for key rotation. This function would allow to restrict the allowed transitions from one challenge to another. Since we're not using any of that, there's no point in maintaining the challenge field in the header, but we can leave the functions here in case we want to restore the field to do something like that later.

jtimon:
Replaced by https://github.com/ElementsProject/elements/pull/189

PR #123, opened 2017-02-09
build: fix qt5.7 build under macOS

kallewoof:
QT does not get the C++11 std flag properly with the 0.13.1 version. This commit by @theuni fixes.

[from https://github.com/bitcoin/bitcoin/pull/9169]

OBJCXX's std flags don't get defined by our cxx macro. Rather than hard-coding
to c++11, just force OBJCXX to be the same as CXX unless the user specified
otherwise.

Github-Pull: #9169
Rebased-From: 70266e98292bf9c8f019ca4b11d681b575f71448

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -43,7 +43,6 @@ else
   CXXFLAGS_overridden=no
 fi
 AC_PROG_CXX
-m4_ifdef([AC_PROG_OBJCXX],[AC_PROG_OBJCXX])
 
 dnl By default, libtool for mingw refuses to link static libs into a dll for
 dnl fear of mixing pic/non-pic objects, and import/export complications. Since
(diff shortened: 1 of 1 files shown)


martindale:
Thank you for contributing, @kallewoof!  LGTM.  Does anyone have a Mac that they can test this with?

wtogami:
@martindale Don't you have a Mac?

martindale:
@wtogami I do, but I run Fedora.  OSX won't boot... :joy:

bteitelb:
Works for me. Without this change, elements fails to compile on OSX with the error "Qt requires C++11 support". Previously, I worked around this by doing ```export CXXFLAGS=-std=c++11```, but this PR obviates the need to do that. Builds out of the box now just fine, albeit with 11 warnings about using deprecated functions like OSAtomicCompareAndSwap64 and OSAtomicAdd64, but those were there before. Thanks @kallewoof.

PR #128, opened 2017-02-22
[Trivial] Change 'bitcoin-cli' to 'elements-cli' in error message

anditto:
Fixed the error message when trying to use RPC commands with elementsd.


instagibbs:
ACK

PR #132, opened 2017-02-28
RPC: Introduce -rpcamountdecimals for the RPC to use other units than BTC

jtimon:
Backport of https://github.com/bitcoin/bitcoin/pull/9882
Even if that doesn't get merged upstream, we need something like this before the assets PR. It would be nicer to use satoshis/minimum_unit for every asset instead of using 8 decimals for every asset.


jtimon:
From the discussion in bitcoin/bitcoin#9882, it's not clear what the plan is regarding using satoshis for everything in the rpc.
For assets other than BTC, it still doesn't make sense to use 8 decimals.
In a multi-asset branch, we can let the user configure the decimals for named assets, set "bitcoin" to 8 by default and all unknown assets to 0 decimals.
Closing

PR #134, opened 2017-03-01
Update build-unix.md with elements specific changes from bitcoin

kiarafrobles:


martindale:
Great work!  Thanks Kiara!

doc/build-unix.md
```diff
@@ -1,30 +1,11 @@
...
-Note
----------------------
-Always use absolute paths to configure and compile bitcoin and the dependencies,
-for example, when specifying the path of the dependency:
-
-	../dist/configure --enable-cxx --disable-shared --with-pic --prefix=$BDB_PREFIX
-
-Here BDB_PREFIX must be an absolute path - it is defined using $(pwd) which ensures
-the usage of the absolute path.
+This will build Elements-Qt as well if the dependencies are met.
```
martindale: I don't think anyone has built the Qt-enabled binary for some time.  Unless we are sure this works, I don't know that we should include it in the guide.

instagibbs: I build the binary... can't say that it's safe to use though.

doc/build-unix.md
```diff
@@ -82,7 +63,7 @@ install necessary parts of boost:
 BerkeleyDB is required for the wallet. db4.8 packages are available [here](https://launchpad.net/~bitcoin/+archive/bitcoin).
 You can add the repository and install using the following commands:
 
-    sudo add-apt-repository ppa:bitcoin/bitcoin
+    sudo add-apt-repository ppa:ElementsProject/elements
```
instagibbs: is that true? do we have a ppa for this? I kind of doubt it.

doc/build-unix.md
```diff
@@ -147,6 +128,31 @@ Notes
 The release is built with GCC and then "strip elementsd" to strip the debug
 symbols, which reduces the executable size by about 90%.
 
+Build `Elements`
+---------------------
+
+1. Clone the elements source code and cd into `elements`
+
+		git clone https://github.com/ElementsProject/elements.git
```
instagibbs: spacing in this section looks wonky to me

jtimon:
Concept ACK, but as said I would minimize differences with upstream and maybe do an upstream PR if they make sense. Related to @instagibbs comment on not moving things. 
Also I think we could remove all instructions referencing bdb, ideally removing it from the build (although perhaps that last part belongs to another PR). Bitcoin Core doesn't want to screw people who upgrade to a newer version with an old wallet. We don't care about that, there's no elements "old wallets" depending on bdb.

doc/build-unix.md
martindale: We can make one if it doesn't already exist!  Once I finish with signed builds, we most likely will.

jtimon:
Closing in favor of https://github.com/ElementsProject/elements/pull/208

PR #164, opened 2017-04-20
Refactor: Easier support for chains with multiple genesis assets

jtimon:
Some POC chains may be interested in having more than one genesis asset for spiking the additional step of issuing a second asset (or more) for a particular demonstration. Presumably they would copy and modify CElementsParams to create their own chain. 

For the 2 way peg, as a next step, a multiasset chain can support pegs to the same parent chain with different federation scripts, or even pegged to different parent chains. We don't support that yet, but in the same way this makes testing that kind of extension easier to test without needing to issue new pegged assets during the tests. Issuing them on the fly can be a next step.

Intentionally, the new "feature" isn't tested, for that would require either a "genesis block id HF" or a new chain. I would prefer if that is done in a later PR (see https://github.com/jtimon/elements/compare/e13-multi-genesis...jtimon:e13-multi-genesis-test for a WIP) and this is merged without testing first purely on the grounds that the behavior should remain perfectly equivalent and if at any point this is not the case for this PR, it should be nacked by reviewers.

Dependencies:

- [ ] HF: Move block.proof.challenge to Consensus::Params::signblockScript #114

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -19,6 +19,8 @@
 
 #include "chainparamsseeds.h"
 
+static const uint32_t MAX_GENESIS_OUTPUTS = 500;
+
 // Safer for users if they load incorrect parameters via arguments.
 static std::vector<unsigned char> CommitToArguments(const Consensus::Params& params, const std::string& networkID)
 {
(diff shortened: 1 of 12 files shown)


instagibbs:
quick read ACK

jtimon:
Replaced by https://github.com/ElementsProject/elements/pull/191

PR #166, opened 2017-05-02
Separate config files from assets tutorial

jtimon:
... and some minor additions

I believe this makes the tutorial slightly more readable and practical.


instagibbs:
IIRC there were sections that required editing the configuration file and reloading. I don't think that's supported here?

jtimon:
I can't find any reference in the file to editing the config files beyond the comment added here starting with "Alternatively". And I don't see how this prevent editing the files. This is just a simpler way to create them (by copying them once), but they can still be edited afterwards like before.

instagibbs:
Nevermind I was confused; I removed those steps a long time ago.

Christewart:
I agree with the readable/practical sentiment. Concept ACK.

instagibbs:
@Christewart would you mind walking through the new format, make sure everything works and see what you think?

Christewart:
So I tried running the new shell script and got the following errors: 

https://pastebin.com/r1Br6795

Admittedly, I never ran the old file. I would always do the setup from the elements project website. Maybe I am missing an obvious step.

jtimon:
The script didn't run previously as a whole. I tried the issuance and block-singing parts but not the things below "######## Pegging #######" (although I'm pretty sure this doesn't break them).

contrib/assets_tutorial/assets_tutorial.sh
```diff
@@ -1,89 +1,20 @@
...
-daemon=1
-txindex=1
-EOF
+# Let's have some testing user directories for 1 bitcoin node and 2 elements nodes.
+rm -r ~/bitcoindir ; rm -r ~/elementsdir1 ; rm -r ~/elementsdir2
+mkdir ~/bitcoindir ; mkdir ~/elementsdir1 ; mkdir ~/elementsdir2
+
+# Also configure the nodes by copying the configuration files from
+# this directory (and read them):
+cp ./contrib/assets_tutorial/bitcoin.conf ~/bitcoindir/bitcoin.conf
```
instagibbs: this assumes which directory we're currently in without stating what it is, including the `cd src` part.

Maybe state this.

jtimon: I think this was the case already, but sure, I can clarify.

PR #168, opened 2017-05-02
Document -signblockscript and -fedpegscript in daemon help

jtimon:
Also document -validatepegin.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -403,6 +403,8 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-uacomment=<cmt>", _("Append comment to the user agent string"));
     if (showDebug)
     {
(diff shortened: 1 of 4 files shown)


Christewart:
May be a dumb question, but what happens to the data directory if you change `signblockscript` / `fedpegscript`? Since we are creating a new chain, does it over write everything?

instagibbs:
@Christewart a different `signblockscript` will trigger an "incorrect genesis block" error since the block changes. We also have a change to commit to the `fedpegscript` as well, to reduce chances of losing money: https://github.com/ElementsProject/elements/pull/164

This can be bypassed by reindexing or deleting the folder.

instagibbs:
@jtimon you're missing `DEFAULT_VALIDATE_PEGIN` definition.

jtimon:
Oops, fixed, thanks.

instagibbs:
compile/utACK

PR #170, opened 2017-05-10
[backport] Update for OpenSSL 1.1 API

jonasnick:
Without this PR elements doesn't compile on my machine with openssl 1.1.0e. (`wallet/test/crypto_tests.cpp:45:20: error: aggregate ‘EVP_CIPHER_CTX ctx’ has incomplete type and cannot be defined`)


instagibbs:
Note: this was backported to Core 0.13.2

PR #175, opened 2017-05-26
Trivial: Fix help text in rpc/rawtransaction.cpp

AkioNak:
1. rawblindrawtransaction
- command line
add inputamount, inputasset, inputassetblinder and ignoreblindfail
change totalblinder to simple value
set totalblinder, ignoreblindfail to optional
- parameter
add optional to ignoreblindfail

2. blindrawtransaction
- command line
move ignoreblindfail to 2nd
change totalblinder to simple value
set totalblinder, array of assetcommitment, ignoreblindfail to optional
- parameter
add optional to ignoreblindfail
fix assetcommitments to assetcommitment


instagibbs:
note: "Trivial" upstream means things that don't change the binary... but this is simple, thanks! utACK

PR #178, opened 2017-06-06
Untested: Rpc call for per-block fee statistics

jtimon:
I still need to test it and squash, and there's one TODO, but review welcomed already.

### DIFF
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -201,6 +201,17 @@ CAmountMap CTransaction::GetFee() const
     return fee;
 }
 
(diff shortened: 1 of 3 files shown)


instagibbs:
Is there an associated issue with this?

jtimon:
No, it's just a proposed new feature. In Bitcoin Core this would be much more complicated, but since CT needs explicit fees anyway, it is pretty simple to do it here.

jtimon:
squashed for clarity

jtimon:
A more correct version (with corrections after writting the tests) is in https://github.com/ElementsProject/elements/pull/185 for 0.14 instead of 0.13. Closing this one.

PR #182, opened 2017-06-12
fixed gettxout rpc bug + invalidateblock.py syntax

tdudz:
instead of cast to uint256, just convert commitment to hex string using HexStr()

invalidateblock.py had incorrect python syntax, it now passes


apoelstra:
Please add a RPC test for this in `qa/`

instagibbs:
Note we should also be returning `asset`/`assetcommitment` as well.

jtimon:
@instagibbs Can you confirm that this needs rebase to 0.14 ?

PR #184, opened 2017-06-19
[upstream #10627] fixed listunspent rpc convert parameter

tnakagawa:


instagibbs:
This appears to be a bug upstream in Core. Mind opening another PR there too to make sure we don't need this next rebase?

jtimon:
Perhaps also add { "listunspent", 4, "asset" } (only here, not upstream the asset field to bitcoin core, obviously).
Apart from that, utACK

tnakagawa:
'VRPCConvertParams' can only set parameters other than strings?

instagibbs:
Yes only for no strings.

jtimon:
Oh, right, my bad. Should correct #185 ...
re-utACK

apoelstra (approved):
ACK 789cb198eac580a8c5bb42da65cd6551e8707dce

PR #185, opened 2017-06-21
Rpc: Introduce call for per-block fee statistics

jtimon:
EDIT: Don't merge until https://github.com/bitcoin/bitcoin/pull/10757 is resolved (please, review that first)

New feature. In Bitcoin Core this would be much more complicated, but since CT needs explicit fees anyway, it is pretty simple to do it here.
The idea is to generate data to plot.

Replaces https://github.com/ElementsProject/elements/pull/178 (which was incomplete and based on 0.13 instead of 0.14).


jtimon:
Upstreamed as https://github.com/bitcoin/bitcoin/pull/10757

jtimon:
With respect to bitcoin/bitcoin#10757 this adds an asset optional parameter to the call that is not tested, the tests assume the default policyAsset is used for everything.
If not for fees, perhaps it would be nice to show charts for per-asset total amounts publicly moved, total number of txs involving a public use of an asset, opting out of the CA feature, even if you hide the amounts, ins, outs, number of public keys involved per block per asset, I bet other people can find more creative ways to chart inadvisable things as "public" than just using "the null nounce" for your CA txs just to show off on these "you shouldn't be doing this per-public-asset" charts.
I guess the same trick "null nounce" trick could apply to CT txs, and there's also public amounts without CT, and we could maybe plot the different combinations and their effect for some heuristics for "privacy" (I can't write that function, or those functions), and maybe the privacy/weight ratio too.

Anyway, future work...

jtimon:
Rebased just to get travis and one fix related to building without wallet in particular documented in https://github.com/ElementsProject/elements/pull/243 (through https://github.com/ElementsProject/elements/pull/243/commits/2d88f075ba6f992c778b1f24cb2eec634158d319 )

jtimon:
Re-backported

instagibbs:
unrelated, merging since it looks near-final upstream, minus stupid unicorns

nkostoulas:
Hello

I am getting this:
rpc_getblockstats.py:
Initializing test directory /var/folders/62/hp31g7h106q0t6_yjdyz19bw0000gn/T/testynxolldw/243
JSONRPC error: Block does not start with a coinbase
Stopping nodes
Not cleaning up dir /var/folders/62/hp31g7h106q0t6_yjdyz19bw0000gn/T/testynxolldw/243
Failed

when trying to run this test. Is this expected?

PR #186, opened 2017-06-21
[rpc] Add getblindedaddress RPC command

kallewoof:
This turns an unblinded address into its blinded equivalent, if known.

This command is useful when examining (raw) transactions which have outpoints to your wallet. It is possible to get the blinded version of the address by passing the address argument in getrawtransaction to getblindedaddress.

(Note that elements-0.14 is broken on macOS atm, which is why I am PR'ing the 0.13 branch.)

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -654,6 +654,39 @@ UniValue dumpwallet(const UniValue& params, bool fHelp)
     return NullUniValue;
 }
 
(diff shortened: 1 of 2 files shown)


instagibbs:
Not sure if you didn't know this but I think you can get the same functionality calling `validateaddress`.

kallewoof:
D'oh. I didn't realize it worked backwards like that. Thanks.

instagibbs:
Obscure feature, maybe @achow101's work to split upstream would allow it to be in a more logical place

PR #190, opened 2017-06-22
elementsd does not start on MacOS

AkioNak:
elementsd does not start on MacOS because of LOCK(cs_args) failuer when setting CChainParams::supportedChains.
the type of supportedChains is simple map (key/string, value/uint256), but its value were never used.
so change its type to 'vectorstd::string' instead of.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -340,10 +340,10 @@ public:
 };
 
 
-const std::map<std::string, uint256> CChainParams::supportedChains =
-    boost::assign::map_list_of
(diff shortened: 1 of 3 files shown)


jtimon:
Yes, the hash used to be hardcoded there and then there was an assert on each respective chainparams constructor. But if we're putting it there dynamially the assert cannot be done and there's no point in having the hash there.

utACK https://github.com/ElementsProject/elements/pull/190/commits/dd0431578ea4b6d1979ff93846851779b96e0be2

kallewoof:
ACK dd0431578ea4b6d1979ff93846851779b96e0be2

PR #189, opened 2017-06-22
HF: Move block.proof.challenge to Consensus::Params::signblockScript

jtimon:
Like #114 but rebased to 0.14.

Dependencies:

- [x] HF: Genesis block includes fedpegscript #187
- [x] Remove CChainParams::bitcoinID #188

PR #188, opened 2017-06-22
Remove CChainParams::bitcoinID

jtimon:
Remove unneeded code.
Using Bitcoin genesis mainnet hash for pegged-bitcoin asset id was a
mistake in the first place, anyway.

PR #187, opened 2017-06-21
HF: Genesis block includes fedpegscript and signblockscript

jtimon:
The first part of https://github.com/ElementsProject/elements/pull/114 rebased to 0.14.
Since the whole thing isn't getting the review I thought it would be good to separate it.

PR #191, opened 2017-06-22
Refactor: Easier support for chains with multiple genesis assets

jtimon:
Replaces #164 .
See documentation there.

Dependencies:

- [x] HF: Genesis block includes fedpegscript #187
- [x] Remove CChainParams::bitcoinID #188
- [ ] HF: Move block.proof.challenge to Consensus::Params::signblockScript #189

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -20,13 +20,13 @@
 #include "chainparamsseeds.h"
 
 // Safer for users if they load incorrect parameters via arguments.
-static std::vector<unsigned char> CommitToArguments(const Consensus::Params& params, const std::string& networkID, const CScript& signblockscript)
+static std::vector<unsigned char> CommitToArguments(const Consensus::Params& params, const std::string& networkID)
 {
     CSHA256 sha2;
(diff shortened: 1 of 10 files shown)


jtimon:
Rebased after #187 was merged for clarity.

jtimon:
Rebased to get the travis build.
But probably this will be closed after #260 has been merged.

jtimon:
@instagibbs  I'm still very early in review of https://github.com/ElementsProject/elements/pull/260 but it seems the pegged asset is issued and can be issued further after genesis so there's no need to support several genesis assets as this PR proposed, as far as I understand, one could create a chain with 100 pegged assets (even if mutli-pegged-asset is not supported in the interpreter and the rest of the 2wp componenets yet) and there's no need to put anything in the gensis tx besides a single unit of a single genesis asset since that used once to create a second asset is enough entropy as long as you don't destroy the genesis asset and all the other assets created after it. Besides tahat destroying it is fine, but the point is it should be possible to use it only once and and destroy it.

And because of this text wall, closing, this should be no longer necessary.

PR #195, opened 2017-06-30
Fix help text in wallet/rpcwallet.cpp

AkioNak:
- command line
add assetlabel 
- arguments
change 'ALL ACCOUNT ARGUMENTS' to 'ALL OTHER ARGUMENTS'
- examples
specify asset label instead of DEPRECATED pattern


instagibbs:
utACK

PR #196, opened 2017-07-04
Remove redundant code (issueasset)

AkioNak:
nAmount and nTokens never be negative value because AmountFromValue throws JSONRPCError if its argument holds negative value.


instagibbs:
ACK

Issue #180, opened 2017-06-12
"amount" (also "asset"?) field in transaction inputs in `createrawtransaction` is required but not actually used


apoelstra:

instagibbs:
Vestigial arguments, we should remove them.

instagibbs:
fixed in #197

PR #197, opened 2017-07-06
Remove nValue/amount arguments from createrawtransaction

instagibbs:
Previously these values were used to explicitly calculate the fee. Since this construction was more brittle to advanced use-cases with multiple fees/actors, explicit fee outputs are now added, therefore this is now unneeded. 

TODO is to remove all the passed nValue/amount arguments, but no hurry since they are simply ignored.

Solves https://github.com/ElementsProject/elements/issues/180


apoelstra:
ACK

PR #205, opened 2017-07-11
f 'Add fast Merkle branch functions.'

maaku:
Turns out the path calculation code, as written, did not work. Additionally, doing the position -> path calculation at validation time is both wasteful and introduces a source of malleability. Thankfully none of this code is used anywhere yet.

### DIFF
--- a/src/consensus/merkle.cpp
+++ b/src/consensus/merkle.cpp
@@ -181,36 +181,37 @@ uint256 ComputeFastMerkleRoot(const std::vector<uint256>& leaves) {
     return hash;
 }
 
-std::vector<uint256> ComputeFastMerkleBranch(const std::vector<uint256>& leaves, uint32_t position) {
+std::pair<std::vector<uint256>, uint32_t> ComputeFastMerkleBranch(const std::vector<uint256>& leaves, uint32_t position) {
     std::vector<uint256> ret;
     MerkleComputation(leaves, NULL, NULL, position, &ret, MERKLE_COMPUTATION_FAST);
-    return ret;
-}
-
-uint256 ComputeFastMerkleRootFromBranch(const uint256& leaf, const std::vector<uint256>& vMerkleBranch, uint32_t nIndex) {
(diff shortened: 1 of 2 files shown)


maaku:
This is a fix for the referenced commit in the elements repo, which introduces "fast Merkle branches." These Merkle tree structures are used for the witness data in confidential assets, but are a general data structure with potentially many use cases. It is slightly more than 3x faster to validate a branch of a fast Merkle tree then the regular ones currently in use by Bitcoin Core. 

As currently used, the confidential assets witness commitment code creates a lists of hashes and then does a fast Merkle root calculation. This use case is entirely unaffected by this change.

Where the code failed is in validation of an extracted path of a branch. Specifically, the 'fast Merkle tree' does not do any unnecessary hashing along the right-hand side of the tree, when the number of items in the tree is not a perfect power of 2. The commit in this pull request adds a comment to the header explaining what I mean with an example. When that happens the path through the tree is not the same as the index of the element, as it normally would be. So there is code that calculates what the path should be, in these corner cases, and then validates based on the calculated path. That code was wrong.

Additionally, this path modification code was being done during validation. This unfortunately makes it very easy to foot-gun malleability into protocols because there are then multiple initial paths that are accepted since all 'compress' down to the same underlying path. This commit therefore does the path compression in the function that calculates a Merkle branch, and now returns a (branch, path) pair where it was originally only returning branch. So when you extract a branch you are told what path to pass in to the validator to have it work -- which is usually the same as the 'position' argument you passed in, but can sometimes be reduced when following the right-hand side of the tree.

instagibbs:
Merging as is since bugfix seems correct. Would still like to revisit the explanation text at some point.

Christewart:
This might be unrelated, but it seems like this stuff would be easy to test with the property based testing pull request I opened upstream. https://github.com/bitcoin/bitcoin/pull/8469

Property based testing is really good at testing edge cases to make sure your code you wrote is correct.

PR #212, opened 2017-07-14
[trivial] remove unused variable "amount" in rawtransaction.cpp

tdudz:
[trivial] remove unused variable "amount" in rawtransaction.cppthe keyword "amount" was changed to "value" and is handled earlier if the output value is explicit. the "amount" variable was unused and currently is not necessary.


instagibbs:
utACK

PR #216, opened 2017-07-18
listissuances should outputs asset/tokenamount as -1 if the asset is blinded.

AkioNak:


instagibbs:
could you add a test in confidential_transactions.py ?

AkioNak:
Yes, I will add a test.

AkioNak:
@instagibbs done.

instagibbs:
utACK

PR #217, opened 2017-07-18
fix listreceivedbyaddress to sum amounts for each asset.

AkioNak:
Even if it receive multiple assets at the same address, listreceivedbyaddress ignores the asset type and sums it. listreceivedbyaccount and getreceivedbyaccount does same behavior too.

I think, like gettreceivedbyaddress, to sum amounts for each asset is better.


instagibbs:
tACK https://github.com/ElementsProject/elements/pull/217/commits/e491e5bdffd59878a96a1ec7c286d50b2b1125e6

Could you add a test in confidential_transactions.py ?

AkioNak:
Yes, I will add a test.

AkioNak:
@instagibbs done.

PR #230, opened 2017-08-07
fix rpc-tests to avoid illegal termination.

AkioNak:
Following fixes make "./qa/pull-tester/rpc-tests.py -extended" to normal termination.

- rpcbind_test.py
  - fix invalid indentation.
- test_framework/util.py
  - stop_nodes() accept NoneType. (rpcbind.py and smartfees.py call stop_nodes with no nodes because  their setup_network() does not start any nodes and actually does not test.)

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -169,7 +169,7 @@ if ENABLE_ZMQ:
 testScriptsExt = [
     'pruning.py',
     # vv Tests less than 20m vv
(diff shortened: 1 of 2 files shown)


instagibbs:
I'd prefer we remove the `return` statements, and comment out the tests in rpc-tests.py.

I'm ok with tests failing is nodes fail to start. I'd prefer a quick PR which does that.

AkioNak:
@instagibbs done.

PR #235, opened 2017-08-09
getnewblockhex would use regtest coinbase hard coded, instead of fetc…

kallewoof:
…hing CoinbaseDestination from Params.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -168,7 +168,9 @@ UniValue getnewblockhex(const JSONRPCRequest& request)
             + HelpExampleCli("getnewblockhex", "")
         );
 
-    std::unique_ptr<CBlockTemplate> pblocktemplate(BlockAssembler(Params()).CreateNewBlock(CScript() << OP_TRUE));
+    CScript feeDestinationScript = Params().CoinbaseDestination();
+    if (feeDestinationScript == CScript()) feeDestinationScript = CScript() << OP_TRUE;
+    std::unique_ptr<CBlockTemplate> pblocktemplate(BlockAssembler(Params()).CreateNewBlock(feeDestinationScript));
     if (!pblocktemplate.get())
         throw JSONRPCError(RPC_INTERNAL_ERROR, "Wallet keypool empty");
     {


instagibbs:
I think this will actually blackhole fees in regtest by making them unspendable. We use `CScript()` as the market for "anything goes" and set it as whatever we like. For `generate` we initially set it to OP_TRUE, then try and get an address from the wallet.

Could you ape that functionality, at least the OP_TRUE part?

kallewoof:
@instagibbs Hum.. wouldn't just setting the CoinbaseDest for regtest to be `CScript() << OP_TRUE` be enough? I'll have to try it.

kallewoof:
Yeah, that worked. I guess that means all regtests will have a new genesis block, though. That's probably not ideal.

Actually, I tried recompiling current 0.14 branch and running against same chain state and it worked. Not entirely sure why (even with `reindex` and `rescan`). Am I missing something?

instagibbs:
That would mean you cannot send it anywhere but OP_TRUE. generate calls
with a wallet will break, for example.

kallewoof:
I'm confused.. I can generate and getnewblockhex without problems, with transactions in the mempool. Can you show me an example of it breaking?

instagibbs:
We may be talking past eachother, the relevant line is here: https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/validation.cpp#L2549

Took a look at the logic. Indeed it will work, but then it means you *must* send fees to OP_TRUE. This isn't the desired behavior. In regtest we want to be able to send it anywhere we like.

kallewoof:
Ahh, okay. The line you pointed out could be changed to check for (instead/also) `CScript() << OP_TRUE`. That would solve it, right?

instagibbs:
Then you'd have to change the other places that check for the `CScript()` condition, like `generate`(and possibly other places, cannot recall). I'd prefer the minimal fix here of checking for current condition.

kallewoof:
OK, so in getnewblockhex, we use CScript << OP_TRUE if coinbase dest is CScript() and we use coinbase otherwise.

That's what you mean, right?

instagibbs:
Yep!

kallewoof:
Done.

instagibbs:
ACK

PR #240, opened 2017-08-11
fix SetToAsset bug where vchCommitment wasn't cleared before writing

tdudz:
clear the vchCommitment vector first, otherwise you'll end up appending to the end of what was already in it instead of replacing it

### DIFF
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -12,6 +12,7 @@
 
 void CConfidentialAsset::SetToAsset(const CAsset& asset)
 {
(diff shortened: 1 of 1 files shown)


src/primitives/transaction.cpp
```diff
@@ -12,6 +12,7 @@
 
 void CConfidentialAsset::SetToAsset(const CAsset& asset)
 {
+	vchCommitment.clear();
```
instagibbs: spacing is off

instagibbs:
aaaand a squash?

PR #248, opened 2017-08-26
p2p: Remove -mempoolreplacement

jtimon:
~~Is this doing anything beyond killing death code?~~
Always allow mempool replacement.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -80,7 +80,6 @@ bool fCheckpointsEnabled = DEFAULT_CHECKPOINTS_ENABLED;
 size_t nCoinCacheUsage = 5000 * 300;
 uint64_t nPruneTarget = 0;
 int64_t nMaxTipAge = DEFAULT_MAX_TIP_AGE;
-bool fEnableReplacement = DEFAULT_ENABLE_REPLACEMENT;
 
 uint256 hashAssumeValid;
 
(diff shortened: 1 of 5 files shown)


src/validation.cpp
```diff
@@ -1107,15 +1106,10 @@ bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, const C
...
-                if (fEnableReplacement)
-                {
-                    BOOST_FOREACH(const CTxIn &_txin, ptxConflicting->vin)
-                    {
-                        if (_txin.nSequence < std::numeric_limits<unsigned int>::max()-1)
-                        {
-                            fReplacementOptOut = false;
-                            break;
-                        }
+                BOOST_FOREACH(const CTxIn &_txin, ptxConflicting->vin) {
```
instagibbs: this whole section is still about replacement, I think it should be removed.

As well as line 1115?

jtimon: Well, that's done on https://github.com/ElementsProject/elements/pull/249 I will rebase on top of this, just wanted to keep it separated in case we want the other parts but not this one.

instagibbs:
what do you mean "unused?"

jtimon:
> what do you mean "unused?"

Nothing. I mean,, I'm assuming nobody wants to use this in elements, but I should remove that from the commit message.

instagibbs:
concept ACK

I think we should attempt to re-enable `replace-by-fee.py` functional test first, since we are creating additional deltas from upstream. Taking a look now.

jtimon:
Rebased.

instagibbs:
Kind of crazy none of this is covered under tests anyways.

utACK https://github.com/ElementsProject/elements/pull/248/commits/e17e0b81b6c5a60fce5b79c45bbb036fe029cdff

PR #249, opened 2017-08-26
p2p: Remove opt-in part of RBF

jtimon:
We can remove this independently of the removal on the wallet safely. We will break elements expecting the opt-in sequences to mean anything, but this is just a bad assumption.

EDIT: Rebased on top of #248

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1080,35 +1080,8 @@ bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, const C
     BOOST_FOREACH(const CTxIn &txin, tx.vin)
     {
         auto itConflicting = pool.mapNextTx.find(txin.prevout);
-        if (itConflicting != pool.mapNextTx.end())
-        {
-            const CTransaction *ptxConflicting = itConflicting->second;
-            if (!setConflicts.count(ptxConflicting->GetHash()))
-            {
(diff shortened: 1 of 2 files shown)


src/validation.cpp
```diff
@@ -1089,40 +1089,8 @@ bool AcceptToMemoryPoolWorker(CTxMemPool& pool, CValidationState& state, const C
...
-                        {
-                            fReplacementOptOut = false;
-                            break;
-                        }
-                    }
-                }
-                if (fReplacementOptOut)
-                    return state.Invalid(false, REJECT_CONFLICT, "txn-mempool-conflict");
-
-                setConflicts.insert(ptxConflicting->GetHash());
```
instagibbs: I don't think you can just delete these insertions? It's used in following checks.

jtimon: The inserts are kept but indented down.

jtimon:
Rebased on top of https://github.com/ElementsProject/elements/pull/248 but it doesn't really need it.

instagibbs:
utACK

Could you rebase and update the functional tests? This should break something, hopefully.

jtimon:
Rebased.

instagibbs:
could you rebase and remove the first commit now that it is merged? Test fails.

jtimon:
Rebased, removed the part of the test that tests the opt-in.

PR #251, opened 2017-08-28
HF: Replace bip34 with simpler equivalent 

jtimon:
The height is mandatory in the coinbase's nLockTime instead of in its script.
The miner still puts it in the script.
The new rule is active from the beginning instead of a given height, thus there's no need to implement bip30, which is removed.

### DIFF
--- a/contrib/devtools/check-doc.py
+++ b/contrib/devtools/check-doc.py
@@ -23,7 +23,7 @@ REGEX_DOC = re.compile(r'HelpMessageOpt\(\"(\-[^\"=]+?)(?:=|\")')
 # list unsupported, deprecated and duplicate args as they need no documentation
 SET_DOC_OPTIONAL = set(['-rpcssl', '-benchmark', '-h', '-help', '-socks', '-tor', '-debugnet', '-whitelistalwaysrelay', '-prematurewitness', '-walletprematurewitness', '-promiscuousmempoolflags', '-blockminsize'])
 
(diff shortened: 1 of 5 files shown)


instagibbs:
concept ACK, will review

luke-jr:
At least with Bitcoin, height-based locktime must be greater than the current height, so setting this to equal should be invalid? For https://github.com/BitcoinHardfork/bitcoin/pull/3, I set locktime=height-1 because of this.

luke-jr:
Summary of extra-Github discussion:

* You could also change the final check to `<=`, but this would be liable to create confusion between Bitcoin and Elements, so IMO `height - 1` is the better solution.
* Care needs to be taken, especially if removing the extranonce also, that the coinbase is not `<2` bytes. Or remove that rule.
* Removing and/or refactoring code that isn't a problem will make rebasing harder. For this reason, I suggest at least leaving the `pindexPrev` parameter in `IncrementExtraNonce` unused. But whether making rebasing easier is a goal, is a project-wide policy to decide, and should probably be discussed independently from this PR.

jtimon:
Rebased, squashed, hopefully fixed nits.

jtimon:
Rebased on top of https://github.com/ElementsProject/elements/pull/276 and made a little bit more like https://github.com/BitcoinHardfork/bitcoin/pull/3/files as suggested by @luke-jr to reduce rebase disruption, no longer removing bip30 not bip34Height.
Also settled on not removing the extra nonce for now, that can be discussed in another PR.
Instead of consensus.HardforkTime, since we may have many and it's good to separate them, I'll use something like https://github.com/bitcoin/bitcoin/pull/11426 for hardforks.

instagibbs:
Closing due to lack of activity and since we've launched Liquid without this change. Porting to master with backwards compatibility(in chain sense) still could work.

PR #253, opened 2017-08-29
Update trusted-git-root

jtimon:
Replaces #252 which was merged by accident and then removed from history via force push.
The goal is that I learn how to merge signing the merge commit too.


instagibbs:
ACK, will let you merge this one as practice.

PR #255, opened 2017-09-07
fix incorrect bounds check to wit.vtxinwit access in ContextualCheckBlock

arvidn:


src/validation.cpp
```diff
@@ -3888,7 +3888,9 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, const Co
         int commitpos = GetWitnessCommitmentIndex(block);
         if (commitpos != -1) {
             uint256 hashWitness = BlockWitnessMerkleRoot(block);
-            if ((block.vtx[0]->wit.vtxinwit.size() > 0 &&block.vtx[0]->wit.vtxinwit[0].scriptWitness.stack.size() != 1) || block.vtx[0]->wit.vtxinwit[0].scriptWitness.stack[0].size() != 32) {
```
arvidn: if ``block.vtx[0]->wit.vtxinwit.empty()``, the last term in this condition will cause an out-of-bound access to that array.

instagibbs:
Ah, nice catch. One of the hazards of us moving the witness input data back into its own structure with bounds checking. Don't think we can hit this at all due to serialization, but great to get this right. utACK.

arvidn:
I'm not sure how to interpret the travis output. is there a formatting check that's failing?

instagibbs:
kicked Travis, the errors seemed very random.

instagibbs:
"pyenv: python3.6: command not found", might be some travis update we need to support?

instagibbs:
@arvidn could you cherry-pick https://github.com/bitcoin/bitcoin/pull/11271 and see if it passes?

instagibbs:
ACK

PR #256, opened 2017-09-11
work towards bringing mininode.py up to date with the main node

arvidn:
and towards re-enabling rpc tests. This patch introduces ``CProof``, ``CTxOutValue``, ``CTxOutAsset`` and ``CTxOutNonce`` to mininode.py. It also adds some debug output to the main node that was helpful in trouble-shooting the interaction between mininode.py and the main node.

This patch is a step in the direction of having mininode be compatible with elements. it does not bring it nor the tests all the way there to pass.

### DIFF
--- a/qa/rpc-tests/test_framework/blocktools.py
+++ b/qa/rpc-tests/test_framework/blocktools.py
@@ -8,7 +8,7 @@ from .mininode import *
 from .script import CScript, OP_TRUE, OP_CHECKSIG, OP_RETURN
 
 # Create a block (with regtest difficulty)
-def create_block(hashprev, coinbase, nTime=None):
+def create_block(hashprev, coinbase, nTime=None, height=0):
     block = CBlock()
     if nTime is None:
         import time
@@ -16,9 +16,11 @@ def create_block(hashprev, coinbase, nTime=None):
     else:
         block.nTime = nTime
     block.hashPrevBlock = hashprev
-    block.nBits = 0x207fffff # Will break after a difficulty adjustment...
     block.vtx.append(coinbase)
     block.hashMerkleRoot = block.calc_merkle_root()
+    block.nHeight = height
+    block.proof.challenge = CScript([OP_TRUE])
+    block.proof.solution = b''
     block.calc_sha256()
     return block
 
@@ -39,39 +41,28 @@ def add_witness_commitment(block, nonce=0):
 
     # witness commitment is the last OP_RETURN output in coinbase
     output_data = WITNESS_COMMITMENT_HEADER + ser_uint256(witness_commitment)
-    block.vtx[0].vout.append(CTxOut(0, CScript([OP_RETURN, output_data])))
+    block.vtx[0].vout.append(CTxOut(CTxOutValue(0), CScript([OP_RETURN, output_data])))
     block.vtx[0].rehash()
     block.hashMerkleRoot = block.calc_merkle_root()
     block.rehash()
 
-
-def serialize_script_num(value):
-    r = bytearray(0)
-    if value == 0:
-        return r
-    neg = value < 0
-    absvalue = -value if neg else value
-    while (absvalue):
-        r.append(int(absvalue & 0xff))
-        absvalue >>= 8
-    if r[-1] & 0x80:
-        r.append(0x80 if neg else 0)
-    elif neg:
-        r[-1] |= 0x80
-    return r
-
 # Create a coinbase transaction, assuming no miner fees.
 # If pubkey is passed in, the coinbase output will be a P2PK output;
 # otherwise an anyone-can-spend output.
-def create_coinbase(height, pubkey = None):
+def create_coinbase(height, pubkey = None, amount = 0):
     coinbase = CTransaction()
-    coinbase.vin.append(CTxIn(COutPoint(0, 0xffffffff), 
-                ser_string(serialize_script_num(height)), 0xffffffff))
+    # coinbase transaction scriptsigs must be at least 2 bytes
+    coinbase.vin.append(CTxIn(COutPoint(0, 0xffffffff),
+        CScript([height, OP_TRUE]), 0xffffffff))
     coinbaseoutput = CTxOut()
-    coinbaseoutput.nValue = 50 * COIN
-    halvings = int(height/150) # regtest
-    coinbaseoutput.nValue >>= halvings
-    if (pubkey != None):
+
+    coinbaseoutput.nValue.setToAmount(amount)
+    coinbaseoutput.nAsset.setToAsset(b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
+
+    if amount == 0:
+        # zero utxo's must be made unspendable
+        coinbaseoutput.scriptPubKey = CScript([OP_RETURN])
+    elif pubkey != None:
         coinbaseoutput.scriptPubKey = CScript([pubkey, OP_CHECKSIG])
     else:
         coinbaseoutput.scriptPubKey = CScript([OP_TRUE])
@@ -85,7 +76,7 @@ def create_transaction(prevtx, n, sig, value, scriptPubKey=CScript()):
     tx = CTransaction()
     assert(n < len(prevtx.vout))
     tx.vin.append(CTxIn(COutPoint(prevtx.sha256, n), sig, 0xffffffff))
-    tx.vout.append(CTxOut(value, scriptPubKey))
+    tx.vout.append(CTxOut(CTxOutValue(value), scriptPubKey))
     tx.calc_sha256()
     return tx
 
--- a/qa/rpc-tests/test_framework/mininode.py
+++ b/qa/rpc-tests/test_framework/mininode.py
@@ -359,57 +359,119 @@ class CTxIn(object):
             % (repr(self.prevout), bytes_to_hex_str(self.scriptSig),
                self.nSequence)
 
-def setCTxOutValue(amount):
-    vchCommitment = [0]*33
-    for i in range(8): #8 bytes
-        vchCommitment[33-1-i] = ((amount >> (i*8)) & 0xff)
-    return ''.join(map(chr, vchCommitment))
-
-#Serialization only does value commitment
-class CTxOutValue(object):
-    def __init__(self, vchCommitment=[], vchRangeProof="", vchNonceCommitment=""):
+class CTxOutAsset(object):
+    def __init__(self, vchCommitment=b"\x00"):
         self.vchCommitment = vchCommitment
-        self.vchRangeProof = vchRangeProof
-        self.vchNonceCommitment = vchNonceCommitment
 
     def deserialize(self, f):
-        self.vchCommitment = f.read(33)
-        #self.vchRangeProof = deser_string(f)
-        #self.vchNonceCommitment = deser_string(f)
+        version = ord(f.read(1))
+        if version == 0: self.vchCommitment = b'\x00'
+        elif version == 1: self.vchCommitment = b'\x01' + f.read(32)
+        elif version == 0xff: self.vchCommitment = b'\xff' + f.read(32)
+        elif version == 10 or version == 11: self.vchCommitment = bytes([version]) + f.read(32)
+        else: raise 'invalid CTxOutAsset in deserialize'
 
     def serialize(self):
         r = b""
         r += self.vchCommitment
-        #r += ''.join(map(chr, self.vchCommitment))
-        #r += ser_string(self.vchRangeProof)
-        #r += ser_string(self.vchNonceCommitment)
+        return r
+
+    def setToAsset(self, val):
+        if len(val) != 32:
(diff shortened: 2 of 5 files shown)


qa/rpc-tests/test_framework/blocktools.py
```diff
@@ -39,15 +41,19 @@ def add_witness_commitment(block, nonce=0):
...
 
 
 def serialize_script_num(value):
-    r = bytearray(0)
     if value == 0:
+        return b'\x00'
+    # short form OP_1 - OP_16
+    r = bytearray()
+    if value > 0 and value < 17:
+        r.append(0x50 + value)
```
arvidn: previously this function would not produce the canonical encoding for values [1, 16]

qa/rpc-tests/test_framework/blocktools.py
```diff
@@ -58,20 +64,26 @@ def serialize_script_num(value):
         r.append(0x80 if neg else 0)
     elif neg:
         r[-1] |= 0x80
+    r.insert(0, len(r))
```
arvidn: it also would not include the length-prefix

qa/rpc-tests/test_framework/blocktools.py
```diff
@@ -58,20 +64,26 @@ def serialize_script_num(value):
...
+    coinbase.vin.append(CTxIn(COutPoint(0, 0xffffffff),
+                serialize_script_num(height) + b'\x51', 0xffffffff))
     coinbaseoutput = CTxOut()
-    coinbaseoutput.nValue = 50 * COIN
-    halvings = int(height/150) # regtest
-    coinbaseoutput.nValue >>= halvings
-    if (pubkey != None):
+
+    coinbaseoutput.nValue.setToAmount(amount)
+    coinbaseoutput.nAsset.setToAsset(b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
```
arvidn: I'm not sure what the most reasonable default value for asset is. Should it be another parameter?

qa/rpc-tests/test_framework/blocktools.py
```diff
@@ -58,20 +64,26 @@ def serialize_script_num(value):
...
 
 # Create a coinbase transaction, assuming no miner fees.
 # If pubkey is passed in, the coinbase output will be a P2PK output;
 # otherwise an anyone-can-spend output.
-def create_coinbase(height, pubkey = None):
+def create_coinbase(height, pubkey = None, amount = 0):
     coinbase = CTransaction()
-    coinbase.vin.append(CTxIn(COutPoint(0, 0xffffffff), 
-                ser_string(serialize_script_num(height)), 0xffffffff))
+    # coinbase transaction scriptsigs must be at least 2 bytes
```
arvidn: this comment refers to [this](https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/validation.cpp#L534) condition. Come to think of it, perhaps that check is incorrect. It seems to assume that the height at the start of the script must use a two-byte encoding (which is only true for values > 16)

qa/rpc-tests/test_framework/blocktools.py
```diff
@@ -58,20 +64,26 @@ def serialize_script_num(value):
...
-    coinbaseoutput.nValue = 50 * COIN
-    halvings = int(height/150) # regtest
-    coinbaseoutput.nValue >>= halvings
-    if (pubkey != None):
+
+    coinbaseoutput.nValue.setToAmount(amount)
+    coinbaseoutput.nAsset.setToAsset(b'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa')
+
+    if amount == 0:
+        # zero utxo's must be made unspendable
```
arvidn: this comment refers to [this](https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/validation.cpp#L1007) condition.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -359,57 +359,131 @@ def __repr__(self):
             % (repr(self.prevout), bytes_to_hex_str(self.scriptSig),
                self.nSequence)
 
-def setCTxOutValue(amount):
-    vchCommitment = [0]*33
-    for i in range(8): #8 bytes
-        vchCommitment[33-1-i] = ((amount >> (i*8)) & 0xff)
-    return ''.join(map(chr, vchCommitment))
+class CTxOutAsset(object):
```
arvidn: the C++ version of these 3 classes are a single template. Perhaps these should be merged as well and use variables instead of template parameters.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -558,38 +624,6 @@ def serialize_with_witness(self):
         r += struct.pack("<I", self.nLockTime)
         return r
 
-    # Only serialize with witness when explicitly called for
-    def serialize_with_witness(self):
```
arvidn: the ``serialize_with_witness`` function was duplicated. This removes the second copy as the first one matched the C++ code better

instagibbs:
seems you left some random net_processing.cpp changes, please remove.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -359,57 +359,131 @@ def __repr__(self):
...
+            ret <<= 8
+            ret |= self.vchCommitment[i+1]
+        return ret
+
+    def __repr__(self):
+        return "CTxOutValue(vchCommitment=%s vchRangeProof=%s, vchnonceCommitment=%s)" \
+            % (self.vchCommitment, self.vchRangeProof,
+                self.vchNonceCommitment)
+
+class CTxOutNonce(object):
```
instagibbs: Nonce class only has a 33-byte commitment, not a rangeproof or other witness data.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -359,57 +359,131 @@ def __repr__(self):
...
 class CTxOutValue(object):
-    def __init__(self, vchCommitment=[], vchRangeProof="", vchNonceCommitment=""):
+
+    def __init__(self, value=None):
+        self.setNull()
+        if value != None: self.setToAmount(value)
+
+    def setNull(self):
+        self.vchCommitment = b'\x00'
+        self.vchRangeProof =b''
```
instagibbs: rangeproof no longer resides in the output. It resides in the `CTxWitness` based field `wit` in `CTransaction`

qa/rpc-tests/test_framework/blocktools.py
instagibbs: It's likely going the be whatever `policyAsset` is set to, in most cases "bitcoin"'s asset type.

qa/rpc-tests/test_framework/blocktools.py
instagibbs: Not a script expert, but these values correspond to opcodes, not the resulting value pushed on the stack.

The code as-is matches the C++ code quite well.

qa/rpc-tests/test_framework/blocktools.py
instagibbs: Again, I don't think is required. If someone is pushing the number onto the stack, they will append this value.

qa/rpc-tests/test_framework/blocktools.py
instagibbs: is this upstreamable?

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -359,57 +359,131 @@ def __repr__(self):
...
 
-def setCTxOutValue(amount):
-    vchCommitment = [0]*33
-    for i in range(8): #8 bytes
-        vchCommitment[33-1-i] = ((amount >> (i*8)) & 0xff)
-    return ''.join(map(chr, vchCommitment))
+class CTxOutAsset(object):
+    def __init__(self, vchCommitment=b"\x00", vchRangeProof=b"", vchNonceCommitment=b""):
+        self.vchCommitment = vchCommitment
+        self.vchRangeProof = vchRangeProof
```
instagibbs: Rangeproof is in witness structure.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -359,57 +359,131 @@ def __repr__(self):
...
+        self.vchNonceCommitment = vchNonceCommitment
```
instagibbs: it's not a nonce, just an asset commitment

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -359,57 +359,131 @@ def __repr__(self):
...
-    def __init__(self, vchCommitment=[], vchRangeProof="", vchNonceCommitment=""):
+
+    def __init__(self, value=None):
+        self.setNull()
+        if value != None: self.setToAmount(value)
+
+    def setNull(self):
+        self.vchCommitment = b'\x00'
+        self.vchRangeProof =b''
+        self.vchNonceCommitment = b''
```
instagibbs: value, not a nonce

qa/rpc-tests/test_framework/blocktools.py
arvidn: this function is used in exactly one place (line 69 to the left). Its result is passed to ``ser_string()`` and passed in as a script. Perhaps I'm missing something, but it's not clear to me how that could possibly generate a valid script (apart from the special case of 0 and possibly some others). [This](https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/validation.cpp#L3872) check was failing in the test.

arvidn: that said, it's probably better to leave this function as-is, and instead wrap it in ``CScript()`` where it's used instead

qa/rpc-tests/test_framework/mininode.py
arvidn: will fix

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -429,28 +491,59 @@ def is_null(self):
...
     def is_null(self):
         return self.scriptWitness.is_null()
 
+class CTxOutWitness(object):
+    def __init__(self):
+        self.vchSurjectionproof = b'';
+        self.vchRangeproof = b'';
+
+    def deserialize(self, f):
+        self.vchIssuanceAmountRangeproof = deser_string(f)
```
instagibbs: copy and paste error, these should be range/surj proofs

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -429,28 +491,59 @@ def is_null(self):
...
+        self.vchInflationKeysRangeproof = deser_string(f)
+
+    def serialize(self):
+        r = b''
+        r += ser_string(self.vchIssuanceAmountRangeproof)
```
instagibbs: ditto

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -429,28 +491,59 @@ def is_null(self):
...
+        r += ser_string(self.vchInflationKeysRangeproof)
+        return r
+
+    def __repr__(self):
+        return repr(self.scriptWitness)
```
instagibbs: scriptWitness is for inputs

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -429,28 +491,59 @@ def is_null(self):
...
+
+    def is_null(self):
+        return self.scriptWitness.is_null()
```
instagibbs: scriptWitness is for inputs

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -429,28 +491,59 @@ def is_null(self):
...
     def serialize(self):
-        return ser_string_vector(self.scriptWitness.stack)
+        r = b''
+        r += ser_string(self.vchIssuanceAmountRangeproof)
+        r += ser_string(self.vchInflationKeysRangeproof)
+        r += ser_string_vector(self.scriptWitness.stack)
+        return r
 
     def __repr__(self):
         return repr(self.scriptWitness)
```
instagibbs: scriptWitness is for inputs (below as well)

qa/rpc-tests/test_framework/blocktools.py
arvidn: you mean the comment or the patch to make the scriptsig at least 2 bytes? I have a feeling that this requirement of the scriptsig being at least 2 bytes is a mistake. But I can't say I understand the motivation of it well enough to be confident. I imagine it's not a problem in the wild, as the block height most likely will get encoded as at least two bytes in the script. But in tests where the height < 16, it is.

instagibbs:
did some work on top: https://github.com/instagibbs/elements/commits/mininode

re-activated a test that uses mininode transaction serialization(without any witness data).

instagibbs:
tACK https://github.com/ElementsProject/elements/pull/256/commits/3c21411c67a2c23932ed8d8f12c25db0c621f2e7, got the entire test working.

PR #257, opened 2017-09-11
Speed up peg-in validation

jonasnick:


instagibbs:
This results in an order of magnitude speedup on checking the pegin transactions, according to @jonasnick .

instagibbs:
Pretty sure the `getblockhash` check is completely redundant as well. We now do a startup check to ensure we're querying the right bitcoind instance.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/257/commits/25c8a49b322152486eeea09213ca0eb5463f1b9e

jtimon:
Mhmm, it seems it genesishash was passed but never used?
utACK 25c8a49b322152486eeea09213ca0eb5463f1b9e

instagibbs:
@jtimon was used to check the result of the first call, but it's duplication of effort as we do this check on init.

PR #262, opened 2017-09-20
Remove CHAINPARAMS_ELEMENTS 

jtimon:
CHAINPARAMS_OLD_MAIN is meant to be used only for specific tests depending on base58, not as default for all unittests through CBaseChainParams::MAIN.
But that's not the reality, more tests are currenly dependent on CBaseChainParams::MAIN being equal to  CHAINPARAMS_OLD_MAIN or CHAINPARAMS_ELEMENTS 
So removing the constant CHAINPARAMS_ELEMENTS is moving forwards.

Also:
- [X] no point in CBaseChainParams::CUSTOM.
- [X] rename "main" to "old_main", just in case.
- [X] Chainparams: Custom: Remove -chainpetname like in upstream #8994 aftersome review
- [X] Remove CHAINPARAMS_ELEMENTS 

Dependencies:

- [x] Chainparams: decouple anyonecanspend are mine from regtest #270 
- [x] QA: Modernize confidential_transactions.py #275
~~- [ ] QA: Chainparams: Run functional tests with custom chain #274~~

### DIFF
--- a/src/rpc/protocol.cpp
+++ b/src/rpc/protocol.cpp
@@ -79,7 +79,6 @@ boost::filesystem::path GetAuthCookieFile()
 boost::filesystem::path GetMainchainAuthCookieFile()
 {
     boost::filesystem::path path(GetArg("-mainchainrpccookiefile", COOKIEAUTH_FILE));
-    if (!path.is_complete() && BaseParams().DataDir() == CHAINPARAMS_ELEMENTS) path = "testnet3" / path;
     if (!path.is_complete() && BaseParams().DataDir() == CHAINPARAMS_REGTEST) path = "regtest" / path;
     if (!path.is_complete()) path = GetDataDir(false) / path;
     return path;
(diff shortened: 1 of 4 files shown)


jtimon:
Although this can be done independently, rebased on top of  https://github.com/ElementsProject/elements/pull/274 since they need some common changes.

jtimon:
Decoupled from #274 since that is currently failing replace-by-fee.py

src/rpc/protocol.cpp
```diff
@@ -79,8 +79,6 @@ boost::filesystem::path GetAuthCookieFile()
 boost::filesystem::path GetMainchainAuthCookieFile()
 {
     boost::filesystem::path path(GetArg("-mainchainrpccookiefile", COOKIEAUTH_FILE));
-    if (!path.is_complete() && BaseParams().DataDir() == CHAINPARAMS_ELEMENTS) path = "testnet3" / path;
-    if (!path.is_complete() && BaseParams().DataDir() == CHAINPARAMS_REGTEST) path = "regtest" / path;
```
instagibbs: wouldn't this break connecting via rpc cookie auth?

instagibbs:
concept ACK, but think this introduces a regression for rpc cookie auth.

src/rpc/protocol.cpp
jtimon: They should just put it in elementsregtest, no?

instagibbs: this is specifically looking for the `bitcoind` cookie, not elementsd.

jtimon:
Fixed nit. Rebased

instagibbs:
If this is really only used for base58 tests, can we remove more? Or was your point that we *are* using it for more general tests?

jtimon:
Yes, I intend to go further and remove more, see #266 which still doesn't remove CMainParams because apart from being used for base58 unittests it is also used for bip9 unittests (using consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY], consensus.nRuleChangeActivationThreshold and consensus.nMinerConfirmationWindow).

Perhaps we can get rid of those 2 tests and CMainParams altogether but this is a step forward, I think.

instagibbs:
utACK

jtimon:
Merged

PR #265, opened 2017-09-21
Remove chainparamsseeds.h

jtimon:
Remove that warning first of all.
But also I think some times the best way to make rebases easy is not to minimize the diff but make clear which part you don't care about with the commit. As in removing that part and stop testing it if you were in the first place.

Removes a warning we're getting consistently.


jtimon:
Sorry for the auto-merge, but this is really simple and I just didn't want to keep looking at the warning.

PR #269, opened 2017-09-21
include the CTxWitness structure in memory usage accounting

arvidn:
~~Also fixed comment in memusage.h~~

### DIFF
--- a/src/core_memusage.h
+++ b/src/core_memusage.h
@@ -26,27 +26,49 @@ static inline size_t RecursiveDynamicUsage(const CTxOut& out) {
     return RecursiveDynamicUsage(out.scriptPubKey);
 }
 
-//TODO Account for CTxWitness.
+static inline size_t RecursiveDynamicUsage(const ::CTxInWitness& wit) {
+    return memusage::DynamicUsage(wit.vchIssuanceAmountRangeproof)
+        + memusage::DynamicUsage(wit.vchInflationKeysRangeproof);
+}
+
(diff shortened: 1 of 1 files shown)


src/memusage.h
```diff
@@ -38,8 +38,9 @@ template<typename X> static inline size_t DynamicUsage(X * const &v) { return 0;
 template<typename X> static inline size_t DynamicUsage(const X * const &v) { return 0; }
 
 /** Compute the memory used for dynamically allocated but owned data structures.
- *  For generic data types, this is *not* recursive. DynamicUsage(vector<vector<int> >)
- *  will compute the memory used for the vector<int>'s, but not for the ints inside.
```
arvidn: I think this comment is misleading at best. It *does* account for the array of ints ("the ints").

arvidn: oh, I just realized the example had nested vectors. allright, I'll revert this

src/core_memusage.h
```diff
@@ -26,27 +26,49 @@ static inline size_t RecursiveDynamicUsage(const CTxOut& out) {
     return RecursiveDynamicUsage(out.scriptPubKey);
 }
 
-//TODO Account for CTxWitness.
+static inline size_t RecursiveDynamicUsage(const CTxInWitness& wit) {
```
arvidn: all these ``static`` are redundant, I copied them for symmetry, though they should probably be removed.

PR #270, opened 2017-09-23
Chainparams: decouple anyonecanspend are mine from regtest

jtimon:
Spending op_true outputs has more to do with the wallet than with the chain. Perhaps this options should be per wallet in the multiwallet future?

Alternatively we can make it a field in CChainParams, but ```Params().NetworkIDString() == CHAINPARAMS_REGTEST``` defeats the whole point of having chainparams IMO.

### DIFF
--- a/contrib/devtools/check-doc.py
+++ b/contrib/devtools/check-doc.py
@@ -21,8 +21,9 @@ CMD_GREP_DOCS = r"egrep -r -I 'HelpMessageOpt\(\"\-[^\"=]+?(=|\")' %s" % (CMD_RO
 REGEX_ARG = re.compile(r'(?:map(?:Multi)?Args(?:\.count\(|\[)|Get(?:Bool)?Arg\()\"(\-[^\"]+?)\"')
 REGEX_DOC = re.compile(r'HelpMessageOpt\(\"(\-[^\"=]+?)(?:=|\")')
 # list unsupported, deprecated and duplicate args as they need no documentation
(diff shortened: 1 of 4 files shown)


src/script/ismine.h
```diff
@@ -13,6 +13,8 @@
 class CKeyStore;
 class CScript;
 
+static const bool DEFAULT_ANYONECANSPEND_AREMINE = true;
```
instagibbs: imo this should be false by default. It's a hack for regtest, for the most part.

src/script/ismine.h
```diff
@@ -28,6 +30,9 @@ enum isminetype
 /** used for bitflags of isminetype */
 typedef uint8_t isminefilter;
 
+// TODO support for multiwallet ?
+extern bool g_anyonecanspend_aremine;
```
instagibbs: Not sure this needs multiwallet support. It's for testing only.

instagibbs:
would be easy to add a basic test; please do

src/script/ismine.h
jtimon: I'll remove the comment, sorry.

src/script/ismine.h
jtimon: Mhmm, my thought was that custom/regtest is the only chain that we support directly...perhaps it should be a field in CChainParams instead?
The important part is removing ```Params().NetworkIDString() == CHAINPARAMS_REGTEST```

jtimon:
Updated making a new field in CChainParams instead of an independent global.
Now it requires the custom chain to be configurable, so the test is in the last commit of https://github.com/ElementsProject/elements/pull/274

instagibbs:
utACK

PR #271, opened 2017-09-23
RPC: getinfo: Replace testnet with more generic chain

jtimon:
As in #270, ```Params().NetworkIDString() == CHAINPARAMS_REGTEST``` defeats the whole point of having chainparams IMO.

Note getinfo gets removed upstream and getblockchaininfo and getmininginfo already do this.


instagibbs:
concept ACK.

What's the alternative way to get this information? `getinfo` is completely removed in Bitcoin Core master. I can't seem to find the same information.

instagibbs:
utACK

PR #275, opened 2017-09-28
QA: Modernize confidential_transactions.py

jtimon:
Small changes to do things more like they're done upstream in confidential_transactions.py


instagibbs:
utACK

PR #276, opened 2017-09-30
Optimization: Remove Consensus::Params::BIP34Hash

jtimon:
Like upstream https://github.com/bitcoin/bitcoin/pull/11427

#251 needs less disruption after this (no need to remove bip30's code anymore).

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2587,10 +2587,7 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // time BIP34 activated, in each of the existing pairs the duplicate coinbase had overwritten the first
     // before the first had been spent.  Since those coinbases are sufficiently buried its no longer possible to create further
     // duplicate transactions descending from the known pairs either.
-    // If we're on the known chain at height greater than where BIP34 activated, we can save the db accesses needed for the BIP30 check.
-    CBlockIndex *pindexBIP34height = pindex->pprev->GetAncestor(chainparams.GetConsensus().BIP34Height);
-    //Only continue to enforce if we're below BIP34 activation height or the block hash at that height doesn't correspond.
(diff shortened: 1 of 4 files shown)


jtimon:
I think we should merge this even if it's rejected upstream or fully remove bip30 instead as discussed in #251 (although that option is slightly harder to rebase).

src/validation.cpp
```diff
@@ -2587,10 +2587,7 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
     // time BIP34 activated, in each of the existing pairs the duplicate coinbase had overwritten the first
     // before the first had been spent.  Since those coinbases are sufficiently buried its no longer possible to create further
     // duplicate transactions descending from the known pairs either.
-    // If we're on the known chain at height greater than where BIP34 activated, we can save the db accesses needed for the BIP30 check.
-    CBlockIndex *pindexBIP34height = pindex->pprev->GetAncestor(chainparams.GetConsensus().BIP34Height);
-    //Only continue to enforce if we're below BIP34 activation height or the block hash at that height doesn't correspond.
-    fEnforceBIP30 = fEnforceBIP30 && (!pindexBIP34height || !(pindexBIP34height->GetBlockHash() == chainparams.GetConsensus().BIP34Hash));
+    fEnforceBIP30 = fEnforceBIP30 && pindex->nHeight > chainparams.GetConsensus().BIP34Height;
```
instagibbs: I think you've got the logic backwards here. We don't need BIP30 at all if we have BIP34 activated(which we do by block 1 IIRC)

jtimon: oops, yeah, corrected the nit upstream but not here.

instagibbs:
Looks like upstream PR was abandoned, closing.

PR #277, opened 2017-09-30
BIP90: Make buried deployments slightly more easily extensible

jtimon:
Similar to https://github.com/bitcoin/bitcoin/pull/11426 but based on Based on https://github.com/ElementsProject/elements/pull/276

EDIT: while also:

- BIP90: Restore rebase compatibility
- Elements: BIP90: Enforce all buried deployments from block 0
- Also backport configurable bip16 ( https://github.com/bitcoin/bitcoin/pull/11430 )


jtimon:
@luke-jr perhaps for more easier rebases with respect to bitcoin deployments we should start here, restoring some of it while still getting the upgrades from block 0 for elements chain (one can always opt out of good upgrades with this too).

Also see https://github.com/bitcoin/bitcoin/pull/11430 for a potential backport.

jtimon:
Let's not merge this if it isn't merged upstream to simplify rebases.

jtimon:
close like https://github.com/bitcoin/bitcoin/pull/11426

PR #286, opened 2017-11-06
2WP: -mainchainrpccookiefile defaults to regtest/.cookie for all chains

jtimon:
Decouple GetMainchainAuthCookieFile() from the selected chain's datadir.

### DIFF
--- a/qa/rpc-tests/pegging.py
+++ b/qa/rpc-tests/pegging.py
@@ -50,28 +50,44 @@ sidechain_datadir = get_temp_dir('sidechain')
 sidechain_pass = get_pseudorandom_str()
 sidechain2_datadir = get_temp_dir('sidechain2')
 sidechain2_pass = get_pseudorandom_str()
+bitcoin2_datadir = get_temp_dir('bitcoin2')
+bitcoin2_rpccookiefile = bitcoin2_datadir + '/regtest/.cookie'
 
 bitcoin_port = 8000 + os.getpid()%999
 sidechain_port = bitcoin_port + 1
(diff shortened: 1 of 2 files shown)


instagibbs:
how do you see this scaling when we have more networks built in?

jtimon:
My preferred option would be to make -mainchainrpccookiefile mandatory if you're validating pegs or simply let it at regtest.
Another option would be to introduce a default_mainchainrpccookiefile to the configurable chainparams.
Ideally support for more networks should be simply provided by configuring the custom/regtest chainparams.

jtimon:
Rebased

instagibbs:
request: in `pegging.py`, have one of the two elementsd use the option instead of rpcuser/password. Manually testing now.

instagibbs:
ah hm, can't figure out how to get AuthProxy to work with cookie auth for pegging.py. Manual testing showed a related bug: https://github.com/ElementsProject/elements/pull/292

tACK, but I'd like to get some tests in too if at all possible

jtimon:
The bitcoin node either uses has user/pass defined or it creates the cookie file.
I've changed pegging so that -mainchainrpccookiefile is tested rather than -mainchainrpcpassword and -mainchainrpcuser.
Does that make sense?

instagibbs:
Could we test both? Spin up a second bitcoind, check that it accepts the block with a pegged transaction?

jtimon:
Done, now each sidechain node connects to each own bitcoin node, which also connect between them.
I can squash the last 3 commits if you want.

instagibbs:
utACK, don't mind the commits

PR #295, opened 2017-12-12
Fix mainchain host arg name from mainchainhost to mainchainrpchost

azuchi:
In the help message and the description of the [project site](https://www.elementsproject.org/sidechains/creating-your-own.html), the argument name which specify the host of the mainchain is `-mainchainrpchost`, but in the implementation it is `-mainchainhost`.

So change it from `-mainchainhost` to `-mainchainrpchost`.


instagibbs:
ACK

PR #301, opened 2018-01-22
In pegging.py test use sidechain binary in the same repo per default

jonasnick:


instagibbs:
this assumes I run it in the rpc-tests directory, yes? I normally don't...

jonasnick:
No. This is the same line of code that is executed to find the relevant binary for any other script in the rpc-tests dir.

instagibbs:
oh I see, only relative to the pegging test itself. Will test.

PR #266, opened 2017-09-21
Chainparams: Remove CElementsParams

jtimon:
Dependencies:

- [x] Chainparams: decouple anyonecanspend are mine from regtest #270 
- [x] QA: Modernize confidential_transactions.py #275
~~- [ ] QA: Chainparams: Run functional tests with custom chain #274~~
- [x] Remove CHAINPARAMS_ELEMENTS #262
~~- [ ] Chainparams: Remove CRegTestParams and use CCustomParams instead #263~~
~~- [ ] RPC: getinfo: Replace testnet with more generic chain #271~~

~~- [ ] Chainparams: Every chain should have a name #307~~

PR #306, opened 2018-02-17
Chainparams: Unify CRegTestParams and CCustomParams

jtimon:
Any class extending CChainParams directly should be fine.
Any class extending CRegTestParams should be fine, just extend CCustomParams instead (or just use the custom params if that's enough for your chain).

Dependencies:

~~- [ ] Chainparams: Every chain should have a name #307~~ 
- [x] Chainparams: Remove CElementsParams #266


jtimon:
replace-by-fee.py was failing because fRequireStandard was true by default instead of false, updated

instagibbs:
merged

PR #307, opened 2018-02-17
Chainparams: Every chain has a name

jtimon:
Every chain should have a name (aka strNetworkID), but CElementsParams is not needed.
CMainParams can simply extend CRegTestParams instead of CElementsParams.


jtimon:
Sorry, closing this, there's not need to have this pr separated even if it's shared.

PR #317, opened 2018-03-13
Tutorial: Use sendtomainchain

jtimon:
sendtomainaddress was replaced with sendtomainchain.
I didn't test it following the tutorial.


instagibbs:
ACK, don't even think it was ever called that....

PR #318, opened 2018-03-14
RPC: If a scriptPubKey is a pegout, show pegout info too

jtimon:
This is to give more rich info on RPC calls that return info about scriptPubKey's in outputs if the scriptPubKey is a pegout.

~~Needs tests.~~

### DIFF
--- a/qa/rpc-tests/pegging.py
+++ b/qa/rpc-tests/pegging.py
@@ -88,6 +88,8 @@ def write_bitcoin_conf(datadir, rpcport, rpcpass=None, p2p_port=None, connect_po
         f.write("testnet=0\n")
         f.write("txindex=1\n")
         f.write("daemon=1\n")
+        # To make sure bitcoind gives back p2pkh no matter version
+        f.write("addresstype=legacy\n")
         if connect_port:
             f.write("connect=localhost:"+str(connect_port)+"\n")
             f.write("listen=1\n")
@@ -135,6 +137,23 @@ with open(os.path.join(sidechain2_datadir, "elements.conf"), 'w') as f:
         f.write("connect=localhost:"+str(sidechain1_p2p_port)+"\n")
         f.write("listen=1\n")
         f.write("fallbackfee=0.0001\n")
+
+def test_pegout(parent_chain_addr, sidechain):
+    pegout_txid = sidechain.sendtomainchain(parent_chain_addr, 1)
+    raw_pegout = sidechain.getrawtransaction(pegout_txid, True)
+    assert 'vout' in raw_pegout and len(raw_pegout['vout']) > 0
+    pegout_tested = False
+    for output in raw_pegout['vout']:
+        scriptPubKey = output['scriptPubKey']
+        if 'type' in scriptPubKey and scriptPubKey['type'] == 'nulldata':
+            assert ('pegout_hex' in scriptPubKey and 'pegout_asm' in scriptPubKey and 'pegout_type' in scriptPubKey and
+                    'pegout_chain' in scriptPubKey and 'pegout_reqSigs' in scriptPubKey and 'pegout_addresses' in scriptPubKey)
+            assert scriptPubKey['pegout_chain'] == '0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206' #testnet3
+            assert scriptPubKey['pegout_reqSigs'] == 1
+            assert parent_chain_addr in scriptPubKey['pegout_addresses']
+            pegout_tested = True
+            break
+    assert pegout_tested
 
 try:
 
@@ -255,7 +274,7 @@ try:
         raise Exception("Peg-in should be back to 6 confirms.")
 
     # Do many claims in mempool
-    n_claims = 100
+    n_claims = 5
 
     print("Flooding mempool with many small claims")
     pegtxs = []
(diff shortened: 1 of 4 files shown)


jtimon:
Added tests, it seems the line commented in the new tests is failing due to the parent chain using different base58 prefixes from the host chain, unless I'm missing something.

jtimon:
Fixed code and uncommented failing test.

jtimon:
Needed rebase after #319 's merge

instagibbs:
travis pooping the bed again, restarted jobs

src/rpc/rawtransaction.cpp
```diff
@@ -63,28 +63,48 @@ class RPCRawTransaction_ECC_Init {
...
 
-    out.push_back(Pair("asm", ScriptToAsmStr(scriptPubKey)));
+    out.push_back(Pair(prefix + "asm", ScriptToAsmStr(scriptPubKey)));
     if (fIncludeHex)
-        out.push_back(Pair("hex", HexStr(scriptPubKey.begin(), scriptPubKey.end())));
+        out.push_back(Pair(prefix + "hex", HexStr(scriptPubKey.begin(), scriptPubKey.end())));
 
     if (!ExtractDestinations(scriptPubKey, type, addresses, nRequired)) {
-        out.push_back(Pair("type", GetTxnOutputType(type)));
+        out.push_back(Pair(prefix + "type", GetTxnOutputType(type)));
```
instagibbs: there could be an odd case where this line is hit, yet it still claims return type of P2PK, where the pubkey is invalid... that's kind of odd. I'll check upstream.

instagibbs:
Could we get expanded coverage for other address type pegout(p2sh?), and perhaps unknown pegout script? 

I believe these can be hand-crafted with `createrawtransaction` with output `data` argument.

Looks good otherwise

jtimon:
Corrected auto-nit, added 2 more tests (segwit [p2sh] and garbage)

qa/rpc-tests/pegging.py
```diff
@@ -271,6 +289,23 @@ def write_bitcoin_conf(datadir, rpcport, rpcpass=None, p2p_port=None, connect_po
...
+
+    print("Test pegout Segwit (P2SH of witness script)")
+    parent_chain_addr = bitcoin.getnewaddress()
+    parent_chain_sw_addr = bitcoin.addwitnessaddress(parent_chain_addr)
+    test_pegout(parent_chain_sw_addr, sidechain)
+
+    print("Test pegout Garbage")
+    parent_chain_addr = "garbage"
+    try:
+        test_pegout(parent_chain_addr, sidechain)
```
instagibbs: sorry, wasn't being clear enough here, I want the transaction to "pass", but for the output introspection of it to behave how we expect

instagibbs: ie handcraft using rawtx api, then introspect

jtimon:
pushed one more test with @instagibbs 's clarification.

qa/rpc-tests/pegging.py
```diff
@@ -271,6 +288,46 @@ def write_bitcoin_conf(datadir, rpcport, rpcpass=None, p2p_port=None, connect_po
...
+    raw_pegout = sidechain.decoderawtransaction(rawtx)
+    
+    assert 'vout' in raw_pegout and len(raw_pegout['vout']) > 0
+    pegout_tested = False
+    for output in raw_pegout['vout']:
+        scriptPubKey = output['scriptPubKey']
+        if 'type' in scriptPubKey and scriptPubKey['type'] == 'nulldata':
+            assert ('pegout_hex' in scriptPubKey and 'pegout_asm' in scriptPubKey and 'pegout_type' in scriptPubKey and
+                    'pegout_chain' in scriptPubKey and 'pegout_reqSigs' not in scriptPubKey and 'pegout_addresses' not in scriptPubKey)
+            assert scriptPubKey['pegout_chain'] == 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa01f4'
```
instagibbs: why is there `01f4` at the end? Looks like you made 500 bytes worth of `aa` and the first 32 should just be `aa`?

jtimon: yeah, I think it has to do with the conversion to int256_t, but is worth investigating more.

jtimon: The 01f4 prefix seems to change if more data is pushed. It seems to have to do with the pushdata ops needed depending on the size of the data.

jtimon:
Pushed a better version of the latest test with what we discussed on chat.
I guess I should remove the earlier version and squash, unless we want to keep investigating your latest comment.

instagibbs:
first glance looks good, squash and I can review

qa/rpc-tests/pegging.py
```diff
@@ -271,6 +288,47 @@ def write_bitcoin_conf(datadir, rpcport, rpcpass=None, p2p_port=None, connect_po
         if "confirmations" not in tx or tx["confirmations"] == 0:
             raise Exception("Peg-in confirmation has failed.")
 
+    print("Test pegout")
+    test_pegout(bitcoin.getnewaddress(), sidechain)
+
+    print("Test pegout Segwit (P2SH of witness script)")
+    parent_chain_addr = bitcoin.getnewaddress()
+    parent_chain_sw_addr = bitcoin.addwitnessaddress(parent_chain_addr)
```
instagibbs: oops, this is deprecated in 0.16 and will be removed in 0.17.

In fact, getnewaddress will by default return a p2sh-p2wpkh address.

We could bump the mandatory version to 0.16 for this test, and explicitly set address types in the getnewaddress call.

Or we could set the default address at startup, older nodes will ignore it, and just make a 1-of-1 multisig addr that is p2sh and test that.

instagibbs:
diff --git a/qa/rpc-tests/pegging.py b/qa/rpc-tests/pegging.py
index 5bc318d..ea959e6 100755
--- a/qa/rpc-tests/pegging.py
+++ b/qa/rpc-tests/pegging.py
@@ -88,6 +88,8 @@ def write_bitcoin_conf(datadir, rpcport, rpcpass=None, p2p_port=None, connect_po
         f.write("testnet=0\n")
         f.write("txindex=1\n")
         f.write("daemon=1\n")
+        # To make sure bitcoind gives back p2pkh no matter version
+        f.write("addresstype=legacy\n")
         if connect_port:
             f.write("connect=localhost:"+str(connect_port)+"\n")
             f.write("listen=1\n")
@@ -291,10 +293,11 @@ try:
     print("Test pegout")
     test_pegout(bitcoin.getnewaddress(), sidechain)
 
-    print("Test pegout Segwit (P2SH of witness script)")
+    print("Test P2SH")
     parent_chain_addr = bitcoin.getnewaddress()
-    parent_chain_sw_addr = bitcoin.addwitnessaddress(parent_chain_addr)
-    test_pegout(parent_chain_sw_addr, sidechain)
+    parent_pubkey = bitcoin.validateaddress(parent_chain_addr)["pubkey"]
+    parent_chain_p2sh_addr = bitcoin.createmultisig(1, [parent_pubkey])["address"]
+    test_pegout(parent_chain_p2sh_addr, sidechain)
 
     print("Test pegout Garbage")
     parent_chain_addr = "garbage"

instagibbs:
ACK https://github.com/ElementsProject/elements/pull/318/commits/81a4ed41e7ffd9391b628fbd148be0a4df400cf1

PR #323, opened 2018-03-27
Introduce emergency mode

jtimon:
For things that shouldn't happen like block signers signing invalid blocks or signing diverging histories.
Tested but still not activated for any case.


jtimon:
Needed rebase. 

For testing, ```python3 ./qa/rpc-tests/emergency.py``` works but ```python3 ./qa/pull-tester/rpc-tests.py emergency``` doesn't.

jtimon:
Needed rebase

instagibbs:
Internally I think we've decided that external tools can give us the same assurances, without consensus risk. e.g., call `getchaintips` and make sure there aren't chain splits. If there are, take corrective action.

instagibbs:
Unless there's disagreement on this point, I'll close in a bit.

PR #333, opened 2018-04-09
Fix timing for bitcoin shutdown in pegging.py

jonasnick:
Without this commit pegging.py fails on my machine more than 50% of the cases.


instagibbs:
ACK, seems to be a race where bitcoin2 accepts the new blocks, gets queried by sidechain2, before finishing shutdown.

jtimon:
ACK

Issue #338, opened 2018-04-23
Create compact block RPCs to allow compact passing of block proposals


instagibbs:
A functionary layer should be in charge of passing around proposals and finalized blocks, and an important part is taking advantage of the already-built compact block machinery.

All ignoring "PoW checks":

getcompactsketch -> binary format in hex
consumecompactsketch -> fill in sketch as much as possible, return hex if complete, otherwise return list of missing indices
getgetblocktxn -> binary format in hex
consumegetblocktxn -> fills in sketch, returns complete block if complete, error otherwise

Other improvements to improve sync:
prefill transactions up to limit:
- most recent txns

Don't include transactions in block less than X seconds old (already merged)

fixed in https://github.com/ElementsProject/elements/pull/410

instagibbs:
WIP here: https://github.com/instagibbs/elements/tree/cmpctsketchrpc

Caching isn't so bad, and involves the least amount of maintenance upstream, so I'm going that route.

instagibbs:
merged in elements-0.14+

PR #339, opened 2018-04-25
getnewblockhex: add ability to pass over transactions based on mempoo…

instagibbs:
…l age

Can be used to improve block convergence performance with compact blocks and other mempool sharing based schemes.

related to https://github.com/ElementsProject/elements/issues/338

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -127,7 +127,7 @@ void BlockAssembler::resetBlock()
     blockFinished = false;
 }
 
-std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& scriptPubKeyIn, bool fMineWitnessTx)
+std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& scriptPubKeyIn, bool fMineWitnessTx, int required_age_in_secs)
 {
     int64_t nTimeStart = GetTimeMicros();
 
@@ -172,7 +172,7 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
     //addPriorityTxs(); addPackageTxs will take anything at any rate
     int nPackagesSelected = 0;
     int nDescendantsUpdated = 0;
-    addPackageTxs(nPackagesSelected, nDescendantsUpdated);
+    addPackageTxs(nPackagesSelected, nDescendantsUpdated, required_age_in_secs);
 
     int64_t nTime1 = GetTimeMicros();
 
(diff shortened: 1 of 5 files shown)


src/rpc/mining.cpp
```diff
@@ -162,19 +162,26 @@ UniValue generate(const JSONRPCRequest& request)
 
 UniValue getnewblockhex(const JSONRPCRequest& request)
 {
-    if (request.fHelp || request.params.size() != 0)
+    if (request.fHelp || request.params.size() > 1)
```
instagibbs: need to list optional arg in the line 2 lines down

src/miner.cpp
```diff
@@ -472,16 +473,19 @@ void BlockAssembler::addPackageTxs(int &nPackagesSelected, int &nDescendantsUpda
             }
         }
 
+        // Skip transactions that are under X seconds in mempool
+        if (iter->GetTime() > current_time - required_age_in_secs) {
```
jtimon: Shouldn't this be < ?

src/miner.h
```diff
@@ -180,7 +180,7 @@ class BlockAssembler
     /** Add transactions based on feerate including unconfirmed ancestors
       * Increments nPackagesSelected / nDescendantsUpdated with corresponding
       * statistics from the package selection (for logging statistics). */
-    void addPackageTxs(int &nPackagesSelected, int &nDescendantsUpdated);
+    void addPackageTxs(int &nPackagesSelected, int &nDescendantsUpdated, int required_age_in_secs=0);
```
jtimon: Do we need the default value here too? It defaults to 0 on the rpc already.
Same for CreateNewBlock although I guess there we do need it for generate.

instagibbs: I'm just passing it through, since we only want to use it with "getnewblockhex" type stuff and not generate.

src/miner.cpp
instagibbs: no, I had it backwards first as well, which led to transactions only being accepted within first X seconds :)

src/miner.h
jtimon: But only with the default value in CreateNewBlock should be enough, no?

sipa:
utACK ed121e8df180446775eb729898765da62f7d958f. Is the Travis failure expected?

instagibbs:
@sipa we're not sure... it's 1 of 2 travis runs that is failing the gpg signature checks. We've had this before but can't recall if we just ignored it or fixed it somehow...

instagibbs:
merging, hoping the issue is something that will go away...

instagibbs:
yep tip seems fine

PR #347, opened 2018-05-10
Correct chain selection help

jtimon:
Instead of having petname the name it is directly taken from -chain=<petname>.


instagibbs:
ACK

PR #351, opened 2018-05-11
Default custom chains to no free coins

instagibbs:
In most cases you don't want them.


instagibbs:
pushed fix with added help text to pass documentation checks

instagibbs:
tests are now passing, it's just the `push` verify commits thing that sometimes breaks during PR.

src/init.cpp
```diff
@@ -512,7 +510,10 @@ std::string HelpMessage(HelpMessageMode mode)
     if (showDebug) {
         strUsage += HelpMessageOpt("-fedpegscript=<hex>", _("Change federated peg to use a different script.") +
             " " + _("This creates a new chain with a different genesis block."));
+        strUsage += HelpMessageOpt("-signblockscript=<hex>", _("Change chain to be signed and validated with a different script.") +
```
jtimon: Perhaps -signblockscript and -initialfreecoins could be moved to  https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/chainparamsbase.cpp#L16 ?

instagibbs: My gut feeling is that these are even more advanced than chain-name and should remain hidden.

jtimon: sure, they can remain under showDebug, I'm just proposing to move them from init to chainparamsbase, to the chain selection section.

PR #352, opened 2018-05-11
Make the “21 m to op_true in coinbase” customizable per chain

jtimon:
Default of 0.
Having free coins without a pegin may be useful for regtest-like networks, but for production or testnet-like networks, having 21 M extra allows anyone to empty the sidechain.


instagibbs:
repost of #351

Issue #349, opened 2018-05-10
change initial free coin output amount to 1


instagibbs:
it's only 100 for legacy peg-in reasons which are no longer applicable.

instagibbs:
resolved by #355

PR #355, opened 2018-05-17
Change initial issuance output to 1 

nkostoulas:
#349


nkostoulas:
Can't seem to figure out yet why zapwallettxes.py, preciousblock.py are failing. 

zapwallettxes.py only seems to fail if issuance outputs are less than 3. 

rpc_getblockstats.py also fails, but re generating test data should do it.

instagibbs:
thanks for the submission! Let me know if you need any assistance getting the tests to pass.

jtimon:
Looks good, but it seems some tests are still failing

src/validation.cpp
```diff
@@ -2065,7 +2065,9 @@ bool ApplyTxInUndo(const CTxInUndo& undo, CCoinsViewCache& view, const COutPoint
         CCoinsModifier coins = view.ModifyCoins(out.hash);
         if (undo.nHeight != 0 ||
                 // Special-case genesis since nHeight is always 0, assume DB is clean
-                (Params().GenesisBlock().vtx[0]->GetHash() == out.hash && coins->IsPruned())) {
+                (Params().GenesisBlock().vtx[0]->GetHash() == out.hash && coins->IsPruned()) ||
+                // Special-case for initial issuance since being a single output nHeight is always 0
+                (Params().GenesisBlock().vtx.size() > 1 && Params().GenesisBlock().vtx[1]->GetHash() == out.hash && coins->IsPruned())) {
```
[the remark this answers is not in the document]
instagibbs: I'm actually not totally sure what these lines are doing. Time to dig in.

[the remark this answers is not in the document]
nkostoulas: This didn't seem to be a bug before as the issuance had multiple outputs and the condition `undo.nHeight != 0` was always met. The added clause solves this, but it yea it would be a good idea to check this for any vtx in the genesis.

qa/rpc-tests/wallet.py
```diff
@@ -31,9 +31,9 @@ def setup_network(self, split=False):
     def run_test (self):
 
         # Check that there's 100 UTXOs on each of the nodes
-        assert_equal(len(self.nodes[0].listunspent()), 100)
-        assert_equal(len(self.nodes[1].listunspent()), 100)
-        assert_equal(len(self.nodes[2].listunspent()), 100)
+        assert_equal(len(self.nodes[0].listunspent()), 1)
```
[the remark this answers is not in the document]
nkostoulas: Done

src/validation.cpp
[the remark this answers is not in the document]
instagibbs: are you interested in adding this check?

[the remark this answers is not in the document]
nkostoulas: I could do! Wasn't sure you were expecting it.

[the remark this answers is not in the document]
instagibbs: Would prefer it; I feel if we merge as-is we'll completely forget about it and break something in the future.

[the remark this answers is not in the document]
nkostoulas: Sure I'll add it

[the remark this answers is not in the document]
nkostoulas: I've added the check! Had to rebase because same tests had changed, but it's fine now.

instagibbs:
ping @sipa

instagibbs:
Actually this PR may fix https://github.com/ElementsProject/elements/issues/378.

instagibbs:
light tACK, fixes #378

src/validation.cpp
```diff
@@ -2067,7 +2067,16 @@ bool ApplyTxInUndo(const CTxInUndo& undo, CCoinsViewCache& view, const COutPoint
         CCoinsModifier coins = view.ModifyCoins(out.hash);
         if (undo.nHeight != 0 ||
                 // Special-case genesis since nHeight is always 0, assume DB is clean
-                (Params().GenesisBlock().vtx[0]->GetHash() == out.hash && coins->IsPruned())) {
+                (Params().GenesisBlock().vtx[0]->GetHash() == out.hash && coins->IsPruned()) ||
```
instagibbs: Actually this is redundant now. It can just be the one for loop checking all of them.

instagibbs: We can probably get rid of this lambda too :)

src/validation.cpp
```diff
@@ -2067,7 +2067,16 @@ bool ApplyTxInUndo(const CTxInUndo& undo, CCoinsViewCache& view, const COutPoint
...
+                // Special-case for initial issuance since being a single output nHeight is always 0
+                (Params().GenesisBlock().vtx.size() > 1 && [&]() {
```
jtimon: shouldn't this be ```Params().GenesisBlock().vtx.size() >= 1``` ?

src/chainparams.cpp
```diff
@@ -200,7 +200,7 @@ class CCustomParams : public CChainParams {
 
         genesis = CreateGenesisBlock(consensus, strNetworkID, 1296688602, genesisChallengeScript, 1);
         if (initialFreeCoins != 0) {
-            AppendInitialIssuance(genesis, COutPoint(uint256(commit), 0), parentGenesisBlockHash, 100, initialFreeCoins/100, 0, 0, CScript() << OP_TRUE);
+            AppendInitialIssuance(genesis, COutPoint(uint256(commit), 0), parentGenesisBlockHash, 1, initialFreeCoins, 0, 0, CScript() << OP_TRUE);
```
jtimon: Perhaps we can just remove the parameter and simplify the function. That can be done on a different PR though.

src/validation.cpp
instagibbs: it does an identical check for the first transaction case already. It's redundant however.

src/chainparams.cpp
nkostoulas: Yep thought of that as well, could change this if you want

src/validation.cpp
nkostoulas: Thought it’s covered by the first clause. Should I change something?

src/chainparams.cpp
instagibbs: I don't know, might be useful in the future. Meh :)

src/validation.cpp
nkostoulas: Couldn’t this still happen if we added more than one issuance in the genesis?

instagibbs: I'm trying to say the code block below already covers this case, no?

nkostoulas: Ah yeah sure, i just thought the first time you mentioned it was in case we decided to change this code below

nkostoulas: Just in case :)

instagibbs: I suppose it'll make rebasing easier.

instagibbs:
@nkostoulas thanks! taking as-is.

PR #359, opened 2018-05-25
Rework control flow in VerifyAmount()

real-or-random:
This mainly improves and simplifies the logic control flow VerifyAmount(). It fixes an issue that appeared when 0 units of an asset (or a re-issuance token) are issued using an explicit commitment. In this case, we skipped over some checks that should be present for correct verification.

Since the issue is present both when issuing assets and when issuing re-issuance tokens, and the code is similar, this PR extracts the relevant code into a (non-exported) helper function.

On the way, it fixes a potential access to a non-existent input witness.

### DIFF
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -43,8 +43,6 @@ public:
                     return;
                 /* Explicit value */
                 case 1:
-                /* Trust-me! asset generation */
-                case 0xff:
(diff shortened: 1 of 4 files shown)


instagibbs:
ACK https://github.com/ElementsProject/elements/pull/359/commits/05602edf6fa981bb8f36cb7363b1a750d32322d7

PR #361, opened 2018-05-25
Remove overflow checks for explicit input amounts

real-or-random:
This removes checks from the pre-asset era that reject amounts on the
input side of a transaction if these amounts overflow. The checks are
superficial because VerifyAmounts() already rejects input amounts that
could potentially overflow.

One of the removed checks is wrong because it rejects legitimate
transactions that spend more than MAX_MONEY asset units altogether,
even if those units belong to different assets. For example, a
transaction spending MAX_MONEY units of an asset "apple" and MAX_MONEY
units of an asset "orange" was previously rejected because the code
literally added apples and oranges in this case.


instagibbs:
utACK b2a729c4a594cb9902ac5b6b6123b48223fac761

All amount checking is already done inside `VerifyAmounts`, which already does these individual checks.

Issue #360, opened 2018-05-25
Support other elements chains as parent chains


jtimon:
Currently the parent chain must be always bitcoin-like (ie with proof of work).
Allowing the parent chain to be a chain with signed blocks like elements would be helpful for testing.
For example, we could have a version of pegging.py running on travis, which uses another elements chain as parent chain instead of bitcoin's regtest (requiring an external binary).

PR #362, opened 2018-05-25
2WP: Allow other elements chains as parent chains for pegs

jtimon:
Fix https://github.com/ElementsProject/elements/issues/360 (doesn't really fix it as noted in https://github.com/ElementsProject/elements/pull/362#issuecomment-406085649 , but it is still useful for testing)

Dependencies:

- [x] Refactor: Validation: Make parts of IsValidPeginWitness templated #370

### DIFF
--- a/qa/rpc-tests/feature_fedpeg.py
+++ b/qa/rpc-tests/feature_fedpeg.py
@@ -1,0 +1,340 @@
+#!/usr/bin/env python3
+
+from decimal import Decimal
+import json
+import time
+
+from test_framework.authproxy import JSONRPCException
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    connect_nodes_bi,
+    rpc_auth_pair,
+    rpc_port,
+    start_node,
+    start_nodes,
+    stop_node,
+)
+
+# Sync mempool, make a block, sync blocks
+def sync_all(sidechain, sidechain2, makeblock=True):
+    block = ""
+    timeout = 20
+    while len(sidechain.getrawmempool()) != len(sidechain2.getrawmempool()):
+        time.sleep(1)
+        timeout -= 1
+        if timeout == 0:
+            raise Exception("Peg-in has failed to propagate.")
+    if makeblock:
+        block = sidechain2.generate(1)
+    while sidechain.getblockcount() != sidechain2.getblockcount():
+        time.sleep(1)
+        timeout -= 1
+        if timeout == 0:
+            raise Exception("Blocks are not propagating.")
+    return block
+
+def get_new_unconfidential_address(node):
+    addr = node.getnewaddress()
+    val_addr = node.validateaddress(addr)
+    if 'unconfidential' in val_addr:
+        return val_addr['unconfidential']
+    return val_addr['address']
+
+class FedPegTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 4
+
+    def setup_network(self, split=False):
+
+        # Parent chain args
+        self.extra_args = [[
+            # '-printtoconsole',
+            '-validatepegin=0',
+            '-anyonecanspendaremine',
+            '-initialfreecoins=2100000000000000',
+        ]] * 2
+
+        self.nodes = start_nodes(2, self.options.tmpdir, self.extra_args[:2], chain='parent')
+        connect_nodes_bi(self.nodes, 0, 1)
+        self.parentgenesisblockhash = self.nodes[0].getblockhash(0)
+        print('parentgenesisblockhash', self.parentgenesisblockhash)
+        parent_pegged_asset = self.nodes[0].getsidechaininfo()['pegged_asset']
+
+        # Sidechain args
+        parent_chain_signblockscript = '51'
+        for n in range(2):
+            rpc_u, rpc_p = rpc_auth_pair(n)
+            self.extra_args.append([
+                # '-printtoconsole',
+                '-parentgenesisblockhash=%s' % self.parentgenesisblockhash,
+                '-validatepegin=1',
+                '-anyonecanspendaremine=0',
+                '-initialfreecoins=0',
+                '-peginconfirmationdepth=10',
+                '-mainchainrpchost=127.0.0.1',
+                '-mainchainrpcport=%s' % rpc_port(n),
+                '-mainchainrpcuser=%s' % rpc_u,
+                '-mainchainrpcpassword=%s' % rpc_p,
+                '-parentpubkeyprefix=235',
+                '-parentscriptprefix=75',
+                '-con_parent_chain_signblockscript=%s' % parent_chain_signblockscript,
+                '-con_parent_pegged_asset=%s' % parent_pegged_asset,
+            ])
+            self.nodes.append(start_node(n + 2, self.options.tmpdir, self.extra_args[n + 2], chain='sidechain'))
+
+        connect_nodes_bi(self.nodes, 2, 3)
+        self.is_network_split = True
+        self.sync_all()
+
+    def test_pegout(self, parent_chain_addr, sidechain):
+        pegout_txid = sidechain.sendtomainchain(parent_chain_addr, 1)
+        raw_pegout = sidechain.getrawtransaction(pegout_txid, True)
+        assert 'vout' in raw_pegout and len(raw_pegout['vout']) > 0
+        pegout_tested = False
+        for output in raw_pegout['vout']:
+            scriptPubKey = output['scriptPubKey']
+            if 'type' in scriptPubKey and scriptPubKey['type'] == 'nulldata':
+                assert ('pegout_hex' in scriptPubKey and 'pegout_asm' in scriptPubKey and 'pegout_type' in scriptPubKey and
+                        'pegout_chain' in scriptPubKey and 'pegout_reqSigs' in scriptPubKey and 'pegout_addresses' in scriptPubKey)
+                assert scriptPubKey['pegout_chain'] == self.parentgenesisblockhash
+                assert scriptPubKey['pegout_reqSigs'] == 1
+                assert parent_chain_addr in scriptPubKey['pegout_addresses']
+                pegout_tested = True
+                break
+        assert pegout_tested
+
+    def run_test(self):
+        parent = self.nodes[0]
+        parent2 = self.nodes[1]
+        sidechain = self.nodes[2]
+        sidechain2 = self.nodes[3]
+
+        parent.generate(101)
+        sidechain.generate(101)
+
+        addrs = sidechain.getpeginaddress()
+        addr = parent.validateaddress(addrs["mainchain_address"])
+        print('addrs', addrs)
+        print('addr', addr)
+        txid1 = parent.sendtoaddress(addrs["mainchain_address"], 24)
+        # 10+2 confirms required to get into mempool and confirm
+        parent.generate(1)
+        time.sleep(2)
+        proof = parent.gettxoutproof([txid1])
+        
+        raw = parent.getrawtransaction(txid1)
+        print('raw', parent.getrawtransaction(txid1, True))
+
+        print("Attempting peg-in")
+        # First attempt fails the consensus check but gives useful result
+        try:
+            pegtxid = sidechain.claimpegin(raw, proof)
+            raise Exception("Peg-in should not be mature enough yet, need another block.")
+        except JSONRPCException as e:
+            print('ERROR:', e.error)
+            assert("Peg-in Bitcoin transaction needs more confirmations to be sent." in e.error["message"])
+
+        # Second attempt simply doesn't hit mempool bar
+        parent.generate(10)
+        try:
+            pegtxid = sidechain.claimpegin(raw, proof)
+            raise Exception("Peg-in should not be mature enough yet, need another block.")
+        except JSONRPCException as e:
+            assert("Peg-in Bitcoin transaction needs more confirmations to be sent." in e.error["message"])
+
+        # Should fail due to non-witness
+        try:
+            pegtxid = sidechain.claimpegin(raw, proof, get_new_unconfidential_address(parent))
+            raise Exception("Peg-in with non-matching claim_script should fail.")
+        except JSONRPCException as e:
+            print(e.error["message"])
+            assert("Given or recovered script is not a witness program." in e.error["message"])
+
+        # # Should fail due to non-matching wallet address
+        # try:
+        #     pegtxid = sidechain.claimpegin(raw, proof, get_new_unconfidential_address(sidechain))
+        #     raise Exception("Peg-in with non-matching claim_script should fail.")
+        # except JSONRPCException as e:
+        #     print(e.error["message"])
+        #     assert("Given claim_script does not match the given Bitcoin transaction." in e.error["message"])
+
+        # 12 confirms allows in mempool
+        parent.generate(1)
+        # Should succeed via wallet lookup for address match, and when given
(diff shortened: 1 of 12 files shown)


jtimon:
Rebased after https://github.com/ElementsProject/elements/pull/363 was merged, which was a requirement for this (or some other way to do the same).
Also fixed a couple of bugs and greatly improved the testing. The new test is mostly a copy of pegging.py but adapted for the test framework and with some parts commented that are not properly adapted yet.
Ideally in the future we can get rid of pegging.py by offering an option to force a pow chain as parent chain by providing a path to a binary like we do in pegging.py.

jtimon:
Removed the 'WIP' because now it seems to pass most of the tests.
This doesn't need to replace pegging.py, review and testing should focus on making sure the pervious default behavior is not broken in any way.

instagibbs:
Will review

src/init.cpp
```diff
@@ -516,7 +516,7 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-initialfreecoins", strprintf(_("The amount of OP_TRUE coins created in the genesis block. Primarily for testing. (default: %d)"), 0));
         strUsage += HelpMessageOpt("-parentpubkeyprefix", strprintf(_("The byte prefix, in decimal, of the parent chain's base58 pubkey address. (default: %d)"), 111));
         strUsage += HelpMessageOpt("-parentscriptprefix", strprintf(_("The byte prefix, in decimal, of the parent chain's base58 script address. (default: %d)"), 196));
-
+        strUsage += HelpMessageOpt("-con_parent_chain_has_pow", "Whether parent chain uses pow or signed blocks default: 1");
```
instagibbs: put default in parenthesis

src/validation.cpp
```diff
@@ -2356,6 +2357,87 @@ CScript calculate_contract(const CScript& federationRedeemScript, const CScript&
...
+{
+    if (!txBTC.vout[nOut].nValue.IsExplicit()) {
+        return false;
+    }
+    amount = txBTC.vout[nOut].nValue.GetAmount();
+    return true;
+}
+
+template<typename T>
+static bool CheckPegoutTx(const std::vector<unsigned char>& tx_data, T& pegtx, const COutPoint& prevout, const CAmount claim_amount, const CScript& claim_script)
```
instagibbs: CheckPeginTx?

src/validation.cpp
```diff
@@ -2406,45 +2488,41 @@ bool IsValidPeginWitness(const CScriptWitness& pegin_witness, const COutPoint& p
...
-        CDataStream merkleBlockStream(stack[5], SER_NETWORK, PROTOCOL_VERSION);
-        merkleBlockStream >> merkle_block;
-        if (!merkleBlockStream.empty() || !CheckBitcoinProof(merkle_block.header.GetHash(), merkle_block.header.nBits)) {
-           return false;
+        Sidechain::Bitcoin::CMerkleBlock merkle_block_pow;
+        if (!GetBlockAndTxFromMerkleBlock(block_hash, tx_hash, merkle_block_pow, stack[5])) {
+            return false;
         }
-        if (merkle_block.txn.ExtractMatches(txHashes, txIndices) != merkle_block.header.hashMerkleRoot || txHashes.size() != 1) {
+        // TODO do this inside to prever DoS ?
```
instagibbs: prevent?

instagibbs:
concept ACK. Going to have to carefully comb over these changes.

instagibbs:
Could we split up these changes into basic refactors, then consensus changes?

jtimon:
Uncommented more tests.

src/validation.cpp
```diff
@@ -2359,6 +2360,87 @@ CScript calculate_contract(const CScript& federationRedeemScript, const CScript&
...
+    } catch (std::exception& e) {
+        // Invalid encoding of merkle block
+        return false;
+    }
+    return true;
+}
+
+template<typename T>
+static bool CheckPeginTx(const std::vector<unsigned char>& tx_data, T& pegtx, const COutPoint& prevout, const CAmount claim_amount, const CScript& claim_script)
+{
```
mword: - The `T& pegtx` parameter is set and used internally to this function but is ignore by callers, is it needed in the parameter list to *CheckPeginTx()*? It looks like it's just there for type selection for the plumbing underneath? Alternatively you could add a local instantiation before the `try`:

    `T pegtx; // or shared_ptr<const T> pegtx;`

 - Also, because the function can only take *T*'s of type *CTransactionRef*, maybe a more helpful name could be used for the template param (e.g. *"TransactionType"*)?

src/validation.cpp
```diff
@@ -2359,6 +2360,87 @@ CScript calculate_contract(const CScript& federationRedeemScript, const CScript&
...
+    try {
+        CDataStream pegtx_stream(tx_data, SER_NETWORK, PROTOCOL_VERSION);
+        pegtx_stream >> pegtx;
```
mword: The community seems to be okay with this (paraphrased) construct but it just seems wrong to me:

```
    shared_ptr<const CTransaction> pegtx;  //Uninitialized shared_ptr!!
    pegtx_stream >> pegtx;  //Looks like a bug - how are we to know pegtx is initialized before written to?
```

What be the downside of just using *Unserialize* instead?
```
    shared_ptr<const CTransaction> pegtx;
    Unserialize(pegtx_stream, pegtx);
```

instagibbs: It's doing that under the hood, so I don't see how it could be an uninitialized shared pointer.

mword: Yes, I get that the initialization is happening under the hood in Unserialize.  It's not really a bug but it looked like it could be one to me (as a noob in this code base anyway) while doing a casual reading of the code. It's NBD, it just made the code a little harder to grok for me.

jtimon: I just wanted to leave the code as similar as it was before as possible, but I can use 
```
shared_ptr<const CTransaction> pegtx;
    Unserialize(pegtx_stream, pegtx);
```
if that's preferred.

instagibbs: as-is is fine imo.

src/validation.cpp
jtimon: Yes, if the parameters are only used internally they can just be turned into local variables, good point.
CTransactionRef (of both types) uses shared_ptr instead of unique pointer, I assume because other parts need that. I guess I could use a unique_ptr instead of CTransactionRef, but this is what was being done and it seems simpler.

jtimon: oh, no, sorry, I cannot do that, the paremeter is needed to select which one to use within the if/else clause. I know this is hacky and probably can be done in a more elegant way, but I don't see it at this point.

jtimon:
Rebased after #370 was merged

src/validation.cpp
```diff
@@ -2492,20 +2502,36 @@ bool IsValidPeginWitness(const CScriptWitness& pegin_witness, const COutPoint& p
...
+            return false;
+        }
+    } else {
+
+        CMerkleBlock merkle_block;
+        if (!GetBlockAndTxFromMerkleBlock(block_hash, tx_hash, merkle_block, stack[5])) {
+            return false;
+        }
+
+        if (!CheckProof(merkle_block.header, Params().GetConsensus())) {
```
instagibbs: Doesn't this just check parent chain's signatures against their own?

In other words, it will fail if the signblockscript isn't the same between chains?

jtimon: Right, good catch. I need a parameter for the parent chain's scriptPubKey/challenge, perhaps replacing the has_pow parameter.

jtimon:
Changed from parent_chain_has_pow to parent_chain_signblockscript as discussed in  https://github.com/ElementsProject/elements/pull/362#discussion_r198874921

src/init.cpp
```diff
@@ -516,7 +516,7 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-initialfreecoins", strprintf(_("The amount of OP_TRUE coins created in the genesis block. Primarily for testing. (default: %d)"), 0));
         strUsage += HelpMessageOpt("-parentpubkeyprefix", strprintf(_("The byte prefix, in decimal, of the parent chain's base58 pubkey address. (default: %d)"), 111));
         strUsage += HelpMessageOpt("-parentscriptprefix", strprintf(_("The byte prefix, in decimal, of the parent chain's base58 script address. (default: %d)"), 196));
-
+        strUsage += HelpMessageOpt("-con_parent_chain_signblockscript", _("Whether parent chain uses pow or signed blocks. If the parent chain uses signed blocks, the challenge (scriptPubKey) script. If not, an empty string. (default: empty script [ie parent uses pow])"));
```
instagibbs: we should expose this via `getsidechaininfo`, including a binary saying if it's PoW for ease of reading

src/validation.cpp
```diff
@@ -2492,20 +2502,36 @@ bool IsValidPeginWitness(const CScriptWitness& pegin_witness, const COutPoint& p
...
-
     // Get txout proof
-    Sidechain::Bitcoin::CMerkleBlock merkle_block;
-    if (!GetBlockAndTxFromMerkleBlock(block_hash, tx_hash, merkle_block, stack[5])) {
-        return false;
-    }
-    if (!CheckBitcoinProof(block_hash, merkle_block.header.nBits)) {
-        return false;
-    }
+    if (Params().GetConsensus().parent_chain_signblockscript == CScript()) {
```
instagibbs: Let's make a wrapper function to make this self-documenting.

src/validation.cpp
```diff
@@ -2492,20 +2502,36 @@ bool IsValidPeginWitness(const CScriptWitness& pegin_witness, const COutPoint& p
...
 
-    // Get serialized transaction
-    Sidechain::Bitcoin::CTransactionRef pegtx;
-    if (!CheckPeginTx(stack[4], pegtx, prevout, value, claim_script)) {
-        return false;
+        Sidechain::Bitcoin::CMerkleBlock merkle_block_pow;
+        if (!GetBlockAndTxFromMerkleBlock(block_hash, tx_hash, merkle_block_pow, stack[5])) {
+            return false;
+        }
+        if (!CheckBitcoinProof(block_hash, merkle_block_pow.header.nBits)) {
```
instagibbs: Now we have two very similar things with different naming conventions, `CheckBitcoinProof` and `CheckProofParent`.

Can we just abstract this out to `CheckProofParent` with a boolean arg switch?

jtimon: Yes, I just wanted to avoid changing CheckProof since it is called a few times. I could use a boolean that defaults to false, but I tend to dislike default arguments in C++. I find ```CheckProofParent(merkle_block.header, Params().GetConsensus())``` more clear than ```CheckProof(merkle_block.header, Params().GetConsensus(), true)```, which forces me to go to the function and see what the boolean is about. If it was python, ```CheckProof(merkle_block.header, Params().GetConsensus(), use_parent_chain_script=True)``` is fine. Perhaps I can just add a comment when using the boolean.

instagibbs: fair enough, perhaps a minor renaming then to make them obviously related

instagibbs: or heck just comment what it's doing

jtimon:
Rebased

jtimon:
Note that this is fine for testing, but the feature is not secure.
Right now it is accepting pegins from the parent elements chain without checking the asset being sent in the parent chain. so basically anyone can issue an asset on the parent chain and send that new asset for pegin and it will accepted as if it was the parentchain “genesis asset”.
A separate PR can be created to make the feature useful beyond testing. I guess that could be documented more clearly.

src/validation.cpp
jtimon: Renamed to CheckProofSignedParent as discussed on chat.

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -0,0 +1,338 @@
...
+        print('addrs', addrs)
+        print('addr', addr)
+        txid1 = parent.sendtoaddress(addrs["mainchain_address"], 24)
+        # 10+2 confirms required to get into mempool and confirm
+        parent.generate(1)
+        time.sleep(2)
+        proof = parent.gettxoutproof([txid1])
+        
+        raw = parent.getrawtransaction(txid1)
+        import json
```
stevenroose: That's a strange place to import json..

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -0,0 +1,338 @@
...
+
+        # Should fail due to non-witness
+        try:
+            pegtxid = sidechain.claimpegin(raw, proof, get_new_unconfidential_address(parent))
+            raise Exception("Peg-in with non-matching claim_script should fail.")
+        except JSONRPCException as e:
+            print(e.error["message"])
+            assert("Given or recovered script is not a witness program." in e.error["message"])
+
+        # # Should fail due to non-matching wallet address
```
stevenroose: Is this here because it should fail due to this if it would not fail due to non-witness (clause above)?

This is very good for clarity, but it doesn't really test if it would fail in that case..

instagibbs:
How much extra work is it to do this properly and check for the correct asset id? Shouldn't it just be a single check?

qa/rpc-tests/feature_fedpeg.py
jtimon: That's simply the same test as in https://github.com/ElementsProject/elements/blob/elements-0.14.1/qa/rpc-tests/pegging.py#L233 but commented because I didn't manage to make it work, note here there's a similar one above "Should fail due to non-witness", perhaps that's enough.
Perhaps I should just remove that commented code, but then I will forget about it.
@instagibbs thoughts?

instagibbs:
The output shouldn't be confidential though? The mainchain_address is unconfidential, yes?

jtimon:
Right, sorry, it was just a bug that confused me. It seems to work now. Let me know if you prefer me to squash, rename the commit or something else. I would squash, I think.

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -61,6 +61,8 @@ def setup_network(self, split=False):
         connect_nodes_bi(self.nodes, 0, 1)
         self.parentgenesisblockhash = self.nodes[0].getblockhash(0)
         print('parentgenesisblockhash', self.parentgenesisblockhash)
+        parent_pegged_asset = self.nodes[0].getsidechaininfo()['pegged_asset']
+        print('parent_pegged_asset', parent_pegged_asset)
```
instagibbs: errant print?

src/wallet/rpcwallet.cpp
```diff
@@ -3552,27 +3554,26 @@ UniValue createrawpegin(const JSONRPCRequest& request)
...
     CDataStream ssTxOutProof(txOutProofData, SER_NETWORK, PROTOCOL_VERSION);
-    Sidechain::Bitcoin::CMerkleBlock merkleBlock;
     try {
         ssTxOutProof >> merkleBlock;
     }
     catch (...) {
         throw JSONRPCError(RPC_TYPE_ERROR, "The included txoutproof is malformed. Are you sure that is the whole string?");
     }
 
-    if (!ssTxOutProof.empty() || !CheckBitcoinProof(merkleBlock.header.GetHash(), merkleBlock.header.nBits))
```
instagibbs: why are we dropping this check?

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/362/commits/3f62b3249bae52ffb99c16ad557b5d0372b8a432 other than single question

qa/rpc-tests/feature_fedpeg.py
instagibbs: was this functionality supposed to be working?

PR #344, opened 2018-05-04
RPC: Signblock: Support signblock info on rpc/blockchain

jtimon:
Bigger refactor than I initially thought.
But basically still cleaning up pow stuff we don't need, and including a few new blocksign-related fields that replace those and document them. 

Doesn't come with tests, but if it's not only rpc, then nack, please.

Dependencies:

- [x] QA: Reactivate and improve rpc_blockchain tests https://github.com/ElementsProject/elements/pull/365
~~- [ ] RPC: Stop providing useless fake pow data #366~~

PR #365, opened 2018-06-04
QA: Reactivate and improve rpc_blockchain tests

jtimon:
Since https://github.com/ElementsProject/elements/pull/344 works locally but fails in weird ways in travis, let's just reactivate and improve the test first for now until we understand what's happening on #344 .
More tests shouldn't hurt and if merged should also reduce #344's diff.


instagibbs:
utACK

PR #366, opened 2018-06-05
RPC: Stop providing useless fake pow data

jtimon:
Since https://github.com/ElementsProject/elements/pull/362 is failing, let's start by removing the things we don't need in a way we can test incrementally.
Let's restore later the functionality in header/block.bits (ie the blocksign challenge/scriptsig, via GetChallengeStr) that nobody seems to have used, noticed, tested or documented for a long time.

Depndencies:

- [X] QA: Reactivate and improve rpc_blockchain tests #365


jtimon:
Closing since I found out the problem with https://github.com/ElementsProject/elements/pull/344
see https://github.com/ElementsProject/elements/pull/344#issuecomment-396690206

PR #370, opened 2018-06-12
Refactor: Validation: Make parts of IsValidPeginWitness templated

jtimon:
This should not change behavior in any way.

Separated out of https://github.com/ElementsProject/elements/pull/362

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2359,6 +2359,12 @@ CScript calculate_contract(const CScript& federationRedeemScript, const CScript&
     return scriptDestination;
 }
 
+bool GetAmountFromParentChainPegin(CAmount& amount, const Sidechain::Bitcoin::CTransaction& txBTC, unsigned int nOut)
+{
+    amount = txBTC.vout[nOut].nValue;
+    return true;
+}
+
 template<typename T>
 static bool GetBlockAndTxFromMerkleBlock(uint256& block_hash, uint256& tx_hash, T& merkle_block, const std::vector<unsigned char>& merkle_block_raw)
 {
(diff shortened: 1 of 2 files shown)


src/validation.cpp
```diff
@@ -2359,6 +2359,12 @@ CScript calculate_contract(const CScript& federationRedeemScript, const CScript&
     return scriptDestination;
 }
 
+bool GetAmountFromParentChainPegin(CAmount& amount, const Sidechain::Bitcoin::CTransaction& txBTC, unsigned int nOut)
+{
+    amount = txBTC.vout[nOut].nValue;
+    return true;
```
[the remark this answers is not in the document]
jtimon: The point is to make the interface like https://github.com/ElementsProject/elements/pull/362/commits/23af08444f59fa1ce6a4de0c1b16e43a74ebf98d#diff-24efdb00bfbe56b140fb006b562cc70bR2369 which can fail.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/370/commits/20f9fb3e5254ffcaa8ddf84dba5e1a58529f9cda

I'd like another set of eyes on this as well.

jtimon:
Fixed nits

src/validation.cpp
```diff
@@ -2359,6 +2359,78 @@ CScript calculate_contract(const CScript& federationRedeemScript, const CScript&
...
+}
+
+template<typename T>
```
jhfrontz: I'm curious about the use of a template here (vs., say, an inline static function).  This isn't a header file and there only appears to be one use of this template/function -- is this a project coding convention or is there some performance/maintenance gain?

instagibbs:
re-utACK https://github.com/ElementsProject/elements/pull/370/commits/a20986b40dd8330510fc8970f48d8ad620141adb

src/validation.cpp
mword: I'm not the biggest fan of adding code for future use into a pull request, it could sit there making no sense for a long time.

mword:
I built and ran all tests (including qa tests and the peggin.py test) with no errors. I made some comments in #362 that are relevant to #370 I think (362 seemed the better place to make my comments). I got to the point where I understood the code if not the higher level logic that makes these changes necessary (I'm still coming to terms with that).

All in all: tACK

mword:
ACK

src/validation.cpp
jtimon: Yeah, I did it only because @instagibbs asked. Perhaps I should have done 2 commits but not 2 PRs

PR #373, opened 2018-06-18
QA: Introduce feature_blocksign.py

jtimon:
Let's add this test that @stevenroose wrote, I only slightly modified it.

Later I want to extend it to test things other than 3 of 3, including a single signer (without using multisig if it's 1 of 1).

### DIFF
--- a/qa/rpc-tests/feature_blocksign.py
+++ b/qa/rpc-tests/feature_blocksign.py
@@ -1,0 +1,124 @@
+#!/usr/bin/env python3
+
+import codecs
+import hashlib
+import os
+import random
+
+from test_framework import (
+    address,
+    key,
+    test_framework,
+    util,
+)
+
+# Generate wallet import format from private key.
+def wif(pk):
+    # Base58Check version for regtest WIF keys is 0xef = 239
+    return address.byte_to_base58(pk, 239)
+
+# The signblockscript is a Bitcoin Script k-of-n multisig script.
+def make_signblockscript(num_nodes, required_signers, keys):
+    assert(num_nodes >= required_signers)
+    script = "{}".format(50 + required_signers)
+    for i in range(num_nodes):
+        k = keys[i]
+        script += "41"
+        script += codecs.encode(k.get_pubkey(), 'hex_codec').decode("utf-8")
+    script += "{}".format(50 + num_nodes) # num keys
+    script += "ae" # OP_CHECKMULTISIG
+    print('signblockscript', script)
+    return script
+
+class BlockSignTest(test_framework.BitcoinTestFramework):
+
+    # Dynamically generate N keys to be used for block signing.
+    def init_keys(self, num_keys):
+        self.keys = []
+        self.wifs = []
+        for i in range(num_keys):
+            k = key.CECKey()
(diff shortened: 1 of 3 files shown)


qa/rpc-tests/feature_blocksign.py
```diff
@@ -0,0 +1,114 @@
...
+        signedblock = result["hex"]
+        miner.submitblock(signedblock)
+
+    def mine_blocks(self, num_blocks):
+        for i in range(num_blocks):
+            self.mine_block()
+            self.sync_all()
+
+    def run_test(self):
+        # Check that there's 100 UTXOs on each of the nodes
```
instagibbs: I don't think these checks are useful in this test. Check the block height instead?

qa/rpc-tests/feature_blocksign.py
```diff
@@ -0,0 +1,114 @@
...
+            script += "41"
+            script += k.get_pubkey().hex()
+        script += "{}".format(50+n) # num keys
+        script += "ae" # OP_CHECKMULTISIG
+        return script
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 3
```
instagibbs: can we factor out "3" for all these places in case we want to expand it?

jtimon:
Fixed nits, moved from 3-of-3 to 7-of-9, that makes it slower (from 11s to 34s in my hardware), but I think that's ok. We can expand further afterwards, but this works now.

qa/rpc-tests/feature_blocksign.py
```diff
@@ -0,0 +1,117 @@
...
+        # broadcasts it
+        mineridx = self.nodes[0].getblockcount() % self.num_nodes # assuming in sync
+        miner = self.nodes[mineridx]
+
+        # miner makes a block
+        block = miner.getnewblockhex()
+
+        # collect required_signers signatures
+        sigs = []
+        for i in range(self.required_signers):
```
instagibbs: We should loop and make sure that combining below-threshhold doesn't give you a block considered valid.

At each inner-section we should combine and check that it's not complete, and cannot move the blockchain forward.

stevenroose:
Hehe, nice to see that thing being even the slightest bit inspiring :) It was mostly me getting familiar with the Python framework :)

Can't wait to join you guys!

qa/rpc-tests/feature_blocksign.py
```diff
@@ -0,0 +1,124 @@
...
+            result = miner.combineblocksigs(block, sigs)
+            util.assert_equal(result["complete"], False)
```
instagibbs: Please also do a submitblock, then check that it didn't move the chain forward. Belts and suspenders!

qa/rpc-tests/feature_blocksign.py
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env python3
+# requires Python 3.5
```
instagibbs: We shouldn't require a specific version of python3.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/373/commits/f66e94a7120caa4a9a53395853bc837d75aaed6b

@stevenroose See you soon!

PR #374, opened 2018-06-18
Port tweakfedpegscript from liquid-deamon

mword:
This is a port of the changes made in liquid-deamon to add tweakfedpegscript in rpc/misc.cpp.

### DIFF
--- a/qa/rpc-tests/pegging.py
+++ b/qa/rpc-tests/pegging.py
@@ -201,9 +201,12 @@ try:
 
     sidechain.generate(101)
 
-    addrs = sidechain.getpeginaddress()
-    txid1 = bitcoin.sendtoaddress(addrs["mainchain_address"], 24)
-    # 10+2 confirms required to get into mempool and confirm
+    getpeginaddr_res = sidechain.getpeginaddress()
+    addr = getpeginaddr_res["mainchain_address"]
+    claim_script = getpeginaddr_res["claim_script"]
+    assert(addr == sidechain.tweakfedpegscript(claim_script)["address"])
+    txid1 = bitcoin.sendtoaddress(addr, 24)
+
     bitcoin.generate(1)
     time.sleep(2)
     proof = bitcoin.gettxoutproof([txid1])
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -583,6 +583,38 @@ UniValue echo(const JSONRPCRequest& request)
     return request.params;
 }
 
+UniValue tweakfedpegscript(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 1)
+        throw runtime_error(
+            "tweakfedpegscript \"claim_script\"\n"
+            "\nReturns a tweaked fedpegscript.\n"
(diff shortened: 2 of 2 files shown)


qa/rpc-tests/pegging.py
```diff
@@ -168,6 +168,7 @@ def test_pegout(parent_chain_addr, sidechain):
     bitcoindstart = bitcoin_bin_path+"/bitcoind -datadir="+bitcoin_datadir
     subprocess.Popen(bitcoindstart.split(), stdout=subprocess.PIPE)
 
+
```
instagibbs: unrelated line

src/rpc/misc.cpp
```diff
@@ -593,11 +625,11 @@ static const CRPCCommand commands[] =
     { "util",               "createblindedaddress",   &createblindedaddress,   true,  {} },
     { "util",               "verifymessage",          &verifymessage,          true,  {"address","signature","message"} },
     { "util",               "signmessagewithprivkey", &signmessagewithprivkey, true,  {"privkey","message"} },
-
+    { "util",               "tweakfedpegscript",      &tweakfedpegscript,      true,  {"claim_script"} },
     /* Not shown in help */
     { "hidden",             "setmocktime",            &setmocktime,            true,  {"timestamp"}},
     { "hidden",             "echo",                   &echo,                   true,  {"arg0","arg1","arg2","arg3","arg4","arg5","arg6","arg7","arg8","arg9"}},
-    { "hidden",             "echojson",               &echo,                  true,  {"arg0","arg1","arg2","arg3","arg4","arg5","arg6","arg7","arg8","arg9"}},
+    { "hidden",             "echojson",               &echo,                   true,  {"arg0","arg1","arg2","arg3","arg4","arg5","arg6","arg7","arg8","arg9"}},
```
instagibbs: unrelated space

instagibbs:
ACK otherwise

jtimon:
Perhaps it would be nice to wait for https://github.com/ElementsProject/elements/pull/362 ?
Otherwise #362 could be rebased on top of this.
My only point is that pegging.py and feature_2wp.py should be mirrored in tests until the latter fully replaces the former after adding a --custom-binary option or similar, to support testing with any bitcoin core version as parent chain (like pegging currently does).

qa/rpc-tests/pegging.py
mword: Cleaned this up

src/rpc/misc.cpp
mword: Re-added the line that was removed: The space change was just to fix the column alignment in this line.

mword:
@jtimon this is already implemented in liquid so when #362 is ported to liquid it will be rebased on top of this change there. Which is less confusing?

instagibbs:
If I had to guess I'd say it would be simpler to merge this first and rebase the other PR, since it's still dependent on yet another PR.

instagibbs:
Needs squash

mword:
@instagibbs Squashed (I hope)

instagibbs:
ACK

PR #382, opened 2018-06-28
WIP: RPC: Make generate work with a single signer

jtimon:
It would simplify and optimize things for a single signer chain by using generate with OP_CHECKSIG (there's no gain in p2pkh since the pubkey is public after signing block 1) compared to using  signblock + combineblocksigs + submitblock with OP_CHECKMULTISIG 1 of 1 (which seems a little bit absurd in principle, but it is the only option avaliable now for single signer chains afaik).

This will currently fail its own tests, I don't quite get why this isn't working.

Suggestions welcomed, or if you find the bug, that would be much better.

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -104,6 +104,7 @@ testScripts = [
     'signed_blockchain.py',
     'initial_reissuance_token.py',
     'feature_blocksign.py',
+    'feature_onesigner.py',
     'default_asset_name.py',
     'assetdir.py',
 
--- a/qa/rpc-tests/feature_blocksign.py
+++ b/qa/rpc-tests/feature_blocksign.py
@@ -20,6 +20,13 @@ def wif(pk):
 # The signblockscript is a Bitcoin Script k-of-n multisig script.
 def make_signblockscript(num_nodes, required_signers, keys):
     assert(num_nodes >= required_signers)
+
+    if num_nodes == 1 and required_signers == 1:
+        # OP_CHECKSIG = ac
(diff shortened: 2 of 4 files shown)


real-or-random:
`feature_onesigner.py` is missing.

I have only very limited experience with the code base but I'm not convinced that the optimization is worth the hassle and complexity. Do you think that "single signer" will be a common case?

instagibbs:
@real-or-random one bonus to having generate work with single-signer is that we could more easily test the realistic code paths people will run in anger.

caveat: you may be right re:complexity.

@jtimon to be clear, I think this also may work for n-of-m, where the node has all the keys.

real-or-random:
By the way, I have no idea what our plans for threshold sigs are, but they would make this obsolete anyway.

jtimon:
Added the missing file.
I'm not sure it would be all that interesting for multisigs, it is out of the scope of this PR anyway.

I don't think this will be obsoleted for its use case which is basically chains with a single signer.
I'm working on a testnet with a single signer, I can also use 1 of 1 multisig, but this would be more convenient.

real-or-random:
`feature_onesigner.py` needs to be executable and it needs a python shebang.

jtimon:
Fixed nits and rebased, but it still failing and I still don't understand why.

instagibbs:
`JSONRPC error: Could not generate block signed proof with this wallet, use signblock and combineblocksigs for federated chains`

Is it passing locally?

instagibbs:
unfortunately this seems at-odds with upstream, `generate` is literally gone in a version or so, and this call further conflates wallet and mining calls.

Perhaps it can be reworked to `generatewithkey` which takes a privkey as an arg? Closing for now.

Issue #380, opened 2018-06-27
Append initial issuance to 1 address not 100 breaks tutorial code


wintercooled:
#355 means that the tutorial will no longer work as the first send from e1 leaves no funds accessible for e2.

This tutorial: https://github.com/ElementsProject/elements/tree/elements-0.14.1/contrib/assets_tutorial

Could just send all to e1 then get e1 to send half to e2 to fix? Will look into.

instagibbs:
Sounds right. Thanks!

wintercooled:
Looking at fixing this now...

wintercooled:
PR #383 raised to resolve this issue.

wintercooled:
Closed - fixed with #383

PR #383, opened 2018-06-28
Fix issue #380 caused by change to initial free coins distribution

wintercooled:
Fixed issue where e2 couldn't claim anyone-can-spend coins as they all now initially reside in one address. Fixes issue #380


instagibbs:
brief utACK, thanks

PR #384, opened 2018-06-29
Enable setting of default asset name and initial reissuance token count

wintercooled:
Changes to enable you to use Elements as a stand alone blockchain by setting new params:

-initialreissuancetokens
-defaultpeggedassetname

NOTES:
In order to use reissuance tokens for the default asset they must be moved from anyone-can-spend first. 

When using defaultpeggedassetname as a parameter: sendtoaddress, sendmany requires the asset name parameter passing in in order to use your default asset and not 'bitcoin'. Could change this at a later date as part of more serious decoupling, but might require a lot of knock on changes.

EXAMPLE USE:
(based on https://github.com/ElementsProject/elements/blob/elements-0.14.1/contrib/assets_tutorial/assets_tutorial.sh)

`
ASSET="mynewasset"
trap read debug
e1-dae -initialreissuancetokens=200000000 -defaultpeggedassetname=$ASSET
e2-dae -initialreissuancetokens=200000000 -defaultpeggedassetname=$ASSET

BALANCE=$(e1-cli getwalletinfo | jq '.balance')
TOKEN=$(echo $BALANCE | jq 'keys[0]' | tr -d '"') #Should be at [0] but just to be sure:
if [ "$ASSET" = "$TOKEN" ]; then
        TOKEN=$(echo $BALANCE | jq 'keys[1]' | tr -d '"')
fi

e1-cli sendtoaddress $(e1-cli getnewaddress) 21000000 "" "" true $ASSET
e1-cli generate 101
e1-cli sendtoaddress $(e2-cli getnewaddress) 10500000 "" "" false $ASSET
e1-cli generate 101

#Claim reissuance tokens from anyone can spend:
e1-cli sendtoaddress $(e1-cli getnewaddress) 2 "" "" false $TOKEN
e1-cli generate 101

e1-cli reissueasset $ASSET 1000
e1-cli generate 101

e1-cli sendtoaddress $(e2-cli getnewaddress) 1 "" "" false $TOKEN
e1-cli generate 101

e2-cli reissueasset $ASSET 100000
e2-cli generate 101

e1-cli getwalletinfo #Should be 1050100 mynewasset and 1 token
e2-cli getwalletinfo #Should be 1060000 mynewasset and 1 token

#Stop and start to test issue 378 has indeed been fixed for this (was causing errors):
e1-cli stop
e2-cli stop

e1-dae -initialreissuancetokens=200000000 -defaultpeggedassetname=$ASSET
e2-dae -initialreissuancetokens=200000000 -defaultpeggedassetname=$ASSET

e1-cli getwalletinfo #Should be 1050100 mynewasset and 1 token
e2-cli getwalletinfo #Should be 1060000 mynewasset and 1 token

#Done
e1-cli stop
e2-cli stop
`

### DIFF
--- a/qa/rpc-tests/initial_reissuance_token.py
+++ b/qa/rpc-tests/initial_reissuance_token.py
@@ -1,0 +1,88 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Test chain initialisation when specifying number of reissuance tokens to issue.
+#
+
+from decimal import Decimal
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import *
+
+class InitialReissuanceTokenTest(BitcoinTestFramework):
+    """
+    Test creation of initial reissuance token for default asset on chain set up
+
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 2
+
+        #Set number of initial reissuance tokens and also set initial free coins less than max so we can reissue more later
+        self.node_args = [["-initialreissuancetokens=200000000", "-initialfreecoins=2000000000000000"], ["-initialreissuancetokens=200000000", "-initialfreecoins=2000000000000000"]]
+
+    def setup_network(self, split=False):
+        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, self.node_args)
+        connect_nodes_bi(self.nodes,0,1)
+        self.is_network_split = False
+        self.sync_all()
+
+    def run_test(self):
+        #Claim all anyone-can-spend 'bitcoin'
+        self.nodes[0].sendtoaddress(self.nodes[0].getnewaddress(), 20000000, "", "", True)
+        self.nodes[0].generate(101)
+        self.sync_all()
+
+        walletinfo=self.nodes[0].getwalletinfo()
+        balance=walletinfo['balance']
+        token=""
+
+        #Get the hex of the reissuance token
+        for i in balance:
+            token = i
+            if token != "bitcoin":
+                break
+
+        #Claim all anyone-can-spend reissuance tokens
+        self.nodes[0].sendtoaddress(self.nodes[0].getnewaddress(), 2, "", "", False, token)
+        self.nodes[0].generate(101)
+        self.sync_all()
+
+        #Check balances
+        walletinfo1 = self.nodes[0].getwalletinfo()
+        assert_equal(walletinfo1["balance"]["bitcoin"], 20000000)
+        assert_equal(walletinfo1["balance"][token], 2)
+
+        #Reissue some of the default asset
+        self.nodes[0].reissueasset("bitcoin", 1234)
+        self.nodes[0].generate(101)
+        self.sync_all()
+
+        #Check the reissuance worked
+        walletinfo1 = self.nodes[0].getwalletinfo()
+        assert_equal(walletinfo1["balance"]["bitcoin"], 20001234)
+
+        #Send some 'bitcoin' to node 2 so they can fund a reissuance transaction
+        self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 1, "", "", False)
+
+        #Send a reissuance token to node 2 so they can reissue the default asset
+        self.nodes[0].sendtoaddress(self.nodes[1].getnewaddress(), 1, "", "", False, token)
+        self.nodes[0].generate(101)
+        self.sync_all()
+
+        #Reissue some of the default asset
+        self.nodes[1].reissueasset("bitcoin", 1000)
+        self.nodes[1].generate(101)
+        self.sync_all()
+
+        #Check balance is the 1 'bitcoin' sent from node 1 plus the 1000 'bitcoin' reissued by node 2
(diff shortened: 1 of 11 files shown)


instagibbs:
Since this is two logical changes, could you split them into two atomic commits?

src/chainparams.cpp
```diff
@@ -200,7 +200,11 @@ class CCustomParams : public CChainParams {
 
         genesis = CreateGenesisBlock(consensus, strNetworkID, 1296688602, genesisChallengeScript, 1);
         if (initialFreeCoins != 0) {
-            AppendInitialIssuance(genesis, COutPoint(uint256(commit), 0), parentGenesisBlockHash, 1, initialFreeCoins, 0, 0, CScript() << OP_TRUE);
+	    if (initialReissuanceTokens != 0) {
```
instagibbs: I don't think these checks should be nested? Each argument should individually be applied.

instagibbs:
Also, it's best practice to include at least a basic test of new arguments.

Writing a test that would catch the above problem would be one to do. (have initial reissuance tokens but not asset tokens, for instance)

src/chainparams.cpp
wintercooled: Just thinking - if initialFreeCoins is 0 you can't issue any subsequent assets or reissue as you have nothing to fund the tx to create them and I think you end up with a frozen chain (unless you adjust fee settings to zero I guess). So now you mention it having ifc = 0 itself seems like it could cause issues. Hmmm. 

But yes, I can split out and will do. Thanks.

instagibbs: You could always set the reissuance tokens as fee asset :)

wintercooled: ah ha, will try that later, didn't think you could do that!

I'll split over two commits, add a test, change the nested if and snake_case cheers.

instagibbs: I figure the more flexible, the less we have to change later.

src/init.cpp
```diff
@@ -1090,7 +1092,9 @@ bool AppInitParameterInteraction()
     nMaxTipAge = GetArg("-maxtipage", DEFAULT_MAX_TIP_AGE);
 
     try {
-        InitGlobalAssetDir(mapMultiArgs.count("-assetdir") > 0 ? mapMultiArgs.at("-assetdir") : std::vector<std::string>());
+        std::string strDefaultAsset;
+        strDefaultAsset = GetArg("-defaultpeggedassetname", "bitcoin");
```
jtimon: ```const std::string strDefaultAsset = GetArg("-defaultpeggedassetname", "bitcoin");``` ?

src/assetsdir.cpp
```diff
@@ -37,7 +37,7 @@ void CAssetsDir::SetHex(const std::string& assetHex, const std::string& label)
     Set(CAsset(uint256S(assetHex)), AssetMetadata(label));
 }
 
-void CAssetsDir::InitFromStrings(const std::vector<std::string>& assetsToInit)
+void CAssetsDir::InitFromStrings(const std::vector<std::string>& assetsToInit, const std::string& peggedAsset = "bitcoin")
```
jtimon: I don't think we need to default to "bitcoin" here?

jtimon:
utACK modulo nits

src/init.cpp
wintercooled: yep, good idea.

src/assetsdir.cpp
wintercooled: agreed, will remove.

wintercooled:
Update: in testing this I found that if you issue 0 initial assets and > 0 tokens you can set the feeasset as the token which allows you to claim all the tokens and send them around using token as the fee asset. However - you cannot reissue the asset as you get error -4 (insufficient funds). Raised by wallet.cpp as the following returns false:

if (!SelectCoins(vAvailableCoins, mapValueToSelect, setCoins, mapValueIn, coinControl))

Trying to debug this now. Looks like a wallet edge case that was not covered before. Perhaps the token cannot be used to provide the fee when reissueasset is called but it can for sendtoaddress.

instagibbs:
needs rebase

wintercooled:
Issue #390 has been fixed with PR #391. OK to proceed now.

Will rebase and commit changes for this PR tomorrow.

wintercooled:
Reopening - closed by accident as deleted old commits.

wintercooled:
Will push a commit for tests later. Also - will change:
std::string& pegged_asset_name
to 
const std::string& pegged_asset_name
etc.

wintercooled:
Code and test complete and amended following previous reviews.

qa/rpc-tests/initial_reissuance_token.py
```diff
@@ -0,0 +1,89 @@
...
+        self.nodes[0].generate(101)
+        self.sync_all()
+
+        walletinfo=self.nodes[0].getwalletinfo()
+        balance=walletinfo['balance']
+        token=""
+
+        #Get the hex of the reissuance token
+        for i in balance:
+            print(i)
```
instagibbs: no need for print here

wintercooled: Ah yeah, that was me testing - forgot to remove. Shall I change?

instagibbs:
Please

wintercooled:
@instagibbs - done, print removed.

instagibbs:
going to test that this is backwards compatible with previous versions. then will merge.

instagibbs:
It's indeed backwards compatible. @jtimon if you find this PR good to go, merge.

jtimon:
Reviewed, run tests locally, merged.

PR #313, opened 2018-03-12
HF: Move block.proof.challenge to Consensus::Params::signblockscript

jtimon:
Since the scriptPubKey for signing blocks never changes, there's no
point in repeating it with every block header.

Closed https://github.com/ElementsProject/elements/pull/189 by accident, it seems restoring the branch doesn't make it, so I opened this one instead.

Dependencies:

- [x] RPC: Stop providing useless fake pow data #344 [e14-cleanup-pow]
- [x] NO-HF: Move block.proof.challenge to Consensus::Params::signblockscript #386 [e14-blocksign-no-hf]

PR #386, opened 2018-07-10
NO-HF: Move block.proof.challenge to Consensus::Params::signblockscript

jtimon:
Since the scriptPubKey for signing blocks never changes, there's no
point reading repeating it from every block header.

Subset of https://github.com/ElementsProject/elements/pull/313

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -223,14 +223,15 @@ UniValue combineblocksigs(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "Block decode failed");
 
     UniValue result(UniValue::VOBJ);
+    const Consensus::Params& params = Params().GetConsensus();
     const UniValue& sigs = request.params[1].get_array();
     for (unsigned int i = 0; i < sigs.size(); i++) {
         const std::string& sig = sigs[i].get_str();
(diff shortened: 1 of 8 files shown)


src/rpc/mining.cpp
```diff
@@ -774,8 +775,8 @@ UniValue getblocktemplate(const JSONRPCRequest& request)
         result.push_back(Pair("weightlimit", (int64_t)MAX_BLOCK_WEIGHT));
     }
     result.push_back(Pair("curtime", pblock->GetBlockTime()));
-    result.push_back(Pair("signblock_asm", ScriptToAsmStr(pblock->proof.challenge)));
-    result.push_back(Pair("signblock_hex", HexStr(pblock->proof.challenge.begin(), pblock->proof.challenge.end())));
+    result.push_back(Pair("signblock_asm", ScriptToAsmStr(consensusParams.signblockscript)));
+    result.push_back(Pair("signblock_hex", HexStr(consensusParams.signblockscript.begin(), consensusParams.signblockscript.end())));
```
instagibbs: nit: no need for the begin/end stuff, just do like https://github.com/ElementsProject/elements/pull/386/files#diff-a0c8f511d90e83aa9b5857e819ced344R1082

instagibbs:
ACK

Issue #390, opened 2018-07-11
Can't use reissuancetoken as feeasset when trying to reissue an asset. Works for issuance, sendtoaddress etc.


wintercooled:
This is a bit of an edge case but is preventing #384 from being thoroughly tested.

You can use a reissuance token (RIT) as the -feeasset for every function except for reissueasset - as this cause an error -4 "insufficient funds". To test and reproduce: (based on tutorial code):

#!/bin/bash
set -x

shopt -s expand_aliases

alias b-dae="bitcoind -datadir=$HOME/bitcoindir"
alias b-cli="bitcoin-cli -datadir=$HOME/bitcoindir"
alias e1-dae="$HOME/elements/src/elementsd -datadir=$HOME/elementsdir1"
alias e1-cli="$HOME/elements/src/elements-cli -datadir=$HOME/elementsdir1"
alias e2-dae="$HOME/elements/src/elementsd -datadir=$HOME/elementsdir2"
alias e2-cli="$HOME/elements/src/elements-cli -datadir=$HOME/elementsdir2"

b-cli stop
e1-cli stop
e2-cli stop
sleep 5

cd
rm -r ~/bitcoindir ; rm -r ~/elementsdir1 ; rm -r ~/elementsdir2
mkdir ~/bitcoindir ; mkdir ~/elementsdir1 ; mkdir ~/elementsdir2

cd elements
cd src

cp ~/elements/contrib/assets_tutorial/bitcoin.conf ~/bitcoindir/bitcoin.conf
cp ~/elements/contrib/assets_tutorial/elements1.conf ~/elementsdir1/elements.conf
cp ~/elements/contrib/assets_tutorial/elements2.conf ~/elementsdir2/elements.conf

sleep 1

b-dae
e1-dae
e2-dae

sleep 5

e1-cli sendtoaddress $(e1-cli getnewaddress) 21000000 "" "" true 
e1-cli generate 101
e1-cli sendtoaddress $(e2-cli getnewaddress) 10500000 "" "" false 
e1-cli generate 101

ISSUE=$(e1-cli issueasset 100 1 false)

ASSET=$(echo $ISSUE | jq '.asset' | tr -d '""')
RITOKEN=$(echo $ISSUE | jq '.token' | tr -d '""')
e1-cli generate 101

echo $RITOKEN
e1-cli getwalletinfo
e2-cli getwalletinfo

e1-cli stop
e2-cli stop

sleep 5
[...]

PR #391, opened 2018-07-11
subtract fee from reissuance token if token is fee asset during reiss…

instagibbs:
…uance

should fix https://github.com/ElementsProject/elements/issues/390#issuecomment-404212380


wintercooled:
Tested and passed all test in #390

instagibbs:
had to restart the last job due to weird timeout, passed CI

Issue #387, opened 2018-07-10
listissuances doesn't accept label, only hex - which is a little inconsistent


wintercooled:
Hardly major but as other commands accept either label string or hex maybe this could be amended to also accept the label text and not just the hex.

So as well as:

listissuances b2e15d0d7a0c94e4e2ce0fe6e8691b9e451377f6e46e8045a86f7c4b5d4f0f23

You could do:

listissuances myassetname

As I said - not major but will make the command more consistent with others like reissueasset etc.

wintercooled:
I'm looking at this.

wintercooled:
Closing - fixed with PR #393

PR #393, opened 2018-07-16
Add ability to use listissuance with label as param. Plus tests.

wintercooled:
Addresses Issue #387 

This enables you to pass the label of an asset into listissuances. Previously you had to use a hex.

Added new test that covers assetdir in general and this PR.

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -105,6 +105,7 @@ testScripts = [
     'initial_reissuance_token.py',
     'feature_blocksign.py',
     'default_asset_name.py',
+    'assetdir.py',
 
     # Elements' specially adapted tests second
     'blockchain.py',
--- a/qa/rpc-tests/assetdir.py
+++ b/qa/rpc-tests/assetdir.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#
+# Test use of assetdir to locally label assets.
+# Test listissuances returns a list of all issuances or specific issuances based on asset hex or asset label.
+#
+
+from decimal import Decimal
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.authproxy import JSONRPCException
+from test_framework.util import *
+
+class AssetdirTests(BitcoinTestFramework):
+    """
+    Test use of assetdir to specify asset labels.
+    Test listissuances returns a list of all issuances or specific issuances based on asset hex or asset label.
+    """
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+
+    def setup_network(self, split=False):
+        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
+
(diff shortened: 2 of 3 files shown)


qa/rpc-tests/assetdir.py
```diff
@@ -0,0 +1,86 @@
...
+        assert_equal(issuances[0]["assetlabel"], "asset1")
+
+        #Check we can get a list of issuances for a given hex
+        issuances=self.nodes[0].listissuances(asset2hex)
+        assert_equal(len(issuances), 1)
+        assert_equal(issuances[0]["assetlabel"], "asset2")
+
+        #Check a non-existant label or hex raises an error
+        try:
+            self.nodes[0].listissuances("doesnotexist")
```
instagibbs: you should catch the case where no error triggers. You can set `rpc_error = None` then save the error in the Exception block, and check against that after for example/

qa/pull-tester/rpc-tests.py
```diff
@@ -105,6 +105,7 @@
     'initial_reissuance_token.py',
     'feature_blocksign.py',
     'default_asset_name.py',
+    'assetdir.py',
```
instagibbs: do we plan on porting some of confidential_transactions.py here, based on assetdir stuff? Might be a good idea.

qa/rpc-tests/assetdir.py
```diff
@@ -0,0 +1,86 @@
...
+        except Exception as e:
+            assert_equal(e.error["code"], -4)
+
+        try:
+            self.nodes[0].listissuances("9b4d313040db510e74cf0b50ae6ef140eee2e4eca2a1efcfcd87d9997ef0a235")
```
instagibbs: where is this hex from? Could just do all 0's to make sure it's never valid on its face.

wintercooled: Randomly generated. Will use 0s as suggested.

qa/rpc-tests/assetdir.py
wintercooled: Will do.

qa/pull-tester/rpc-tests.py
wintercooled: I nearly added my tests to confidential_transactions.py as it already had some issuance code in there but it already seemed to be creeping beyond the original intent. So yes, perhaps moving some of the tests in ct into asserdir would be a good idea. Perhaps renaming it to issuance_reissuance.py as well?

instagibbs: Depends on what scope you intend this test to arrive at. We can always rename later of course.

wintercooled:
Updated based on previous comments. Ready for review.

src/wallet/rpcwallet.cpp
```diff
@@ -3958,7 +3958,7 @@ UniValue listissuances(const JSONRPCRequest& request)
             "listissuances ( asset ) \n"
             "\nList all issuances known to the wallet for the given asset, or for all issued assets if none provided.\n"
             "\nArguments:\n"
-            "1. \"asset\"                 (string, optional) The asset whose issaunces you wish to list.\n"
+            "1. \"asset\"                 (string, optional) The asset whose issaunces you wish to list. Accepts both the asset hex or the locally assigned asset label.\n"
```
instagibbs: sorry grammar nit: "both ... or". Should be "either ... or" or "both ... and". "Either or" seems more logical to me since it's one or the other.

PR #394, opened 2018-07-17
Explicit option for disabling 2wp

jtimon:
- [X] Chainparams: Introduce -con_has_parent_chain option
- [x] Make sure -con_has_parent_chain disables 2wp


jtimon:
Tested pegging.py doesn't work if I add 

```
        f.write("con_has_parent_chain=0\n")
        f.write("parentgenesisblockhash=00\n")
```

or only 

```
        f.write("con_has_parent_chain=0\n")
```

to https://github.com/ElementsProject/elements/blob/elements-0.14.1/qa/rpc-tests/pegging.py#L123

Happy to test more if somebody comes up with something else to try.

src/chainparams.cpp
```diff
@@ -134,6 +134,10 @@ class CCustomParams : public CChainParams {
         parentGenesisBlockHash = uint256S(GetArg("-parentgenesisblockhash", "0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206"));
         initialFreeCoins = GetArg("-initialfreecoins", 0);
         initial_reissuance_tokens = GetArg("-initialreissuancetokens", 0);
+        consensus.has_parent_chain = GetBoolArg("-con_has_parent_chain", true);
+        // Either it has a parent chain or not
+        const bool parent_genesis_is_null = parentGenesisBlockHash == uint256();
+        assert(consensus.has_parent_chain + parent_genesis_is_null == 1);
```
gwillen: I would probably say "^" instead of "+" here to make it clear what you mean (either one or the other must be true, but not both), assuming they are both bools.

gwillen:
utACK, what's here looks reasonable to me -- I'm not familiar enough with this part of the codebase to know if there's more that's missing and should be included.

src/chainparams.cpp
gwillen: Oh, apparently a popular idiom for 'logical xor' is to use !=, which would also work if they're both bools.

AkioNak:
Even if ``` con_has_parent_chain=0 ``` and ``` parentgenesisblockhash=00 ``` are designated ,  ``` validatepegin=1 ``` is still in the elements.conf , so I think that ``` sidechain2 ``` will not work.
( ``` bool BitcoindRPCCheck(const bool init) ``` may return ``` false ```. )

https://github.com/ElementsProject/elements/blob/elements-0.14.1/qa/rpc-tests/pegging.py#L136

jtimon:
@gwillen fixed nit

@AkioNak Right, sidechain2 should not work on that test if con_has_parent_chain=0 and parentgenesisblockhash=00. Even if validatepegin=0 the test fails with: "Peg-in has failed to propagate."

If I set 

```
        f.write("validatepegin=0\n")
        f.write("con_has_parent_chain=0\n")
        f.write("parentgenesisblockhash=00\n")
```

in both sidechain and sidechain2, the test fails with:
```
Starting daemons at /tmp/bitcoin_751D61AC41, /tmp/bitcoin2_732B29F0EC, /tmp/sidechain_A4E0895F5B and /tmp/sidechain2_011894F79A
Daemons started
Daemons started, making blocks to get funds
Attempting peg-in
Pegging testing failed, aborting:
```

The idea is that pegins shouldn't work with this option, perhaps I should write a separated test.

PR #397, opened 2018-07-25
Tests: Explicitly destroy zmq object when exiting after not actually using it

jhfrontz:
When run in a docker Ubuntu instance under Centos 7, the script `qa/rpc-tests/zmq_test.py` will typically hang indefinitely (even though it has reported a successful -- albeit mostly stubbed-out -- completion). Using gdb to debug the [core file manually generated from the] associated python instance reveals that the main thread is waiting for a call to zmq_ctx_term() to end -- a scenario similar to  that described in https://github.com/zeromq/libzmq/issues/2586

This change adds the workaround described in the issue (namely to explicitly call `self.zmqContext.destroy(linger=0)` after the process is finished with zmq).

Note: the workaround is otherwise innocuous for any environment not experiencing a hang.

### DIFF
--- a/qa/rpc-tests/zmq_test.py
+++ b/qa/rpc-tests/zmq_test.py
@@ -34,6 +34,9 @@ class ZMQTest (BitcoinTestFramework):
             ])
 
     def run_test(self):
+        # destroy() required for ubuntu docker hang workaround when
+        # exiting without actually messaging (remove with resolution of #TODO)
+        self.zmqContext.destroy(linger=0)
         return #TODO
         self.sync_all()
 


instagibbs:
Why is this bug triggering only due to early stop? Or is this upstreamable?

jhfrontz:
From what I've been able to discern, the problem has something to do with a listener (within the zmq infrastructure) that waits forever (the `linger=0` seems intended to circumvent that; but setting that parameter on a per-socket basis alone doesn't resolve it entirely, thus requiring the explicit `destroy` of the context).  My supposition is that the window exposing the problem is cleared if there is ever any actual communication performed on the connection.

instagibbs:
makes your life easier ACK

jtimon:
I still think it would be preferable to uncomment at least part of the tests or comment the whole test in the test runner. But "not opposed to this ACK"

jhfrontz:
@jtimon I'm willing to try to make the test do more work (I think it already performs some rudimentary testing by virtue of verifying that the zmq package has been properly installed), but I need some guidance on how to enable more functionality.

Issue #369, opened 2018-06-11
Need to address merkle leaf detection weakness in peg-in proof


instagibbs:
https://bitslog.wordpress.com/2018/06/09/leaf-node-weakness-in-bitcoin-merkle-tree-design/

Most coherent solution would be to include the coinbase in the inclusion proof, but this would make the claim size dependent on miners instead of the peg-maker. Alternatively we could employ the single-sha2 variant, for coinbase transactions larger than 64 bytes.

Another issue is we should be fixing our own merkle trees for spv/peg uses, possibly with prefixing as mentioned in the article.

instagibbs:
use this in 0.16.2+: https://github.com/bitcoin/bitcoin/pull/13451

PR #400, opened 2018-08-09
2WP: Verify number of transactions in SPV proof

stevenroose:
Use the getblockheader RPC from Bitcoin Core to verify if the number of
transactions mentioned in the SPV proof matches the number for the
block known by the Core daemon.

Inspired by https://github.com/bitcoin/bitcoin/commit/6b9dc8ceaed597d9c539ba6b09c171b258b66ca3

Partially fixes https://github.com/ElementsProject/elements/issues/369

### DIFF
--- a/src/callrpc.cpp
+++ b/src/callrpc.cpp
@@ -178,7 +178,7 @@ UniValue CallRPC(const std::string& strMethod, const UniValue& params, bool conn
     return reply;
 }
 
-bool IsConfirmedBitcoinBlock(const uint256& hash, int nMinConfirmationDepth)
+bool IsConfirmedBitcoinBlock(const uint256& hash, const int nMinConfirmationDepth, const int nbTxs)
 {
     try {
         UniValue params(UniValue::VARR);
@@ -189,8 +189,21 @@ bool IsConfirmedBitcoinBlock(const uint256& hash, int nMinConfirmationDepth)
         UniValue result = find_value(reply, "result");
         if (!result.isObject())
             return false;
-        result = find_value(result.get_obj(), "confirmations");
-        return result.isNum() && result.get_int64() >= nMinConfirmationDepth;
+
+        UniValue confirmations = find_value(result.get_obj(), "confirmations");
+        if (!confirmations.isNum() || confirmations.get_int64() < nMinConfirmationDepth) {
+            return false;
+        }
+
+        // Only perform extra test if nbTxs has been provided (non-zero).
+        if (nbTxs != 0) {
+            UniValue nTx = find_value(result.get_obj(), "nTx");
+            if (!nTx.isNum() || nTx.get_int64() != nbTxs) {
+                LogPrintf("ERROR: Invalid number of transactions in merkle block for %s", 
+                        hash.GetHex());
+                return false;
+            }
+        }
     } catch (CConnectionFailed& e) {
         LogPrintf("ERROR: Lost connection to bitcoind RPC, you will want to restart after fixing this!\n");
         return false;
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2531,6 +2531,7 @@ bool IsValidPeginWitness(const CScriptWitness& pegin_witness, const COutPoint& p
 
     uint256 block_hash;
     uint256 tx_hash;
+    int num_txs;
     // Get txout proof
     if (Params().GetConsensus().ParentChainHasPow()) {
 
@@ -2546,6 +2547,8 @@ bool IsValidPeginWitness(const CScriptWitness& pegin_witness, const COutPoint& p
         if (!CheckPeginTx(stack[4], pegtx, prevout, value, claim_script)) {
             return false;
         }
+
+        num_txs = merkle_block_pow.txn.GetNumTransactions();
     } else {
 
         CMerkleBlock merkle_block;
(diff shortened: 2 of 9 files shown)


src/callrpc.cpp
```diff
@@ -189,8 +189,11 @@ bool IsConfirmedBitcoinBlock(const uint256& hash, int nMinConfirmationDepth)
         UniValue result = find_value(reply, "result");
         if (!result.isObject())
             return false;
-        result = find_value(result.get_obj(), "confirmations");
-        return result.isNum() && result.get_int64() >= nMinConfirmationDepth;
+
+        UniValue confirmations = find_value(result.get_obj(), "confirmations");
+		UniValue nTx = find_value(result.get_obj(), "nTx");
```
instagibbs: spacing

stevenroose: Ah. Didn't configure Vim to expand tabs for C++ I think.

stevenroose:
@instagibbs didn't notify you last week that I added the issue. Also removed the WIP tag now.

src/callrpc.cpp
```diff
@@ -189,8 +189,11 @@ bool IsConfirmedBitcoinBlock(const uint256& hash, int nMinConfirmationDepth)
         UniValue result = find_value(reply, "result");
         if (!result.isObject())
             return false;
-        result = find_value(result.get_obj(), "confirmations");
-        return result.isNum() && result.get_int64() >= nMinConfirmationDepth;
+
+        UniValue confirmations = find_value(result.get_obj(), "confirmations");
+        UniValue nTx = find_value(result.get_obj(), "nTx");
+        return confirmations.isNum() && confirmations.get_int64() >= nMinConfirmationDepth
+            && (nbTxs == 0 || nTx.get_int64() == nbTxs);
```
instagibbs: it might be pretty interesting to actually print a relevant line for this part failing. Would be nice to record an actual attack.

instagibbs: probably should call `isNum` on it before fetching the value

instagibbs:
I don't think this works. The proof can claim any number of transactions: https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/primitives/bitcoin/merkleblock.h#L94

It's only used to calculate the tree width. https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/primitives/bitcoin/merkleblock.h#L72

What we need to do is take the known number of transactions, compute how many levels of proof the merkle block *should* have, and compare it to the actual proof we're parsing.

instagibbs:
Actually no I'm silly. This is completely sufficient since we're calling `ExtractMatches` (duh). It's the same fix as linked in OP.

utACK

jtimon:
utACK.
From what version do getblockheader and getblock have nTx ?
Can we have it elements too before doing this? (otherwise https://github.com/ElementsProject/elements/pull/362 will stop working, perhaps we can make it before that one instead, but merging that one first helps testing this).

instagibbs:
@jtimon Yes please backport, we're not in a rush here.

instagibbs:
One thing I think is missing here is that on startup(BitcoindRPCCheck or something) we need to make sure that it's a new enough version of Core to continue starting up.

stevenroose:
I added the check for version v0.16.2. It's true this conflicts with #362.

instagibbs:
restarted job, random functional test failing to stop a node

src/validation.cpp
```diff
@@ -2248,6 +2250,21 @@ bool BitcoindRPCCheck(const bool init)
                     }
                 }
                 UniValue result = reply["result"];
+                if (!result.isObject() || !result.get_obj()["version"].isNum() ||
+                        result.get_obj()["version"].get_int() < MIN_PARENT_NODE_VERSION) {
+                    LogPrintf("ERROR: Parent chain daemon too old; need version 0.16.2.\n");
```
instagibbs: "Bitcoin Core 0.16.2 or newer"

instagibbs:
Provided you update the string, tACK https://github.com/ElementsProject/elements/pull/400/commits/4552fa5feba5e7a2bd22ec0c68378948a49a9fad

instagibbs:
It properly exits when running 0.16.1, and continues on 0.16.2

src/validation.cpp
```diff
@@ -2248,6 +2250,22 @@ bool BitcoindRPCCheck(const bool init)
                     }
                 }
                 UniValue result = reply["result"];
+                if (!result.isObject() || !result.get_obj()["version"].isNum() ||
+                        result.get_obj()["version"].get_int() < MIN_PARENT_NODE_VERSION) {
+                    LogPrintf("ERROR: Parent chain daemon too old; "
+                            "need BitcoinCore version 0.16.2 or newer.\n");
```
instagibbs: Space between Bitcoin and Core, sorry :)

stevenroose: Whoops :/

jtimon:
Can you rebase after #362 has been merged just to make sure travis remains happy?

instagibbs:
failed to compile?

stevenroose:
Oh, Jorge's change broke something.

stevenroose:
It works now! (Tests are fixed!)

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/400/commits/85eb5026de26947a4c4d48db3b90af787ea69e38

Issue #408, opened 2018-09-10
policyAsset not set in c++ unit tests


nkostoulas:
Policy/fee asset doesn't seem to be set in the src/test unit-tests. I noticed this in txvalidationcache_tests.cpp, where the fee output was discarded when the tx was entered in the mempool as in AcceptToMemoryPoolWorker the following returns 0:

`CAmount nFees = tx.GetFee()[policyAsset];`

Although it doesn't break anything as zero fees are allowed, it might hide intended functionality if something changed. Do you think it would make sense to add this in the test_bitcoin configuration?

instagibbs:
yes, thank you

instagibbs:
Do you have a backtrace for the issue?

nkostoulas:
```
FAIL: test/test_elements
========================

Running 210 test cases...
unknown location:0: fatal error: in "tx_validationcache_tests/tx_mempool_block_doublespend": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000000)
test/txvalidationcache_tests.cpp:70: last checkpoint

*** 1 failure is detected in the test module "Bitcoin Test Suite"
Assertion failed: (!pthread_mutex_destroy(&m)), function ~recursive_mutex, file /usr/local/include/boost/thread/pthread/recursive_mutex.hpp, line 104.
FAIL test/test_elements (exit status: 134)
```

Is there a way to get a detailed trace?

instagibbs:
Does your change fix it? That's a good indicator!

nkostoulas:
Yea it does! I was just curious why it was happening.

instagibbs:
We can keep this issue open if you like even if we merge the fix.

PR #409, opened 2018-09-12
Set policyAsset for core unit tests

nkostoulas:
Set policyAsset for all core c++ unit tests #408 

src/test/txvalidationcache_tests.cpp was failing with memory access violation somewhere in block creation. I believe it's because now that mempool entry fee is correctly set, the fact that the fee output was larger than the non-fee one was causing an issue. I haven't been able to track this down exactly though. Please let me know if you have some idea about this and if the change makes sense.

### DIFF
--- a/src/test/txvalidationcache_tests.cpp
+++ b/src/test/txvalidationcache_tests.cpp
@@ -45,10 +45,10 @@ BOOST_FIXTURE_TEST_CASE(tx_mempool_block_doublespend, TestChain100Setup)
         spends[i].vin[0].prevout.hash = coinbaseTxns[0].GetHash();
         spends[i].vin[0].prevout.n = 0;
         spends[i].vout.resize(2);
-        spends[i].vout[0].nValue = 11*CENT;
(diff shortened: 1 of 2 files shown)


jtimon:
utACK 2bf55e1d0a91fdb34e1af8e84a67e5e01fb62818

src/test/txvalidationcache_tests.cpp
```diff
@@ -45,10 +45,10 @@ BOOST_FIXTURE_TEST_CASE(tx_mempool_block_doublespend, TestChain100Setup)
         spends[i].vin[0].prevout.hash = coinbaseTxns[0].GetHash();
         spends[i].vin[0].prevout.n = 0;
         spends[i].vout.resize(2);
-        spends[i].vout[0].nValue = 11*CENT;
+        spends[i].vout[0].nValue = MAX_MONEY-11*CENT;
```
instagibbs: why this change?

instagibbs: nevermind

PR #414, opened 2018-09-21
WIP: Support signet as parent chain

jtimon:
signet has simpler signed blocks but no CT or any other extra feature from elements.

It won't work unless
https://github.com/bitcoin/bitcoin/commit/e416403ddc15225507d6bf6b5833f8f63c524278#r30586628
is addressed

In...  bool BlockSignatureCreator::CreateSig(const SigningProvider& provider, std::vector<unsigned char>& vchSig, const CKeyID& keyid, const CScript& scriptCode, SigVersion sigversion) const

46+    vchSig.push_back((unsigned char)SIGHASH_ALL);

"If we don't push here we don't need to pop in
e416403#diff-be2905e2f5218ecdbe4e55637dac75f3R1430 do we?"

ping @kallewoof


src/pow.cpp
```diff
@@ -38,16 +38,39 @@ void ResetChallenge(CBlockHeader& block, const CBlockIndex& indexLast, const Con
...
+                                               SCRIPT_VERIFY_CHECKSEQUENCEVERIFY |
+                                               SCRIPT_VERIFY_LOW_S |
+                                               SCRIPT_VERIFY_WITNESS |
+                                               SCRIPT_VERIFY_DISCOURAGE_UPGRADABLE_WITNESS_PROGRAM |
+                                               SCRIPT_VERIFY_WITNESS_PUBKEYTYPE;
+
+bool CheckBitcoinProof(const uint256& hash, unsigned int nBits, const Consensus::Params& params)
 {
+    if (g_solution_blocks) {
+        if (hash == params.hashGenesisBlock) return true;
```
jtimon: I guess we don't need or want this here.

jtimon:
This can now be tested with:
```
python3 ./qa/pull-tester/rpc-tests.py feature_fedpeg --parent_type=signet --parent_binpath=/home/user/code/bitcoin/src/bitcoind
```

Assuming you're building https://github.com/kallewoof/bitcoin/commits/signet ( b4c9e7838f16e1947804e6cb12c2282bcb0f9deb ) on ```/home/user/code/bitcoin/```.

The changes to feature_fedpeg shouldn't break other ways to call the tests:
```
python3 ./qa/pull-tester/rpc-tests.py feature_fedpeg
python3 ./qa/pull-tester/rpc-tests.py feature_fedpeg --parent_type=bitcoin --parent_binpath=/home/jt/data/programs/binaries/bitcoin-0.16.3/bin/bitcoind
python3 ./qa/pull-tester/rpc-tests.py feature_fedpeg --parent_type=bitcoin --parent_binpath=/home/jt/data/programs/binaries/bitcoin-0.17.0/bin/bitcoind
```

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -74,21 +60,40 @@ def setup_network(self, split=False):
...
                     "-printtoconsole=0",
                     '-validatepegin=0',
                     '-anyonecanspendaremine',
                     '-initialfreecoins=2100000000000000',
                 ])
+
             self.binary = self.options.parent_binpath if self.options.parent_binpath != "" else None
             self.nodes.append(start_node(n, self.options.tmpdir, self.extra_args[n], binary=self.binary, chain=self.parent_chain))
+            if self.options.parent_type == 'signet':
+                self.nodes[n].importprivkey('8Hh8jNjkx1aSCgEk3iq9Vo2APZUSDQJVt3rJ2BRpb5Tavqb68vW')
```
stevenroose: Is this magic value related to the one in `-signet_blockscript`? Perhaps define them as variables somewhere earlier then together so that it's clear they are related.

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -140,13 +153,21 @@ def test_pegout(self, parent_chain_addr, sidechain):
                 break
         assert pegout_tested
 
+    # TODO Remove when signet supports generate
```
stevenroose: Is there an upstream issue to track for this?

src/primitives/bitcoin/block.h
```diff
@@ -12,6 +12,22 @@
...
+/**
+ * If non-zero, defines an enforced size requirement for block header payloads.
+ * It requires that all blocks are of size 80 + (this value) bytes.
+ */
+extern size_t g_solution_block_len;
+ /**
+ * Contains a mapping of hash to signature data for each block header
+ * in signet networks.
+ */
+extern std::map<uint256,std::vector<uint8_t>> g_blockheader_payload_map;
```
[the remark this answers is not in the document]
instagibbs: An in-memory cache that, most importantly, doesn't inflate memory usage when signet feature is off.

This does mean an ever-growing set when on, yes. Wonder what cane be done with this.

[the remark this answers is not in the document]
instagibbs: @kallewoof thoughts?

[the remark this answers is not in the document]
kallewoof: It's ever growing, yep. It could be cached since header sigs are stored along with headers. I haven't thought too much about it, as even a 1 million block chain (that's 2x the lifetime of bitcoin right now) would only have ~100 MB of RAM use.

qa/rpc-tests/feature_fedpeg.py
kallewoof: This will eventually go away, I assume, unless we agree on a global default signet (seems unlikely, at least for upstream to bitcoin/bitcoin).

src/primitives/bitcoin/block.h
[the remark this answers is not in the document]
stevenroose: Also when it's a 11-of-15 multisig? Or even larger if this script doesn't have the same script limits as normal tx scripts.

qa/rpc-tests/feature_fedpeg.py
jtimon: No, all we have for now is https://github.com/bitcoin/bitcoin/compare/master...kallewoof:signet where I'm commenting directly on the code.
Pinging @kallewoof since I can't remember if I suggested this to him.

qa/rpc-tests/feature_fedpeg.py
jtimon: yeah, we could generate the block signing script and the wif dynamically by starting a node, getnewaddress, dumpprivkey, etc beforehand and then starting the actual testing nodes with the resulting script.
This is just the wif for the public key corresponding to the script.

jtimon:
Now we can review signet more easily with https://github.com/kallewoof/bitcoin/pull/4

qa/rpc-tests/feature_fedpeg.py
jtimon: This is what I was using for generating the p2pk sript and the corresponding wif to import:

https://github.com/jtimon/elements/commit/3a597eb472868affd94c470568cbedd90a6e42f7

jtimon:
Needed rebase. Also, removed one TODO and its workaround in the tests since now generate works directly with signet.

src/pow.cpp
kallewoof: You would need to provide the signature for the genesis block otherwise, no?

qa/rpc-tests/feature_fedpeg.py
kallewoof: I intend to make a PR against bitcoin core. For now there is a PR against my repo here: https://github.com/kallewoof/bitcoin/pull/4

src/primitives/bitcoin/block.h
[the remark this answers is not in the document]
kallewoof: Right you would basically store siglen + alpha per block. For a 15 multisig over ECDSA it'd get big quicker, I suspect, but I assume we will eventually switch to Schnorr multisigs long term.

src/pow.cpp
jtimon: No, this is used to validate the parent chain. The sidechain doesn't validate the parent chain's genesis block (in fact, not even the parent chain should validate the genesis block, the genesis block is a consensus rule in itself and thus doesn't need to be validated, not when it's pow nor when it's blocksigned).

jtimon:
Moved to https://github.com/kallewoof/bitcoin/pull/5 and removed the special case for the p2sh test for signet.

jtimon:
Updated, now it doesn't have the wif and blockscript hardcoded but generated aumotatically

jtimon:
Fixed travis.

stevenroose:
tACK 23cb9ae4d26e36bb568bfcd93e340063e46f3405 with parent 2aca2a99154c850843ce6be6474ec918a1dce0f3 (kallewoof/signet-0.17)

stevenroose:
That would be ideal of course. Could be as simple as the Bitcoin extra Travis thing. But then there need to be some kind of "stable" signet release..

instagibbs:
@stevenroose I'm simply reticent to merge something we can't be confident of stability of the protocol itself.

@kallewoof any plans on "finalizing" this, whatever that means to you? :) We clearly can adapt implementation later to later upgrades.

stevenroose:
Maybe when the elements-0.17 work is done, signet is "released" and we can download a stable binary. I wouldn't mind to keep this open. Or either way just have a custom-compiled signet binary in Travis.

jtimon:
Added a commit (not squashed yet) supposedly adapting the test to https://github.com/ElementsProject/elements/pull/433 as parent chain binary.

But even though everything else should have stayed the same, travis seems to be failing.

instagibbs:
@kallewoof have you or do you have intentions of releasing binaries for signet? We could at least get this ported to `master` for support/iteration.

kallewoof:
@instagibbs Sorry for lack of activity. I am intending on proposing a BIP for signet in the near future. Things will be switched around a bit, I suspect, so no binaries yet.

instagibbs:
Ok just curious. Still interested at least in supporting as parent chain.

PR #415, opened 2018-09-24
qa: Support using bitcoind in fedpeg test 

stevenroose:
This ports the functionality from pegging.py to the feature_fedpeg.py
test.  The following two commands are not equivalent:

```
$ ./qa/rpc-tests/pegging.py dir/to/bitcoin
```

```
$ ./qa/pull-tester/rpc-tests.py feature_fedpeg --parent_bitcoin \
    --parent_binpath=bin/to/bitcoin/bitcoind
```


qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,9 +49,30 @@ def __init__(self):
         self.setup_clean_chain = True
         self.num_nodes = 4
 
-    def setup_network(self, split=False):
+    def add_options(self, parser):
+        parser.add_option("--bitcoinddir", dest="bitcoinddir", default="",
```
jtimon: perhaps use --binpath or something directly instead of appending bitcoind in the test?
That way we could also use another elements binary, perhaps an older one to test compatibility or a newer one for testing rebases.
For that we would need to decouple another option --parent_is_bitcoinlike or something of the sort.

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,9 +49,30 @@ def __init__(self):
...
+            "-testnet=0",
+            "-txindex=1",
+            "-addresstype=legacy", # To make sure bitcoind gives back p2pkh no matter version
+            "-connect=localhost:"+str(p2p_port((n+1)%2)),
+            "-listen=1",
+            #"-debug"
+        ] for n in range(2)]
+
+        binary = os.path.join(self.options.bitcoinddir, "bitcoind")
+        self.nodes = start_nodes(2, self.options.tmpdir, self.extra_args[:2], binary=binary, chain="regtest")
```
jtimon: there's still no -chain=regtest option in normal bitcoind binaries, so in https://github.com/ElementsProject/elements/blob/elements-0.14.1/qa/rpc-tests/test_framework/util.py#L343 you will need to replace '-chain='+chain with just '-regtest' if chain == 'regtest'

stevenroose: Yeah I wanted to try. Is this why it fails? I thought it might just ignore it.

qa/rpc-tests/feature_fedpeg.py
stevenroose: Changed to `--parent_bitcoin` and `--parent_binpath=` :)

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,27 +49,57 @@ def __init__(self):
...
-            # '-printtoconsole',
-            '-validatepegin=0',
-            '-anyonecanspendaremine',
-            '-initialfreecoins=2100000000000000',
-        ]] * 2
-
-        self.nodes = start_nodes(2, self.options.tmpdir, self.extra_args[:2], chain='parent')
+        chain = None
+        if self.options.parent_bitcoin:
+            if self.options.parent_binpath == "":
```
instagibbs: Shouldn't this be filtered in rpc-tests.py as well? This way a top-level arg there could trigger this, or not, safely.

stevenroose: I don't understand your point..

instagibbs: rpc-tests.py will run all tests in sequence. I'd rather catch this argument error up front instead of waiting for this to be run.

qa/rpc-tests/test_framework/util.py
```diff
@@ -340,7 +340,9 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
     datadir = os.path.join(dirname, "node"+str(i))
     if binary is None:
         binary = os.getenv("ELEMENTSD", "elementsd")
-    args = [ binary, '-chain='+chain, "-datadir="+datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-mocktime="+str(get_mocktime()) ]
+    args = [ binary, "-datadir="+datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-mocktime="+str(get_mocktime()) ]
+    if chain != "regtest": #TODO(stevenroose) remove if unnecessary or when bitcoind has -chain
+        args.append("-chain="+chain)
```
jtimon: why not?
```
    # Special case for regtest. Remove if bitcoind gets -chain
    if chain == 'regtest': 
        args.append('-regtest')
    else:
        args.append('-chain=' + chain)
```

jtimon: Or perhaps
```
args.append('-regtest' if chain == 'regtest' else: '-chain=' + chain)
```
or some more pythonic way to do it (haven't tried this one, but I know there's something like the '?' keyword in C, java and friends).

stevenroose: Yeah I made it conditional and put a todo to remove it :)

stevenroose:
Like @jtimon pointed out, this also allows to use a custom elements binary for the parent chain. F.e. any liquid binary.

qa/rpc-tests/feature_fedpeg.py
stevenroose: Oh, you can pass these arguments to rpc-tests.py and they are globally set for all tests??

instagibbs: That was my hope, yes

Should be able to just make it some global(?)

stevenroose: Euhm, it seems that that's not possible anyway. If you run rpc-tests.py with those arguments, all other tests will complain that they don't know the arguments:
```
create_cache.py: error: no such option: --parent_bitcoin
```

Also, I think for now it makes more sense to only use those flags when running this test specifically..

instagibbs: meh ok

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,27 +49,61 @@ def __init__(self):
         self.setup_clean_chain = True
         self.num_nodes = 4
 
+    def add_options(self, parser):
+        parser.add_option("--parent_binpath", dest="parent_binpath", default="",
+                          help="Use a different binary for launching nodes")
+        parser.add_option("--parent_bitcoin", dest="parent_bitcoin", default=False, action="store_true",
+                          help="Parent nodes are Bitcoin")
+
+    def start_node(self, idx):
```
jtimon: I'm not so sure this adds a lot of readibility, you could just use:

```
start_node(n, self.options.tmpdir, self.extra_args[n], chain='sidechain')
```
in one place and

```
binary = self.options.parent_binpath if self.options.parent_binpath != "" else None
chain = "regtest" if self.options.parent_bitcoin else "parent"
start_node(n, self.options.tmpdir, self.extra_args[n], binary=binary, chain=chain)
```
In the other, at least for the sidechains case seems unnecessarily complicated.

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,27 +49,61 @@ def __init__(self):
...
+        is_parent = idx < 2
```
instagibbs: this isn't very readable, add a comment or make it more explicit in code?

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -143,14 +183,15 @@ def run_test(self):
...
             assert("Peg-in Bitcoin transaction needs more confirmations to be sent." in e.error["message"])
 
         # Should fail due to non-witness
         try:
-            pegtxid = sidechain.claimpegin(raw, proof, get_new_unconfidential_address(parent))
+            pegtxid = sidechain.claimpegin(raw, proof, get_new_unconfidential_address(sidechain))
             raise Exception("Peg-in with non-matching claim_script should fail.")
         except JSONRPCException as e:
-            print(e.error["message"])
+            print('ERROR:', e.error)
```
jtimon: Are you sure removing ['message'] makes the printed error more clear?

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,27 +49,61 @@ def __init__(self):
...
+
+    def start_node(self, idx):
+        is_parent = idx < 2
+        binary = self.options.parent_binpath if is_parent and self.options.parent_binpath != "" else None
+        chain = "sidechain" if not is_parent else "regtest" if self.options.parent_bitcoin else "parent"
+        return start_node(idx, self.options.tmpdir, self.extra_args[idx], binary=binary, chain=chain)
+
     def setup_network(self, split=False):
+        if self.options.parent_bitcoin and self.options.parent_binpath == "":
+            raise "Can't run with --parent_bitcoin without specifying --parent_binpath"
```
instagibbs: `Unexpected exception caught during testing: TypeError('exceptions must derive from BaseException',)`

Can we handle this?

instagibbs:
let's set `printtoconsole=0` for both bitcoind/elementsd, otherwise 0.17+ nodes make a ton of noise on the test.

instagibbs:
Let's also fix up our handling of `validateaddress` ala https://github.com/ElementsProject/elements/pull/416

qa/rpc-tests/feature_fedpeg.py
jtimon: or just get rid of the function as I propose above

qa/rpc-tests/feature_fedpeg.py
instagibbs: It's a bit too noisy imo.

qa/rpc-tests/feature_fedpeg.py
instagibbs: Just put the quote in `Exception()`

instagibbs:
Here's some suggested fixes: https://github.com/instagibbs/elements/commit/af1989136ae8b6f56f5474624a961f2e856b9fa9

instagibbs:
this too: https://github.com/instagibbs/elements/commit/d084d367e0783377fde0f4778897c4b9cab7bfde

jtimon:
Can we squash "qa: Fix bug in feature_fedpeg test script"? it isn't really fixing anything, it works the same before and after the change.
Also, more suggestions https://github.com/jtimon/elements/tree/pr-415
Feel free to take some but not others.

jtimon:
Ok, either way I had to call it directly from the test to get the 'Given claim_script is not hex' exception. Updated https://github.com/jtimon/elements/tree/pr-415 with that.

qa/rpc-tests/feature_fedpeg.py
stevenroose: I also use it to restart the node lateron.

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -229,11 +264,24 @@ def run_test(self):
         print("Test pegout")
         self.test_pegout(get_new_unconfidential_address(parent), sidechain)
 
-        print("Test pegout P2SH")
-        parent_chain_addr = get_new_unconfidential_address(parent)
-        parent_pubkey = parent.validateaddress(parent_chain_addr)["pubkey"]
-        parent_chain_p2sh_addr = parent.createmultisig(1, [parent_pubkey])["address"]
-        self.test_pegout(parent_chain_p2sh_addr, sidechain)
+        # print("Test pegout P2SH")
```
instagibbs: why?

jtimon: read the commits, perhaps better read the commit tittles in the branch without the squashes: https://github.com/jtimon/elements/tree/pr-415

instagibbs: please don't disable swaths of tests when it's very possible to retain them

jtimon:
Updated https://github.com/jtimon/elements/tree/pr-415 with a squashed version that works with 0.17.0

jtimon:
Tested ACK 498877e8714c17982b81c4fd0277ca3061f366ae

jtimon:
oh, and I guess you can add an extra commit getting rid of pegging.py

stevenroose:
Done :)

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -47,42 +48,74 @@ def __init__(self):
...
 
         # Sidechain args
+        self.fedpeg_script = "512103dff4923d778550cc13ce0d887d737553b4b58f4e8e886507fc39f5e447b2186451ae"
         parent_chain_signblockscript = '51'
         for n in range(2):
             rpc_u, rpc_p = rpc_auth_pair(n)
-            self.extra_args.append([
-                # '-printtoconsole',
+            args = [
+                "-printtoconsole",
```
instagibbs: nit: can we set this to 0?

stevenroose: I added a commit to do this. If you want me to squash it with the others, let me know..

instagibbs:
@stevenroose Sorry I mean explicitly set to 0. Future Core version are 1 by default.

qa/rpc-tests/feature_fedpeg.py
instagibbs: You commented out, not put to 0. Bitcoin Core 0.17 defaults to `1`.

instagibbs:
tACK https://github.com/ElementsProject/elements/pull/415/commits/9545ca3e48ae51ea0df18d88ce062d6e9b811479

instagibbs:
I'll follow-up with cookie auth check

PR #418, opened 2018-09-25
qa: Use bitcoind as parent in Travis tests

stevenroose:


.travis.yml
```diff
@@ -30,6 +30,7 @@ env:
 # Win64
     - HOST=x86_64-w64-mingw32 DPKG_ADD_ARCH="i386" DEP_OPTS="NO_QT=1" PACKAGES="python3 nsis g++-mingw-w64-x86-64 wine1.6 bc" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-reduce-exports"
 # bitcoind
+    - HOST=i686-pc-linux-gnu PACKAGES="g++-multilib bc python3-zmq" DEP_OPTS="NO_QT=1 NO_UPNP=1 DEBUG=1" RUN_TESTS=true RUN_FEDPEG_BITCOIND=true GOAL="install" BITCOIN_CONFIG="--enable-zmq --enable-glibc-back-compat --enable-reduce-exports CPPFLAGS=-DDEBUG_LOCKORDER"
```
instagibbs: We're running into the exact some OOB issue here with the build below...

instagibbs: if you want to add a new test that won't fail, take away `DEBUG=1`. This is what enforces the stricter debugging mode that is getting tripped up.

instagibbs: or we merge #419 first then rebase

instagibbs:
Needs rebase now that we've fixed `DEBUG=1` builds.

.travis.yml
```diff
@@ -70,6 +70,11 @@ script:
     - export LD_LIBRARY_PATH=$TRAVIS_BUILD_DIR/depends/$HOST/lib
     - if [ "$RUN_TESTS" = "true" ]; then make $MAKEJOBS check VERBOSE=1; fi
     - if [ "$RUN_TESTS" = "true" ]; then qa/pull-tester/rpc-tests.py --coverage; fi
+    - if [ "$RUN_FEDPEG_BITCOIND" = "true" ]; then BITCOIND_VERSION=0.16.3; fi
```
instagibbs: any particular reason this can't be one big block inside one conditional?

stevenroose: Just readability I guess. You prefer it alltogether? I know it's a bit stupid to take 5 lines for one extra thing, but those files are already so unreadable, didn't want to contribute more to that.

I could add all the install stuff in one line and then the actual test in a second. Will be one long line though :D

stevenroose: Can't set the variables in the same line as doing 3 commands they don't ripple through. So now have var line, install line and test line.

jtimon:
Awesome.
Not on my computer, but it seems there's a 0.16.3 in the second line that
could be replace with the version var too.

stevenroose:
@jtimon Whoops, that is very much true. Fixed that.

instagibbs:
bitcoind test is taking too long for some reason:

```
Waiting to ensure block is being rejected by sidechain2
JSONRPC error: 'getblockcount' RPC took longer than 30.000000 seconds. Consider using larger timeout for calls that take longer to return.
```

instagibbs:
restarting job to see if this repeats.

stevenroose:
I've seen that error on my machine too. Second attempt usually works.

instagibbs:
ACK, moar testing

PR #421, opened 2018-09-26
QA: feature_fedpeg.py: Remove re-implementation of sync_all

jtimon:
...which is redundant and inferior to the one in the test framework.
We needed to re-implement it to avoid pegging.py depending on the freamework, but we don't need it anymore.

### DIFF
--- a/qa/rpc-tests/feature_fedpeg.py
+++ b/qa/rpc-tests/feature_fedpeg.py
@@ -16,24 +16,6 @@ from test_framework.util import (
     stop_node,
 )
 
-# Sync mempool, make a block, sync blocks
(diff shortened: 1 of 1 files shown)


stevenroose:
Oh I wasn't aware that there was a `self.sync_all()` that is aware of `self.is_network_split`. I broke my elements build yesterday, trying to fix that now.

utACK

instagibbs:
This shouldn't work, as there are two networks here?

instagibbs:
ACK https://github.com/ElementsProject/elements/pull/421/commits/9b030c10a734a4c275bbd22badd1f11a199ab55f

PR #427, opened 2018-10-10
[0.17] Don't edit Chainparams after initialization

jtimon:
This is a backport of https://github.com/bitcoin/bitcoin/pull/13311 since it's already merged, merging this in elements-0.17 should at least make the next rebase slightly easier. And since we're heavily touching chainparams in elements, I think it's a good thing.

PR #432, opened 2018-10-12
[0.17] QA: Use resgtest2 chain instead of regtest for rpc tests

jtimon:
Backport of https://github.com/bitcoin/bitcoin/pull/8994

The tests seem to pass with:

```
python3 ./test/functional/test_runner.py -j4 --extended
```

Let's please try to keep all general review things on https://github.com/bitcoin/bitcoin/pull/8994 and elements-specific things here.

Dependencies:

- [x] [0.17] Don't edit Chainparams after initialization #427
- [x] [0.17] Test: Fix example_test.py #434


jtimon:
Now the -extended tests seem to pass

instagibbs:
The motivation for this is to test the testchains feature?

jtimon:
It's to be able to create custom testchains like we do in elements-0.14, it also adapts the tests to run with -chain=regtest2, which implies a different genesis block which doesn't satisfy pow.
It also makes customizable some chainparams we currently customize in elements-0.14.

stevenroose:
Do you want to merge this right away, or wait for final feedback of upstream? LGTM at least!

jtimon:
Actually that's a good question, while feedback is coming fast in bitcoin core, it seems sensible to keep them as similar as possible and wait here. On the other hand, the bitcoin PR could take too long to get merged or not even get merged, so we don't want to wait forever. So, not sure.

instagibbs:
Perhaps we should just set a deadline and merge by then? This is kind of blocking follow-up PRs.

stevenroose:
ACK if we want to merge it now!

PR #433, opened 2018-10-14
[0.17] Signet implementation

jtimon:
A rebase of https://github.com/kallewoof/bitcoin/pull/5 on top of https://github.com/ElementsProject/elements/pull/432 taking the opportunity to simplify its changes around chainparams code and adding a test that also acts as a script to generate a private key in wallet importable format and its corresponding p2pk blockscript, which should be useful for people wanting to generate their own signets.
This represent a hardfork from that implementation because the genesis block now includes the name of the chain too, like all other custom chains.

Dependencies:

- [ ] [0.17] QA: Use resgtest2 chain instead of regtest for rpc tests #432


contrib/signet/issuer/issuer.sh
```diff
@@ -0,0 +1,75 @@
...
+if [ $conns -lt 1 ]; then
+    echo "warning: node is not connected to any other node"
+fi
+
+log "node OK with $conns connection(s)"
+log "mining in random intervals between $min_time .. $max_time seconds"
+log "hit ^C to stop"
+
+while true; do
+    let rv=$RANDOM%$randinterval
```
instagibbs: this needs `+$min_time`

jtimon:
You're going to need at least part of it here too, assuming https://github.com/ElementsProject/elements/pull/414 is merged and forward ported. putting it all facilitates testing since you wouldn't need a separated binary to test the fedpeg to a parent signet anymore.
But your call.

instagibbs:
Not going to delete the branch to be clear. I made a copy to the repo: https://github.com/ElementsProject/elements/tree/e17-signet 

IIRC I was getting conflicts as the base branch was moving along with features, and for now we just wanted a 0.17 Core branch to work on.

PR #436, opened 2018-10-17
[0.17] Chainparams: Introduce consensus.genesis_style for compatibility

jtimon:
Introduces consensus.genesis_style for compatibility and sanity tests.
Style "default_style" is what is meant to be used and extended, since it opens the door to commit to any chainparams we would like to commit.

Style "regtest2_style" is for compatibility with https://github.com/bitcoin/bitcoin/pull/8994

I'm trying to make sure that https://github.com/ElementsProject/elements/pull/433 can be compatible with https://github.com/kallewoof/bitcoin/pull/5 using "signet_old" in #433, but I'm not getting the genesis block hash that I expect, this is driving me kind of crazy, see https://github.com/kallewoof/bitcoin/pull/6 for additional context.

In the PR for block signing I plan to create another style for elements-0.14 compatibility.


luke-jr:
Why not just provide the raw coinbase instead of a predefined "style", and default it to whatever commitments you want?

jtimon:
Well, ideally the styles would unify with time (at least some of them, for example, liquidV1 has its own style not included here) and this seemed simpler. Feel free to create a competing PR.

By the way, since liquid is now public, liquidv1's style could be implemented and tested here. Or perhaps I can do that on a separated PR on top of this one.

instagibbs:
@luke-jr raw coinbase such as what exactly? Just trying to understand the suggestion

instagibbs:
just noting we can add this test post-merge: https://github.com/ElementsProject/elements/commit/f00e979306cad33fe74746005141746b4d3b3b6b#diff-912bf7afc60ccdf9de453ce3a8e0ffceR29 here

instagibbs:
Closing for now, we're supporting this piecemeal.

jtimon:
I don't follow. What piece of meal?
Why is it closed?

instagibbs:
I meant that Carl has taken half of this PR(note #467) and introduced two genesis styles already.

PR #442, opened 2018-10-22
[0.17] MODIFIED: blockheaders include block height, nodes validate it

jtimon:
Competing version of https://github.com/ElementsProject/elements/pull/431

Waiting for @instagibbs 's review before squashing.

Includes https://github.com/ElementsProject/elements/pull/436 which allows to add a small test for this as suggested in https://github.com/ElementsProject/elements/pull/431#discussion_r226708625


jtimon:
It isn't fully redundant since this version includes an extra test: https://github.com/ElementsProject/elements/pull/442/commits/f00e979306cad33fe74746005141746b4d3b3b6b#diff-912bf7afc60ccdf9de453ce3a8e0ffceR29

but perhaps https://github.com/ElementsProject/elements/pull/436 is not wanted for some reason.

instagibbs:
#436 is still in flight, good catch on test. 

I was trying to avoid stacking PRs when not necessary and missed it, mentioned on the other PR.

PR #461, opened 2018-11-08
Add Asset parameter to MutateTxAddOutData

greenaddress:
from: https://github.com/Blockstream/liquid/pull/2

### DIFF
--- a/src/bitcoin-tx.cpp
+++ b/src/bitcoin-tx.cpp
@@ -395,27 +395,42 @@ static void MutateTxAddOutMultiSig(CMutableTransaction& tx, const std::string& s
 static void MutateTxAddOutData(CMutableTransaction& tx, const std::string& strInput)
 {
     CAmount value = 0;
+ 
+    // separate [VALUE:]DATA[:ASSET] in string
+    std::vector<std::string> vStrInputParts;
+    boost::split(vStrInputParts, strInput, boost::is_any_of(":"));
 
-    // separate [VALUE:]DATA in string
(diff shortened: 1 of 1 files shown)


src/bitcoin-tx.cpp
```diff
@@ -395,27 +395,42 @@ static void MutateTxAddOutMultiSig(CMutableTransaction& tx, const std::string& s
...
+    // separate [VALUE:]DATA[:ASSET] in string
+    std::vector<std::string> vStrInputParts;
+    boost::split(vStrInputParts, strInput, boost::is_any_of(":"));
 
-    // separate [VALUE:]DATA in string
-    size_t pos = strInput.find(':');
-
-    if (pos==0)
+    // Check that there are enough parameters
+    if (vStrInputParts.size()<1)
```
stevenroose: @dgyoshi is it even possible that that is 0? I don't know about boost specifically, but most string splitters do `"".split(":") == [""]` so it would always have 1 element, even for an empty string.

instagibbs: @stevenroose can we stick to the Liquid PR for now?

instagibbs:
closing until we merge in Liquid repo

stevenroose:
I don't think a unit test exists for that method in the first place, but I feel like it makes sense to have one.

src/bitcoin-tx.cpp
stevenroose: Oh, yeah sure.

Issue #465, opened 2018-11-14
Offline raw claim peg-in transaction singing errors


grubles:
### Describe the issue
I'm trying to claim L-BTC on Liquid using one online `liquidd` and one offline. When attempting to sign a raw claim peg-in transaction generated with `createrawpegin`, the signing errors with `Peg-in input has invalid proof.` My setup is as follows: 

`bitcoind` and `liquidd` running online on one computer.

`bitcoind` and `liquidd` running offline on a separate computer. 

### Can you reliably reproduce the issue?
Yes.

#### If so, please list the steps to reproduce below:
1. Gather claim_script and BTC peg-in raw transaction 
2. Generate `txoutproof` with online `bitcoind`
3. Use `createrawpegin` with above information
4. Move raw transaction hex to offline `liquidd`
5. Attempt to sign the raw tx with `signrawtransaction`
6. Signing errors

### Expected behaviour
Signing should complete successfully 

### Actual behaviour
Signing errors with "Peg-in input has invalid proof."

### What version of bitcoin-core are you using?
Bitcoin Core 0.17.0 official binary

instagibbs:
are you running `validatepegin=1` on the offline machine?

grubles:
Running with nothing specified defaults to `1` right? I tried with `0` as well and it didn't work.

instagibbs:
@grubles take a look at https://github.com/ElementsProject/elements/pull/466 to see if I am covering the logic you're hitting?

I can't replicate.

instagibbs:
Ok, I think I figured it out. If you don't let the peg-in mature it won't allow you to sign it currently.

I suppose it should allow signing, especially on an offline machine!

grubles:
Turns out I was typo'ing `validatepegin=0` with validatepegin**s**=0. 

When running with `validatepegin=0`, the offline signing completed successfully and returned a `"completed": true"`.

instagibbs:
Great, moving ahead with my PR that should make this issue smoother for offline signing even if fat-fingered.

PR #466, opened 2018-11-14
Add basic raw pegin creation test and fixup return name in help

instagibbs:
shores up concerns in https://github.com/ElementsProject/elements/issues/465 and fixes a return variable help text. Adds tests.


qa/rpc-tests/feature_fedpeg.py
```diff
@@ -240,25 +240,50 @@ def run_test(self):
...
+        # where raw creation is done on another node
         for i in range(n_claims):
             addrs = sidechain.getpeginaddress()
             txid = parent.sendtoaddress(addrs["mainchain_address"], 1)
-            parent.generate(12)
+            parent.generate(1)
             proof = parent.gettxoutproof([txid])
             raw = parent.getrawtransaction(txid)
-            pegtxs += [sidechain.claimpegin(raw, proof)]
+            if i % 2 == 0:
```
stevenroose: modulo math! much crypto!

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -240,25 +240,50 @@ def run_test(self):
...
+                parent.generate(11)
+                pegtxs += [sidechain.claimpegin(raw, proof)]
+            else:
+                # The raw API doesn't check for the additional 2 confirmation buffer
+                # So we only get 10 confirms then send off. Miners will add to block anyways.
+
+                # Don't mature whole way yet to test signing immature peg-in input
+                parent.generate(8)
+                # Wallet in sidechain2 gets funds instead of first
```
stevenroose: > # Wallet in sidechain2 gets funds instead of first

I don't understand that sentence.

stevenroose: Ah you mean "instead of wallet in sidechain1"?

instagibbs: the `gettransaction` call later would fail if we don't select the right node since it's not in their wallet.

The "peg-in input" is not "in the wallet" at all. This is something we should probably improve at some point, but might require deeper changes.

instagibbs: yes exactly, sorry

qa/rpc-tests/feature_fedpeg.py
```diff
@@ -240,25 +240,50 @@ def run_test(self):
...
+                raw_pegin = sidechain2.createrawpegin(raw, proof, addrs["claim_script"])["hex"]
+                # First node should also be able to make a valid transaction with or without 3rd arg
+                # since this wallet originated the claim_script itself
+                sidechain.createrawpegin(raw, proof, addrs["claim_script"])
```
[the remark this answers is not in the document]
instagibbs: un-caught errors bubble up and fail the test

src/rpc/rawtransaction.cpp
```diff
@@ -1421,10 +1425,14 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
         if (!txin.m_is_pegin && (coins == NULL || !coins->IsAvailable(txin.prevout.n))) {
             TxInErrorToJSON(txin, vErrors, "Input not found or already spent");
             continue;
-        } else if (txin.m_is_pegin && (txConst.wit.vtxinwit.size() <= i || !IsValidPeginWitness(txConst.wit.vtxinwit[i].m_pegin_witness, txin.prevout))) {
+        } else if (txin.m_is_pegin && (txConst.wit.vtxinwit.size() <= i || !IsValidPeginWitness(txConst.wit.vtxinwit[i].m_pegin_witness, txin.prevout, false))) {
             TxInErrorToJSON(txin, vErrors, "Peg-in input has invalid proof.");
             continue;
         }
+        // Report warning about immature peg-in though
+        if(!IsValidPeginWitness(txConst.wit.vtxinwit[i].m_pegin_witness, txin.prevout)) {
```
stevenroose: I would add the `, true` here to make the difference more clear with the earlier call, but I can add that in the rebase, it's not critical.

instagibbs: you can probably just remove the optional part of the argument in the rebase

instagibbs:
Forgot to check for peg-in-ness of the input for the warning. Fixed and addressed comment nit

stevenroose:
reACK or what do you say then? :) I checked the diff.

instagibbs:
restarted one build due to (spurious?) zapwallettx failure

PR #458, opened 2018-11-07
[0.17] Add pegin validation

stevenroose:

PR #467, opened 2018-11-15
[0.17] Add initialfreecoins option, default of 0

dongcarl:
This PR adds a `-initialfreecoins` debug option, and a `-con_genesis_style` to go along with it.

This allows us to correctly set the amount of coins created in the genesis block, either in the `elements` style or `bitcoin` style.

Tests have been modified to test that the correct amount of coins are created.


stevenroose:
Relevant section: https://github.com/ElementsProject/elements/pull/458/commits/7fc252d3ab942b3edd6e2a9b65ef2d77a197c192#diff-64cbe1ad5465e13bc59ee8bb6f3de2e7R516

instagibbs:
@dongcarl this being worked on?

dongcarl:
@instagibbs I remember this one needs to be rebased on some other changes (maybe issuance consensus/policy?) before it can be fleshed out.

src/chainparams.cpp
```diff
@@ -435,6 +435,7 @@ class CCustomParams : public CRegTestParams {
         // All non-zero coinbase outputs must go to this scriptPubKey
         std::vector<unsigned char> man_bytes = ParseHex(gArgs.GetArg("-con_mandatorycoinbase", ""));
         consensus.mandatory_coinbase_destination = CScript(man_bytes.begin(), man_bytes.end()); // Blank script allows any coinbase destination
+        initialFreeCoins = gArgs.GetArg("-initialfreecoins", 0);
```
instagibbs: use `DEFAULT_INITIAL_FREE_COINS`

dongcarl: Would cause circular dependencies... Just used 0

instagibbs: well that constant doesn't appear to be used then

dongcarl: @instagibbs It would seem from surrounding code that the convention is to just specify instead of making a new constant. So I've removed it.

dongcarl:
Seems like `rpc_getblockstats.py` is still not happy, it does pass locally on my machine, and it seems like blocks just aren't connecting for some reason.

instagibbs:
So this is probably because this PR breaks the ability to make and run "vanilla regtest" blockchain, breaking the `bitcoin_functional` run that uses pre-generated data. 

See: https://github.com/ElementsProject/elements/pull/436

I think it makes sense to do a minimal version of that PR, supporting Bitcoin-compatible genesis block generation, as well as elements/liquid-compatible.

instagibbs:
Ok, we "broke" bitcoin regtest block previously to adding the new tests, so I'd suggest you take this, generate the new json block data in the `bitcoin_functional` section of tests: https://github.com/instagibbs/elements/commit/04472668fdf7aa07f463009a8dfaad1a16ae29d0

instagibbs:
otherwise this is looking good!

instagibbs:
unmarked as WIP

instagibbs:
nit: Could you just squash the two commits to make all commits past the test?

test/functional/feature_connect_coinbase.py
```diff
@@ -31,10 +31,17 @@ def run_test(self):
...
-        assert_raises_rpc_error(-5, "The genesis block coinbase is not considered an ordinary transaction and cannot be retrieved", self.nodes[0].getrawtransaction, coinbase_tx)
-        self.nodes[1].getrawtransaction(coinbase_tx)
+
+        # Coinbase transaction is provably unspendable (OP_RETURN), so even AddCoin won't add it
+        assert_raises_rpc_error(-5, "No such mempool transaction. Use -txindex to enable blockchain transaction queries. Use gettransaction for wallet transactions.", self.nodes[0].getrawtransaction, coinbase_tx)
+        assert_raises_rpc_error(-5, "No such mempool transaction. Use -txindex to enable blockchain transaction queries. Use gettransaction for wallet transactions.", self.nodes[1].getrawtransaction, coinbase_tx)
+
+        # Issuance transaction is an OP_TRUE, so will be available to second node
+        assert_raises_rpc_error(-5, "No such mempool transaction. Use -txindex to enable blockchain transaction queries. Use gettransaction for wallet transactions.", self.nodes[0].getrawtransaction, issuance_tx)
+        self.nodes[1].getrawtransaction(issuance_tx)
```
instagibbs: Please add a test showing that `listunspent` has a single oiutput worth the amount you set in the conf line: `5000000000`

instagibbs:
light tACK, just add those quick test conditions to make sure the output looks like something you expect

test/functional/feature_connect_coinbase.py
dongcarl: This would require `anyonecanspendaremine=1` correct? Should I add that flag just for this test or shall we make this the default at least for the non-bitcoin functional tests?

instagibbs: Oh right, I'd set it just for this test.

dongcarl:
@instagibbs Tests added, let me know if there are other things!

instagibbs:
re-ACK https://github.com/ElementsProject/elements/pull/467/commits/9b0632b3d7713edda2085bf865a06b5e3530d271

test/bitcoin_functional/functional/data/rpc_getblockstats.json
```diff
@@ -1,117 +1,117 @@
 {
   "blocks": [
-    "010000000000000000000000000000000000000000000000000000000000000000000000bdb84e02590a29888f409285dfe25361bbd5560efa5a2dd6996c53488bf37e8edae5494dffff7f20020000000101000000010000000000000000000000000000000000000000000000000000000000000000ffffffff1004ffff001d0104087265677465737432ffffffff0100f2052a01000000015100000000",
```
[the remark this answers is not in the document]
instagibbs: good catch, yes with these commits they should match again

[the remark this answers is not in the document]
instagibbs: Well technically no, it's "regtest2" network, which gets mixed into the block definition. The network `main` still should work though. We don't really have tests for that per se, aside from assertions in the chainparams.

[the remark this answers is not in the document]
dongcarl: Not sure what the conclusion is here, does this need to be addressed?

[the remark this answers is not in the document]
instagibbs: I don't think this is a blocker unless this mutation is incorrect behavior based on the rpc_blockstats test...

instagibbs:
needs rebase

instagibbs:
and please add a description :)

test/bitcoin_functional/functional/data/rpc_getblockstats.json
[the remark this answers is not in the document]
stevenroose: I also don't get it. What exactly changed to the regtest genesis block?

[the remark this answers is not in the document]
stevenroose: Can't we skip `CommitToArguments` if the genesis-style is bitcoin?

dongcarl:
@instagibbs Description added.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/467/commits/95432aad58f366ffa349dc72b0148af27c609bf7

stevenroose:
post-merge ACK 95432aa :) was trying something out regarding the bitcoin_functional tests

PR #452, opened 2018-10-30
[WIP, 0.17] Signed blocks

instagibbs:
Need to do history cleanup, and more sanity checks for rpc calls that use signed blocks.

PR #470, opened 2018-11-18
[0.17] Block v4 mandatory

dongcarl:
By setting the BIPXX consensus fields to `0` we allow them to be enforced from genesis block by default.

This means blocks will be required to be at least version 4 by default.

Also adds tests.

Todo for later: `BIP34_IMPLIES_BIP30_LIMIT` right now means all chains will slow down validation at height `1983702`. We should prod upstream to do something smarter, or allow this to be set as an argument.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -417,10 +417,10 @@ class CCustomParams : public CRegTestParams {
 
         consensus.nSubsidyHalvingInterval = args.GetArg("-con_nsubsidyhalvinginterval", consensus.nSubsidyHalvingInterval);
         consensus.BIP16Exception = uint256S(args.GetArg("-con_bip16exception", "0x0"));
-        consensus.BIP34Height = args.GetArg("-con_bip34height", consensus.BIP34Height);
+        consensus.BIP34Height = args.GetArg("-con_bip34height", 0);
         consensus.BIP34Hash = uint256S(args.GetArg("-con_bip34hash", "0x0"));
-        consensus.BIP65Height = args.GetArg("-con_bip65height", consensus.BIP65Height);
-        consensus.BIP66Height = args.GetArg("-con_bip66height", consensus.BIP66Height);
+        consensus.BIP65Height = args.GetArg("-con_bip65height", 0);
+        consensus.BIP66Height = args.GetArg("-con_bip66height", 0);
         consensus.powLimit = uint256S(args.GetArg("-con_powlimit", "7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff"));
         consensus.nPowTargetTimespan = args.GetArg("-con_npowtargettimespan", consensus.nPowTargetTimespan);
         consensus.nPowTargetSpacing = args.GetArg("-con_npowtargetspacing", consensus.nPowTargetSpacing);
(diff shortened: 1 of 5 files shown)


dongcarl:
Made the relevant changes and rebased over https://github.com/ElementsProject/elements/pull/452, having some trouble making tests work. Seems like making `g_signed_blocks` true might fix it, but breaks other tests.

instagibbs:
note that this new method of activating new blocks is an attempt to use the starting block of various rules at block 0. This will need reasonable review.

instagibbs:
For `bitcoin_functional` tests the original regtest activation heights will likely have to be put in the `util.py` function that creates the conf file.

instagibbs:
I believe this will get you to passing tests, but we'll need a test of our own setting them to zero:

```
diff --git a/test/bitcoin_functional/functional/test_framework/util.py b/test/bitcoin_functional/functional/test_framework/util.py
index 9b2129c..145f8ce 100644
--- a/test/bitcoin_functional/functional/test_framework/util.py
+++ b/test/bitcoin_functional/functional/test_framework/util.py
@@ -309,6 +309,9 @@ def initialize_datadir(dirname, n, chain):
         f.write("con_connect_coinbase=0\n")
         f.write("anyonecanspendaremine=0\n")
         f.write("con_blockheightinheader=0\n")
+        f.write("con_bip34height=100000000\n")
+        f.write("con_bip65height=1351\n")
+        f.write("con_bip66height=1251\n")
         os.makedirs(os.path.join(datadir, 'stderr'), exist_ok=True)
         os.makedirs(os.path.join(datadir, 'stdout'), exist_ok=True)
     return datadir
diff --git a/test/functional/test_framework/util.py b/test/functional/test_framework/util.py
index 7630e32..d250fd1 100644
--- a/test/functional/test_framework/util.py
+++ b/test/functional/test_framework/util.py
@@ -308,6 +308,9 @@ def initialize_datadir(dirname, n, chain):
         f.write("con_blocksubsidy=5000000000\n")
         f.write("con_connect_coinbase=0\n")
         f.write("anyonecanspendaremine=0\n")
+        f.write("con_bip34height=100000000\n")
+        f.write("con_bip65height=1351\n")
+        f.write("con_bip66height=1251\n")
         os.makedirs(os.path.join(datadir, 'stderr'), exist_ok=True)
         os.makedirs(os.path.join(datadir, 'stdout'), exist_ok=True)
     return datadir
```

dongcarl:
@instagibbs Added tests, lmk if you want something more extensive.

instagibbs:
unmarked WIP

instagibbs:
oh baby that test :1st_place_medal: I gotta learn that mininode stuff too.

Not even a nit per se but why the CLTV test? To make sure that blocks are being made of v4?

instagibbs:
ACK https://github.com/ElementsProject/elements/pull/470/commits/e04f6c7018b6a0d22d1302f32c524620b5c6e83d

dongcarl:
> oh baby that test 1st_place_medal I gotta learn that mininode stuff too.
> 
> Not even a nit per se but why the CLTV test? To make sure that blocks are being made of v4?

Yup exactly! :smile:

instagibbs:
I added a description to the OP

dongcarl:
> I added a description to the OP

Thank you sir!

PR #477, opened 2018-12-08
remove unnecessary confValue reassignment

dgpv:
A minor nitpick, but removing redundant code improves readability.

I was reading the code for FillBlinds() and noticed that there is unnecessary assignment of confValue, because the value is not used after that - its scope is within the loop, and it is set anew at the start of the loop.

"//Wipe out confidential info from issuance" comment is applicable to vchRangeproof.clear(), but not to confValue reassignment - becasue this variable is just a reference.

NOTE: I did not compile or tested the code.  I just wanted to report this minor thing.  I am not sure that creating a PR is a correct channel to notify about this sort of nitpicks, maybe I should have created an issue instead of PR ?


instagibbs:
Thanks for the report. It indeed looks superfluous.

instagibbs:
Actually no, `confValue` is a reference to a transaction output, so the assignment is over-writing the transaction output to a non-confidential value if it was able to decode it. This is the intended behavior, and the tx is later blinded.

instagibbs:
The fact that this doesn't fail a test means we need to beef up our tests a bit I suppose. All this is going into a refactor/rewrite for elements-0.17, so I shall defer it to then...

instagibbs:
Indeed it would be more readable. I'm not sure what memory concerns we have here, it's 8 bytes or something.

I'll ping you when it comes time for me to write CA stuff for the rebase ;)

Issue #485, opened 2018-12-22
src/blind.cpp does not have standard license header, like other .cpp files


dgpv:
https://github.com/ElementsProject/elements/blob/5467be3b16f014b371ecf6bced73be0254755cfa/src/blind.cpp#L1

Other files have copyright notices and reference to MIT license.

Is the code in blind.cpp also covered under MIT License ?

instagibbs:
It should. Feel free to add as PR!

dgpv:
PR merged, closing.

PR #486, opened 2018-12-22
add standard license header to src/blind.cpp

dgpv:
As discussed in #485, src/blind.cpp should have a standard license header.


instagibbs:
ACK

PR #492, opened 2019-01-03
check amount against MAX_MONEY with unsigned comparison

dgpv:
Because explicit conversion of arbitrarily large unsigned values
to signed values of the same bit width is implementation defined
until c++20, better check against MAX_MONEY with unsigned comparison.


dgpv:
If the MoneyRange check is to be removed, as mentioned in https://github.com/Blockstream/liquid/pull/13#issuecomment-451252624, then just decline this PR.

instagibbs:
this is the more conservative route for now, so utACK

Issue #506, opened 2019-02-14
claimpegin RPC Reference says 10 confirmations needed - is it not 102?


wintercooled:
doc/elements-api.md says:

"The mainchain deposit transaction must have at least 10 confirmations before the claim is valid"

Is it not 102 confirmations? or say "the transaction will not be relayed unless it is at least 102 blocks deep" as elsewhere.

instagibbs:
https://github.com/ElementsProject/elements/blob/elements-0.14.1/src/chainparams.cpp#L131 defines it for each network. Elements uses 8, I believe, +2 for wallet to make sure it doesn't go into a race condition with other nodes. Liquid is 100 + 2.

wintercooled:
Ah I see . Thanks, will close and ditch the associated PR. Cheers.

PR #507, opened 2019-02-14
Chang claimpegin depth to 102 in api docs

wintercooled:
Previously stated deposit tx depth must be 10 but I think it's 102

ref issue #506


instagibbs:
I replied in the issue, it depends on the exact chainparams.

wintercooled:
Closing based on instagibbs comments.

PR #510, opened 2019-02-18
Add partial blinding test to qa/rpc-tests

dgpv:
This PR adds fixes for witness serialization/deseralization in qa/rpc-tests/test_framework/mininode.py
and extends multisig tests in qa/rpc-tests/confidential_transactions.py with partial blinding tests

### DIFF
--- a/qa/rpc-tests/confidential_transactions.py
+++ b/qa/rpc-tests/confidential_transactions.py
@@ -3,8 +3,22 @@
 # Distributed under the MIT/X11 software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+import io
+
+from decimal import Decimal
+
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import *
+from test_framework.mininode import (
+    COIN, CTransaction, CTxOut, CTxOutAsset, CTxOutValue,
+    CTxInWitness, CTxOutWitness,
+)
+from test_framework.util import (
+    start_nodes, connect_nodes_bi, assert_equal,
+    hex_str_to_bytes, bytes_to_hex_str,
+    BITCOIN_ASSET_OUT
+)
+from test_framework.authproxy import JSONRPCException
+
 
 class CTTest (BitcoinTestFramework):
 
@@ -307,7 +321,8 @@ class CTTest (BitcoinTestFramework):
         assert_equal(multi_asset_amount['bitcoin'], value1 + value3 )
         assert_equal(multi_asset_amount[test_asset], Decimal('0.00000003'))
 
-        # Check blinded multisig functionality
+        # Check blinded multisig functionality and partial blinding functionality
+
         # Get two pubkeys
         blinded_addr = self.nodes[0].getnewaddress()
         pubkey = self.nodes[0].validateaddress(blinded_addr)["pubkey"]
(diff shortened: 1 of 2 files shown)


instagibbs:
great contribution, thanks, will review.

qa/rpc-tests/confidential_transactions.py
```diff
@@ -325,16 +329,122 @@ def run_test(self):
...
+        assert_equal(node2_balance[issued3['asset']], Decimal(1))
+        # Send asset to blinded multisig address and check that they were received
+        self.nodes[2].sendtoaddress(blinded_multisig_addr, 1, "", "", False, issued3['asset'])
+        self.sync_all()
+        unspent_asset = self.nodes[0].listunspent(0, 0, [unconfidential_addr], True, issued3['asset'])
+        assert_equal(len(unspent_asset), 1)
+        assert(issued3['asset'] not in self.nodes[2].getbalance())
 
-        self.nodes[0].generate(1)
+        blinded_addr = self.nodes[0].getnewaddress()
```
instagibbs: A bit of a description of the setup would help here, especially with the inter-leaved testing happening

qa/rpc-tests/confidential_transactions.py
```diff
@@ -325,16 +329,122 @@ def run_test(self):
...
+        blindtx = self.nodes[0].blindrawtransaction(
+            bytes_to_hex_str(btx.serialize()), True, [
+                unspent2[0]['assetcommitment'],
+                unspent[0]['assetcommitment'],
+                unspent_asset[0]['assetcommitment']
+            ])
+
+        stx2 = self.nodes[1].signrawtransaction(blindtx)
+        stx = self.nodes[0].signrawtransaction(stx2['hex'])
+        self.nodes[2].sendrawtransaction(stx['hex'])
```
instagibbs: let's generate a block and make sure it has confirmations, to rule out mempool weirdness.

qa/rpc-tests/confidential_transactions.py
```diff
@@ -325,16 +329,122 @@ def run_test(self):
...
+            {dst_addr: unspent[0]['asset'], dst_addr3: unspent_asset[0]['asset']})
+
+        sum_i = unspent2[0]["amount"] + unspent[0]["amount"]
+        sum_o = 0.01 + 0.10 + 0.1
+        assert_equal(int(round(sum_i*COIN)), int(round(sum_o*COIN)))
+
+        # Blind the first part of the transaction - we need to supply the
+        # assetcommmitments for all of the inputs, for the surjectionproof
+        # to be valid after we combine the transactions
+        blindtx = self.nodes[1].blindrawtransaction(
```
instagibbs: wow I'm surprised it does the right thing here, even when the transaction itself only has 1 input! I guess 2 years ago me was thinking ahead...

qa/rpc-tests/confidential_transactions.py
instagibbs: ie I didn't really get the importance of having multiple assets, why you issued an asset first, etc.

A paragraph even describing this entire section(and mark the front/end) would not hurt!

qa/rpc-tests/confidential_transactions.py
```diff
@@ -3,7 +3,10 @@
 # Distributed under the MIT/X11 software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+import io
+
 from test_framework.test_framework import BitcoinTestFramework
+from test_framework.mininode import *
```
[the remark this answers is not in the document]
dgpv: Ok. I was just copying the style used in other tests :-)

qa/rpc-tests/confidential_transactions.py
dgpv: Using multiple assets probably does not matter much, but I thought that covering more functionality makes the tests cover more, and there might happen be some corner cases.. But no specific reason why use multiple assets. I guess this doesn't hurt...

instagibbs: that's a valid answer :)

dgpv:
addressed the comments/suggestions.

instagibbs:
Looks great, could you squash the last 4 commits?

dgpv:
done

instagibbs:
looks like you snuck in one of my commits during your squash...

dgpv:
oops

dgpv:
fixed, also rebased on top of current head of elements-0.14.1

instagibbs:
force push diff is only the parts rebased onto.

utACK https://github.com/ElementsProject/elements/pull/510/commits/6c5e0755fc97950a688799602cf1c417e3455640

instagibbs:
@stevenroose can you ACK and port to 0.17?

qa/rpc-tests/confidential_transactions.py
instagibbs: You note out of band that this is *indeed* broken behavior.

Could you please add this fix in `BlindTransaction` if it seems right:

```
     // Needed to construct the proof itself. Generators must match final transaction to be valid
     std::vector<secp256k1_generator> targetAssetGenerators;
     surjectionTargets.resize(tx.vin.size()*3);
+    if (auxiliary_generators) {
+        // It's fine to over-count here. We truncate before generating proofs.
+        surjectionTargets.resize(tx.vin.size()*3 + auxiliary_generators->size())
+    }
```

instagibbs:
As noted, this test actually uncovered a small OOB issue, asked for a proposed fix in this PR to not lose track of it.

qa/rpc-tests/confidential_transactions.py
dgpv: Submitted #514 - made a little refactoring to avoid extra if(auxiliary_generators), and extra resize. Please check.

instagibbs:
restarted all builds, as they pretty much all failed seconds into building

PR #514, opened 2019-03-15
fix vector length calculation for targets in BlindTransaction

dgpv:
The size of surjectionTargets and targetAssetGenerator vectors was calculated as tx.vin.size()*3, based on the fact that for each input there might also be up to 2 issuance pseudo-inputs, but did not take into account that the number of auxiliary generators may exceed the number of inputs. This fixes the size calculations by taking into account auxiliary generators supplied beyond vin size.

### DIFF
--- a/src/blind.cpp
+++ b/src/blind.cpp
@@ -235,9 +235,6 @@ int BlindTransaction(std::vector<uint256 >& input_blinding_factors, const std::v
     assert(tx.vin.size() == input_asset_blinding_factors.size());
     assert(tx.vin.size() == input_assets.size());
     assert(tx.vin.size() == input_amounts.size());
-    if (auxiliary_generators) {
-        assert(auxiliary_generators->size() >= tx.vin.size());
-    }
 
     std::vector<unsigned char*> blindptrs;
     std::vector<const unsigned char*> assetblindptrs;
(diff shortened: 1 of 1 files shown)


src/blind.cpp
```diff
@@ -255,8 +252,14 @@ int BlindTransaction(std::vector<uint256 >& input_blinding_factors, const std::v
 
     // Needed to construct the proof itself. Generators must match final transaction to be valid
     std::vector<secp256k1_generator> targetAssetGenerators;
-    surjectionTargets.resize(tx.vin.size()*3);
-    targetAssetGenerators.resize(tx.vin.size()*3);
+
+    size_t maxTargets = tx.vin.size()*3;
```
instagibbs: Maybe note here that this is a strict upper-bound, and the size is likely less(and is shrunk later).

dgpv: added the comment

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/514/commits/42a7a34cf3a64d0990b47ab413fc291945d80bf2

dgpv:
Fixed mistype in comment "The vectors will be srunk" -> "The vectors will be shrunk"

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/514/commits/045335a4a63b7b9e76c8f5e7b93b1ca2108ee89c , spurious failure on one test, restarted.

PR #503, opened 2019-01-29
[0.17, WIP] Confidential Assets

stevenroose:
TODO:

Multi-change output script stuff
Squash history

PR #515, opened 2019-03-20
[0.17] Cleaned up CA

stevenroose:
Cleaned up version of https://github.com/ElementsProject/elements/pull/503.


instagibbs:
Seems to have previous branch +- s/elementswitness/elementsmode/, some test renaming, and debugging printing being removed.

Pushed documentation to try and fix linter.

instagibbs:
Currently having two unexpected failures. One is debug build and consistent, will try to reproduce locally. The other is simply taking too long during functional testing. I imagine this is the fee estimation test which takes quite a while even on more powerful machines.

PR #402, opened 2018-08-24
Disallow extended encoding for non-witness transactions

apoelstra:
Backport from upstream https://github.com/bitcoin/bitcoin/pull/14039

PR #525, opened 2019-03-21
[0.17] Disallow extended encoding for non-witness transactions

stevenroose:
backport of https://github.com/ElementsProject/elements/pull/402 which is a backport of upstream

### DIFF
--- a/src/primitives/transaction.h
+++ b/src/primitives/transaction.h
@@ -351,6 +351,10 @@ inline void UnserializeTransaction(TxType& tx, Stream& s) {
             const_cast<CTxWitness*>(&tx.witness)->vtxinwit.resize(tx.vin.size());
             const_cast<CTxWitness*>(&tx.witness)->vtxoutwit.resize(tx.vout.size());
             s >> tx.witness;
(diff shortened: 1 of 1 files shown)


src/primitives/transaction.h
```diff
@@ -351,6 +351,10 @@ inline void UnserializeTransaction(TxType& tx, Stream& s) {
             const_cast<CTxWitness*>(&tx.witness)->vtxinwit.resize(tx.vin.size());
             const_cast<CTxWitness*>(&tx.witness)->vtxoutwit.resize(tx.vout.size());
             s >> tx.witness;
+            if (!tx.HasWitness()) {
+                /* It's illegal to encode witnesses when all witness stacks are empty. */
+                throw std::ios_base::failure("Superfluous witness record");
+            }
```
instagibbs: I think we already check this condition in the witness serialization itself, but doesn't hurt here either.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/525/commits/e11c6f558091c8367c0d74b594187f0947fd3e7f

PR #554, opened 2019-04-01
Update issuance.cpp

roconnor-blockstream:
Typo in GenerateAssetEntropy comment.


src/issuance.cpp
```diff
@@ -9,7 +9,7 @@ void GenerateAssetEntropy(uint256& entropy, const COutPoint& prevout, const uint
     // E : entropy
     // I : prevout
     // C : contract
-    // E = H( H(I) || H(C) )
+    // E = H( H(I) || C )
```
instagibbs: Maybe it's a bit confusing, but C assumed to be a "full contract" and `contracthash` pre-hashed. Any better way to explain it?

PR #595, opened 2019-04-23
Add unit test for the PolyMod method in b(l)ech32

stevenroose:


src/test/bech32_tests.cpp
```diff
@@ -66,4 +66,22 @@ BOOST_AUTO_TEST_CASE(bip173_testvectors_invalid)
     }
 }
 
+BOOST_AUTO_TEST_CASE(bech32_polymod_sanity)
```
instagibbs: this a suitable upstream test?

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/595/commits/0de2b056d12f0aaa74a7cca1a74e320a9de03815

stevenroose:
Pushed two more commits with a fix.

src/test/bech32_tests.cpp
```diff
@@ -2,7 +2,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-#include <bech32.h>
+#include <bech32.cpp>
```
sipa: Yuck (including a .cpp file).

instagibbs: *chants* export export export

instagibbs: linter doesn't like it either :D

stevenroose: Yeah was to test the internal methods. Changed that meanwhile.

instagibbs:
ACK https://github.com/ElementsProject/elements/pull/595/commits/4a218cd0efdf300471341cdced18968f944578f2

PR #600, opened 2019-05-01
PSBT for Confidential Assets

gwillen:
This PR extends the PSBT format and RPCs to handle Confidential Assets transactions.

New fields in PSBT inputs: Unblinded value, Value blinder, Unblinded asset, Asset blinder.
New fields in PSBT outputs: Recipient blinding pubkey, Value commitment, Value blinder, Asset commitment, Asset blinder, Nonce commitment, Range proof, Surjection proof.

We preserve the existing invariant that the unsigned transaction inside a PSBT never changes during the process; all updates that need to be applied to produce the final signed transaction are accumulated in the PSBT fields listed above.

The process is as follows:
* Create a PSBT using `converttopsbt` [deprecated], `walletcreatefundedpsbt`, `createpsbt`
   * At this point, the output pubkey fields will be filled, for any outputs to confidential addresses.
   * Input fields may be filled, if the wallet was used; otherwise they will be filled in the next step.
   * Incremental creation of the unsigned transaction itself is OUTSIDE the scope of this work. The unsigned transaction must be fully populated with inputs and outputs before the PSBT is created.
   * Transactions with peg-in, issuance, or reissuance outputs are NOT supported at this time.
* If necessary, update the psbt with input data using `walletfillpsbtdata`.
   * This is like the old `walletprocesspsbt` RPC [deprecated], which tried to both fill and sign the PSBT (which is not workable in the Confidential Assets setting.)
* Once all inputs have had necessary data updated, possibly using multiple wallets if necessary, any wallet can be used to blind the transaction using `blindpsbt`.
   * This uses the input blinding data, along with the output blinding pubkeys, to compute the output blinding data.
   * Incremental blinding is not supported. All input blinding data must be available when `blindpsbt` is called.
* Then, the blinded PSBT must be signed using `walletsignpsbt`. As with updating, this can be done by multiple wallets as necessary for the inputs being signed.
* Once all signatures are present, `finalizepsbt` is used to create the final transaction in the regular transaction format, as before.
* Then `sendrawtransaction` is used, which will check to make sure that blinding was performed properly before sending.


instagibbs:
could you rebase?

src/wallet/wallet.cpp
```diff
@@ -5151,48 +5151,41 @@ void CWalletTx::GetBlindingData(const unsigned int map_index, const std::vector<
     if (asset_out) *asset_out = asset_tag;
 }
 
-CAmount CWalletTx::GetOutputValueOut(unsigned int output_index) const {
+void CWalletTx::GetNonIssuanceBlindingData(const unsigned int output_index, CPubKey* blinding_pubkey_out, CAmount* value_out, uint256* value_factor_out, CAsset* asset_out, uint256* asset_factor_out) const {
```
instagibbs: this refactoring could be its own commit

src/wallet/wallet.cpp
```diff
@@ -2940,7 +2940,7 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CTransac
                 const OutputType change_type = TransactionChangeType(coin_control.m_change_type ? *coin_control.m_change_type : m_default_change_type, vecSend);
                 // One change script per output asset.
                 size_t index = 0;
-                for (const std::pair<CAsset, CAmount>& value : mapValue) {
+                for (const std::pair<const CAsset, CAmount>& value : mapValue) {
```
instagibbs: compiler complaint squashing could be its own commit

src/rpc/rawtransaction.cpp
```diff
@@ -1270,6 +1286,123 @@ static std::string WriteHDKeypath(std::vector<uint32_t>& keypath)
...
+            psbtx.outputs[keyIndex].blinding_pubkey = CPubKey();
+            return EncodePSBT(psbtx);
+        } else {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Unable to blind transaction: Add another output to blind in order to complete the blinding.");
+        }
+    }
+
+    CMutableTransaction tx_tmp = tx;  // We don't want to mutate the transaction in the PSBT yet, just extract blinding data
+
+    // TODO(gwillen): Make this do something better than fail silently if there are any issuances, reissuances, pegins, etc.
```
instagibbs: PSBTs can't really encode these anyways yet, I think we're safe for now.

Supporting the creation of issuance/peg-in inputs in a PSBT format is the harder part.

src/rpc/rawtransaction.cpp
```diff
@@ -1270,6 +1286,123 @@ static std::string WriteHDKeypath(std::vector<uint32_t>& keypath)
...
+
+        o.value_commitment = tx_tmp.vout[i].nValue;
+        o.asset_commitment = tx_tmp.vout[i].nAsset;
+        o.nonce_commitment = tx_tmp.vout[i].nNonce;
+        o.value_blinding_factor = output_value_blinds[i];
+        o.asset_blinding_factor = output_asset_blinds[i];
+        o.range_proof = tx_tmp.witness.vtxoutwit[i].vchRangeproof;
+        o.surjection_proof = tx_tmp.witness.vtxoutwit[i].vchSurjectionproof;
+
+        o.blinding_pubkey = CPubKey();  // Once we're done blinding, remove the pubkeys to signal that it's complete
```
instagibbs: I guess this makes sense. We remove partial sigs once a full sig is complete right?

src/rpc/rawtransaction.cpp
```diff
@@ -1270,6 +1286,123 @@ static std::string WriteHDKeypath(std::vector<uint32_t>& keypath)
...
+        const CPubKey& key = output_pubkeys[i];
+        if (key.IsValid()) {
+            num_pubkeys++;
+            keyIndex = i;
+        }
+    }
+
+    CMutableTransaction& tx = *psbtx.tx;
+
+    // TODO(gwillen): Replace all this with the 'bonus output' scheme to use an OP_RETURN to balance blinders, with a rangeproof exponent of -1 (public).
```
instagibbs: with known blinding pubkey, right?

src/rpc/rawtransaction.cpp
gwillen: Yes, and I suspect that there's more pruning I could do with the goal of keeping the PSBTs small (which I expect is why we do that), but this specific removal is actually important as a signal to signpsbt that it's okay to sign. (Since the hack with the nonce field is gone in this flow.)

src/rpc/rawtransaction.cpp
gwillen: I believe the scheme we talked about -- which I don't expect to fully understand before implementing it -- was to just make up a random pubkey and then throw it away after, because nobody ever needs to be able to rewind these outputs.

src/rpc/rawtransaction.cpp
instagibbs: why wouldn't the presence of the other fields show that?

src/rpc/rawtransaction.cpp
instagibbs: that works too. Could just be the generator point G or something stupid as well.

src/rpc/rawtransaction.cpp
gwillen: The only way to tell that an output is _intended_ to be blinded is that it has a pubkey. Absence of blinders could mean "not going to be blinded" or "hasn't been blinded yet".

gwillen: So I guess you could indeed check for "has a pubkey and blinders" or "has no pubkey", but just checking for not having a pubkey seems easier (and is similar to what the code was doing before). But I can go either way as you like.

src/rpc/rawtransaction.cpp
gwillen: Well, the PSBT can encode anything the CTransaction can encode, so I can make a PSBT of an issuance -- I just can't blind it. (Perhaps I also can't sign it -- I didn't try a fully unblinded one.)

gwillen:
Rebased, still need to split things up a bit. Also looks like I killed CI with the rebase, will investigate.

src/wallet/wallet.cpp
gwillen: done

src/wallet/wallet.cpp
gwillen: I think this must have gone away in the rebase

instagibbs: ah yeah

gwillen:
apparently somehow I lost the compiler warning squashing in the rebase, but not because it was fixed upstream? So I added it back as its own commit.

Also fixed tests by removing all the stray printfs, none of them appeared worthy of upgrading to LogPrintf.

I think this should be ready for a real review.

gwillen:
fixed travis complaints (coming from lints that don't run on my system because they have some python dependency I can't get to install.)

instagibbs:
Linters are hard to replicate locally for me as well. Thanks.

instagibbs:
ok looks like everything but bitcoin compatibility test for psbt is working:

```
test_framework.authproxy.JSONRPCException: TX decode failed Duplicate Key, output asset_commitment already provided: iostream error (-22)
```

gwillen:
Hm, interesting. It only fails on a single platform, which makes me worry it's nondeterministic. Also, I didn't realize those compatibility tests existed, so I wasn't running them -- I will try and see what it does locally.

ALSO, I think I'm still waiting for you to help me figure out where to condition on g_con_elementsmode. So assuming the compatibility tests run without that, that may fix it?

gwillen:
@instagibbs formal reminder that I am waiting for your review on this (and guidance on where to put g_con_elementsmode checks)

instagibbs:
@stevenroose promised a review Soon(TM), I think after his next catchup PR.

gwillen:
@stevenroose @instagibbs piiiing? :-) I would really like to get this cleaned up and in. I definitely still need help sorting out what g_con_elementsmode is for, and where it should fit in (to make the bitcoin backwards-compat tests pass, which still fail at present.)

instagibbs:
please add tests for `walletcreatefundedpsbt`, I don't think it works correctly. It's adding the old hack-style nonce commitments as pubkeys and not filling out the blinding pubkeys as psbt fields.

gwillen:
Ok so, belated comment on your last comment @instagibbs -- there already were tests for walletcreatefundedpsbt, but nothing is actually testing that things that should be blinded are, and nothing is testing that nonces don't show up where they shouldn't.

I have fixed walletcreatefundedpsbt and converttopsbt, and tested by hand, and the tests pass; but I haven't gone back and made the tests check that expected blinding is actually happening.

I expect that the code itself should be done at this point, though -- can you take another look?

src/rpc/rawtransaction.cpp
```diff
@@ -1294,7 +1294,8 @@ UniValue blindpsbt(const JSONRPCRequest& request)
     if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
         throw std::runtime_error(
             "blindpsbt \"psbt\" ( ignoreblindfail )\n"
-            "\nUse the blinding data from the PSBT inputs to generate the blinding data for the PSBT outputs.\n"
+            "\nUse the blinding data from the PSBT inputs to generate the blinding data for the PSBT outputs.\n\n"
+            "TODO: Not expected to work on issuance/reissuance/peg transactions yet.\n"
```
instagibbs: I'd rather not have a TODO in help text :) Just telling the user it isn't supported is enough

src/rpc/rawtransaction.cpp
```diff
@@ -2018,14 +2019,20 @@ UniValue converttopsbt(const JSONRPCRequest& request)
...
-    psbtx.tx = tx;
     for (unsigned int i = 0; i < tx.vin.size(); ++i) {
         psbtx.inputs.push_back(PSBTInput());
     }
     for (unsigned int i = 0; i < tx.vout.size(); ++i) {
         psbtx.outputs.push_back(PSBTOutput());
+        // At this point, if the nonce field is present it should be a smuggled
+        //   pubkey, and not a real nonce. Convert it back to a pubkey and strip
+        //   it out.
+        psbtx.outputs[i].blinding_pubkey = CPubKey(tx.vout[i].nNonce.vchCommitment);
```
instagibbs: do we want to sanity check it's a `IsFullyValid` key?

src/wallet/rpcwallet.cpp
```diff
@@ -4275,10 +4275,12 @@ void FillPSBTInputsData(const CWallet* pwallet, PartiallySignedTransaction& psbt
...
 bool SignPSBT(const CWallet* pwallet, PartiallySignedTransaction& psbtx, int sighash_type, bool sign, bool imbalance_ok)
 {
-    // Check that the transaction is not still in need of blinding
-    for (const PSBTOutput& o : psbtx.outputs) {
-        if (o.blinding_pubkey.IsValid()) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, "PSBT must be blinded before signing; call blindpsbt.");
+    // If we're signing, check that the transaction is not still in need of blinding
+    if (sign) {
+        for (const PSBTOutput& o : psbtx.outputs) {
+            if (o.blinding_pubkey.IsValid()) {
```
instagibbs: I usually prefer `IsFullyValid()`

test/functional/rpc_psbt.py
```diff
@@ -11,6 +11,8 @@
 import json
 import os
 
+import IPython
```
instagibbs: que?

gwillen: whoop lol

was using this for debugging. Removed.

src/wallet/rpcwallet.cpp
gwillen: Leaving this per discussion.

src/rpc/rawtransaction.cpp
gwillen: Leaving this per discussion.

src/rpc/rawtransaction.cpp
gwillen: Yeah fair. Fixing.

gwillen:
Addressed comments, still have to debug the issue I uncovered with the walletcreatefundedpsbt change output coming out improperly unblinded.

gwillen:
Sigh, ok, gave up on doing createfundedpsbt in (what I felt was) the nice way. The mystery of FundTransaction runs way too deep. Instead, I let FundTransaction do its thing, then extract the pubkeys from the nonces afterwards.

Still need to rebase on 0.18.

instagibbs:
@gwillen right, seems to be the least painful thing to do. We let Core-ish stuff alone if it makes rebasing upstream easier.

gwillen:
@instagibbs Heh... I didn't even think about the ease-of-rebase issue. ^^;

I am still working on the rebase onto 0.18. Currently the rpc_psbt.py test is erroring out immediately at the top of the very first test with "insufficient funds", which confuses me and I'm not sure what's up.

gwillen:
Rebased onto master.

After a bunch of fixups, the errors left from Travis are (1) an apparent memory-exhaustion running tsan, and (2) a timeout. Not sure what's the best way to deal with those but they're not obviously something I can fix.

instagibbs:
Sounds like the builds we haven't been able to get working from upstream.
Will take a look Monday!

src/wallet/rpcwallet.cpp
```diff
@@ -4370,62 +4370,131 @@ void AddKeypathToMap(const CWallet* pwallet, const CKeyID& keyID, std::map<CPubK
...
+                {
+                    {"psbt", RPCArg::Type::STR, RPCArg::Optional::NO, "The transaction base64 string"},
+                    {"sighashtype", RPCArg::Type::STR, /* default */ "ALL", "The signature hash type to sign with if not specified by the PSBT. Must be one of\n"
+                        "       \"ALL\"\n"
+                        "       \"NONE\"\n"
+                        "       \"SINGLE\"\n"
+                        "       \"ALL|ANYONECANPAY\"\n"
+                        "       \"NONE|ANYONECANPAY\"\n"
+                        "       \"SINGLE|ANYONECANPAY\""},
+                    {"imbalance_ok", RPCArg::Type::BOOL, /* default */ "false", "Sign even if the transaction amounts do not balance"},
```
instagibbs: this needs a quick test, just create an imbalanced PSBT and try both options to make sure one errors and other doesn't at least

test/functional/rpc_psbt.py
```diff
@@ -158,56 +195,85 @@ def run_test(self):
...
 
+        for out in rt1['vout']:
+            if out['scriptPubKey']['type'] == "fee":
+                next
+            elif out['scriptPubKey']['addresses'][0] == node1_unconf_addr:
+                vout1 = out['n']
+
+        for out in rt2['vout']:
+            if out['scriptPubKey']['type'] == "fee":
+                next
```
instagibbs: not sure `next` is the thing you want? Seems to be an iterator thing. My guess is it's just a no-op but the logic is otherwise correct

test/functional/rpc_psbt.py
```diff
@@ -158,56 +195,85 @@ def run_test(self):
...
-        # Combine, finalize, and send the psbts
-        combined = self.nodes[0].combinepsbt([psbt1, psbt2])
-        finalized = self.nodes[0].finalizepsbt(combined)['hex']
-        self.nodes[0].sendrawtransaction(finalized)
         self.nodes[0].generate(6)
         self.sync_all()
 
+        for out in rt1['vout']:
+            if out['scriptPubKey']['type'] == "fee":
+                next
```
instagibbs: not sure `next` is the thing you want? Seems to be an iterator thing. My guess is it's just a no-op but the logic is otherwise correct

test/functional/rpc_psbt.py
```diff
@@ -286,7 +357,120 @@ def run_test(self):
         # Unload extra wallets
         for i, signer in enumerate(signers):
             self.nodes[2].unloadwallet("wallet{}".format(i))
+    """
+
+    def run_ca_tests(self):
```
instagibbs: For these tests let's put in some output-checking tests, making sure they're blinded(or not) based on the scenario. I think in this case we can just check that *all* outputs aside from fee are blinded or not.

src/psbt.h
```diff
@@ -27,11 +27,24 @@ static constexpr uint8_t PSBT_IN_WITNESSSCRIPT = 0x05;
 static constexpr uint8_t PSBT_IN_BIP32_DERIVATION = 0x06;
 static constexpr uint8_t PSBT_IN_SCRIPTSIG = 0x07;
 static constexpr uint8_t PSBT_IN_SCRIPTWITNESS = 0x08;
+static constexpr uint8_t PSBT_IN_VALUE = 0x09;
```
instagibbs: Let's move these values to somewhere vendor-specific

src/psbt.h
```diff
@@ -294,6 +385,48 @@ struct PSBTOutput
...
+            SerializeToVector(s, PSBT_OUT_VALUE_COMMITMENT);
+            SerializeToVector(s, value_commitment);
+        }
+
+        if (!value_blinding_factor.IsNull()) {
+            SerializeToVector(s, PSBT_OUT_VALUE_BLINDER);
+            SerializeToVector(s, value_blinding_factor);
+        }
+
+        // XXX: If g_con_elementsmode is false, something strange and unexpected happens here.
```
instagibbs: ???

src/psbt.h
```diff
@@ -479,6 +692,7 @@ struct PartiallySignedTransaction
                     // Make sure that all scriptSigs and scriptWitnesses are empty
                     for (unsigned int i = 0; i < tx->vin.size(); i++) {
                         const CTxIn& txin = tx->vin[i];
+                        tx->witness.vtxinwit.resize(tx->vin.size());
```
instagibbs: resize it outside the loop?

src/rpc/rawtransaction.h
```diff
@@ -17,6 +17,6 @@ class Chain;
 UniValue SignTransaction(interfaces::Chain& chain, CMutableTransaction& mtx, const UniValue& prevTxs, CBasicKeyStore *keystore, bool tempKeystore, const UniValue& hashType);
 
 /** Create a transaction from univalue parameters */
-CMutableTransaction ConstructTransaction(const UniValue& inputs_in, const UniValue& outputs_in, const UniValue& locktime, const UniValue& rbf, const UniValue& assets_in);
+CMutableTransaction ConstructTransaction(const UniValue& inputs_in, const UniValue& outputs_in, const UniValue& locktime, const UniValue& rbf, const UniValue& assets_in, std::vector<CPubKey>* output_pubkeys_out = nullptr);
```
instagibbs: can you document `output_pubkeys_out`? Looks like it's a flag for hack key in nonce :)

test/functional/rpc_psbt.py
instagibbs: Same with `walletcreatefundedpsbt`, just make sure an expected-to-be-blinded psbt ends up getting blinded...

instagibbs:
Please update the OP to list some of the common usage sequences for RPC.

Also an explicit list of the new fields in the description is probably apt.

instagibbs:
oh could you also say what the PR isn't doing in the OP text? e.g., doesn't handle issuance/peg-ins.

src/rpc/rawtransaction.h
gwillen: Hm, it's possible that it could go away, actually. Originally, it was not so much a flag, as an alternative way to get the pubkeys from ConstructTransaction -- if it was non-null, the pubkeys would be put into output_pubkeys_out _instead_ of the nonces.

However, I ended up abandoning this mechanism (I think completely?) in favor of letting it stuff nonces, then extracting the nonces later. This is because I was unable to adequately control FundTransaction's behavior -- it was insisting on stuffing pubkeys into nonces when creating _change_, and I gave up on trying to dig deeply enough to understand how to prevent it.

I do think we should switch to something like this mechanism in the future, to get rid of the nonce-stuffing hack entirely. But I think PSBT needs to be full-featured and regular unsigned transactions deprecated first, so perhaps the world is not ready for this yet.

src/psbt.h
gwillen: Heh, that would be clever. I think I did this in the rebase.

src/psbt.h
gwillen: Oh, I think all this stuff needs to be gated on g_con_elementsmode.

The specific unexpected thing that happens is that asset_commitment is NOT null; instead it is some kind of special Bitcoin value, so the asset gets printed even when elementsmode is off.

src/psbt.h
gwillen: Right, sounds like we're moving them up to start at 0xF0.

test/functional/rpc_psbt.py
gwillen: Oh, lol, I meant 'continue'. Which is called 'next'.... in Perl.

test/functional/rpc_fundrawtransaction.py
```diff
@@ -485,7 +485,9 @@ def run_test(self):
         rawtx = self.nodes[2].createrawtransaction(inputs, outputs)
         fundedTx = self.nodes[2].fundrawtransaction(rawtx)
 
-        signedTx = self.nodes[2].signrawtransactionwithwallet(fundedTx['hex'])
+        blindedTx = self.nodes[2].blindrawtransaction(fundedTx['hex'])
+        signedTx = self.nodes[2].signrawtransactionwithwallet(blindedTx)
+        #fail next line: test_framework.authproxy.JSONRPCException: Transaction output has nonce, but is not blinded. Did you forget to call blindpsbt, blindrawtranssaction, or rawblindrawtransaction? (-25)
```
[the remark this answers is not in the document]
gwillen: Oh, no, that comment should go away.

This was already broken before my PR, because addresses are confidential by default, but the test was not calling blindrawtransaction. It was caught by my PR because I added the "no bullshit nonces left" check in sendrawtransaction, which caught the transaction in the test going out unblinded.

gwillen:
Updated the PR description.

test/functional/rpc_fundrawtransaction.py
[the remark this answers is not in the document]
gwillen: Fixed.

src/wallet/rpcwallet.cpp
```diff
@@ -4370,62 +4370,131 @@ void AddKeypathToMap(const CWallet* pwallet, const CKeyID& keyID, std::map<CPubK
...
+    UniValue result(UniValue::VOBJ);
+    result.pushKV("psbt", EncodePSBT(psbtx));
+    result.pushKV("complete", complete);
+
+    return result;
 }
 
 UniValue walletprocesspsbt(const JSONRPCRequest& request)
 {
+    if (!g_con_elementsmode)
```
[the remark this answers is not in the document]
gwillen: In !g_con_elementsmode, a bunch of psbt-related stuff is broken now. I don't know that I really trust any of the code that I fiddled with to operate in non-elements-mode. Certainly none of it has been tested there. So we had discussed just gating everything on elementsmode being on, since that's the only mode that Elements is really meant to be operated in anyway.

This RPC is noted in its help to be deprecated (at least, implicitly via the recommendation to use other RPCs instead.) I could give a stronger deprecation warning, or just remove it entirely. The sequence of walletfillpsbtdata / blindpsbt / walletsignpsbt is a complete replacement for it. It will still work in elementmode, if the transaction to be processed does not have any blinded outputs.

src/wallet/rpcwallet.cpp
```diff
@@ -4370,62 +4370,131 @@ void AddKeypathToMap(const CWallet* pwallet, const CKeyID& keyID, std::map<CPubK
...
+                    {"imbalance_ok", RPCArg::Type::BOOL, /* default */ "false", "Sign even if the transaction amounts do not balance"},
+                },
+                RPCResult{
+                    "{\n"
+                    "  \"psbt\" : \"value\",          (string) The base64-encoded partially signed transaction\n"
+                    "  \"complete\" : true|false,   (boolean) If the transaction has a complete set of signatures\n"
+                    "}\n"
+                },
+                RPCExamples{
+                    HelpExampleCli("walletprocesspsbt", "\"psbt\"")
```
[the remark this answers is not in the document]
gwillen: fixed

src/wallet/rpcwallet.cpp
```diff
@@ -4370,62 +4370,131 @@ void AddKeypathToMap(const CWallet* pwallet, const CKeyID& keyID, std::map<CPubK
...
+                    + HelpExampleRpc("walletprocesspsbt", "\"psbt\"")
```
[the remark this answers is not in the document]
gwillen: fixed

test/functional/rpc_psbt.py
gwillen: woops fixed, see below.

test/functional/rpc_psbt.py
gwillen: (But actually 'pass' will work fine here, fixed.)

src/wallet/rpcwallet.cpp
gwillen: Added a test!

src/psbt.h
gwillen: Done.

src/psbt.h
gwillen: Gated all this behind g_con_elementsmode.

src/psbt.h
gwillen: Done.

test/functional/rpc_psbt.py
gwillen: Done!

gwillen:
Addressed all comments. PTAL?

instagibbs:
Changes look good, one more linter:

```
./test/functional/rpc_psbt.py:493:58: E703 statement ends with a semicolon
266^---- failure generated from test/lint/lint-python.sh
```

gwillen:
Lol, it's really hard for me to not type semicolons at the end of lines. I blame C/C++ for this. Fixed. (One day I will get the python lints working on OS X.)

gwillen:
Added a sentence of explanation of output_pubkeys_out in ConstructTransaction. I did leave it in because there's still once place I did use it.

gwillen:
Squashed down to a few sensible commits.

instagibbs:
ACK

gwillen:
Thanks @instagibbs! Been a long road for this one.

instagibbs:
appreciate the heavy lifting!

src/rpc/rawtransaction.cpp
```diff
@@ -1284,6 +1310,126 @@ static std::string WriteHDKeypath(std::vector<uint32_t>& keypath)
...
+        }
+    }
+
+    for (auto& output : psbtx.outputs) {
+        output_pubkeys.push_back(output.blinding_pubkey);
+    }
+
+    // How many are we trying to blind?
+    int num_pubkeys = 0;
+    unsigned int keyIndex = -1;
```
apoelstra, line 1381: Flagged by ubsan for being an implicit signedness conversion (which is not UB but ubsan lints for it). But also this is literally UB because keyIndex is never checked against this -1 value, it's just used as an array index.

apoelstra: Should be fixed by #900, just noting here for posterity.

apoelstra: Oh, this block appears to have been copied from rawblindrawtransaction which has had this issue since at least https://github.com/ElementsProject/elements/pull/515

apoelstra: Oh, my bad, in both cases it is actually safe becasue `keyIndex` is only -1 if `num_pubkeys` is left at 0, and `keyIndex` is only used as an index when `num_pubkeys` is nonzero.

PR #607, opened 2019-05-03
Fix default permitsigdata value not working on converttopsbt command

stevenroose:
We're not really supporting PSBT for Elements yet. The `rpc_psbt.py` integration test is disabled because of this; that's also why we didn't catch this. It's fixed upstream. 

I added an extra test to catch the error, but it's not actually executed.

@instagibbs, up to you to decide if it's worth having this patched right now. I cherry-picked that fix commit literally from the catchup PR I'm working on, so if we don't patch it here, it will get patched while catching up.

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -1765,13 +1765,13 @@ UniValue converttopsbt(const JSONRPCRequest& request)
 
     // Remove all scriptSigs and scriptWitnesses from inputs
     for (CTxIn& input : tx.vin) {
-        if ((!input.scriptSig.empty()) && (request.params[1].isNull() || (!request.params[1].isNull() && request.params[1].get_bool()))) {
+        if (!input.scriptSig.empty() && !permitsigdata) {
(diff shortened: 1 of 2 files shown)


instagibbs:
-0 on fixing this versus merging with upstream, but not a hard NACK

If this gets in the way of useful PSBT usage, I'd probably change my mind. @gwillen might have better idea based on his PR.

gwillen:
Oh, I remember noticing that and probably should have filed it at the time I noticed it. I believe the effect of the bug is that "permitsigdata" is actually "prohibitsigdata", defaulting to true (that is, the sense of the flag is reversed if you supply it but the default is correct.)

I can't imagine when you would run into this; trying to convert a signed transaction to a (unsigned) PSBT doesn't really seem like a sensible thing to ever do. So I don't have strong feelings about fixing it.

stevenroose:
I changed my mind; I'd like to just pull this in here. Please review & ACK :)

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/607/commits/794c305cdda1fe1d2ae0454e6809170eb63386ea

Issue #613, opened 2019-05-07
claimpegin should not succeed if it is double-claiming a tx already in mempool or blockchain


stevenroose:

instagibbs:
To elaborate, the wallet should not report success(a txid).

Instead:
1) Check `IsPeginSpent` for UTXO db check
2) Check if that input is already in the mempool as spent.

PR #614, opened 2019-05-07
Improve double pegin claim behavior

### LINKED ISSUE #612: Transaction that double-spends a pegin should not be accepted in mempool

stevenroose:
Fixes #612 and #613.

- prevents a double claim from entering mempool
- makes `claimpegin` throw an exception when trying to double claim a pegin
- check the last commit message on all the extra cases that are tested.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -681,10 +681,27 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
         }
 
         // do all inputs exist?
-        for (const CTxIn& txin : tx.vin) {
+        for (unsigned int i = 0; i < tx.vin.size(); i++) {
+            const CTxIn& txin = tx.vin[i];
+
             // ELEMENTS:
-            // Don't look for coins that only exist in parent chain
+            // For pegin inputs check whether the pegins have already been claimed before.
+            // This only checks the UTXO set for already claimed pegins. For mempool conflicts,
+            // we rely on the GetConflictTx check done above.
             if (txin.m_is_pegin) {
+                // Quick sanity check on witness first.
+                if (tx.witness.vtxinwit.size() <= i ||
+                        tx.witness.vtxinwit[i].m_pegin_witness.stack.size() < 6 ||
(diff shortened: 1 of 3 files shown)


src/validation.cpp
```diff
@@ -681,10 +681,25 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
         }
 
         // do all inputs exist?
-        for (const CTxIn& txin : tx.vin) {
+        for (unsigned int i = 0; i < tx.vin.size(); i++) {
+            const CTxIn& txin = tx.vin[i];
+
             // ELEMENTS:
-            // Don't look for coins that only exist in parent chain
+            // For pegin inputs check whether the pegins have already been claimed before.
```
instagibbs: need to note this is just a coin UTXO check, not a mempool check at all. At the mempool level we rely on the `GetConflictTx` check.

instagibbs:
Can you also explain the regression from 0.14 in the OP for people reading code history?

src/validation.cpp
stevenroose: I added some explanation.

stevenroose:
> Can you also explain the regression from 0.14 in the OP for people reading code history?

Please verify the commit message of the new commit.

stevenroose:
I think I also addressed all comments on the other PR about using `raises_rpc_exception` and the case-by-case explanation.

src/wallet/rpcwallet.cpp
```diff
@@ -5317,6 +5317,15 @@ UniValue claimpegin(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "TX decode failed");
     }
 
+    // To check if it's not double spending an existing pegin UTXO, we check mempool acceptance.
+    CValidationState acceptState;
+    bool accepted = ::AcceptToMemoryPool(mempool, acceptState, MakeTransactionRef(mtx), nullptr /* pfMissingInputs */,
+                            nullptr /* plTxnReplaced */, false /* bypass_limits */, maxTxFee, true /* test_accept */);
```
instagibbs: ah, cute!

instagibbs:
ACK!

PR #663, opened 2019-06-12
Add support for blinding pegins

stevenroose:
Adds pegin handling support for `blindrawtransaction` and `rawblindrawtransaction`.


test/functional/feature_fedpeg.py
```diff
@@ -524,6 +531,63 @@ def run_test(self):
...
+        sidechain.sendtoaddress(blind_addr, 15)
+        sidechain.generate(5)
+        # start pegin
+        pegin_addrs = sidechain.getpeginaddress()
+        assert_equal(sidechain.decodescript(pegin_addrs["claim_script"])["type"], "witness_v0_keyhash")
+        pegin_addr = addrs["mainchain_address"]
+        txid_fund = parent.sendtoaddress(pegin_addr, 10)
+        # 10+2 confirms required to get into mempool and confirm
+        parent.generate(11)
+        time.sleep(2)
```
instagibbs: what is sleep for?

test/functional/feature_fedpeg.py
```diff
@@ -565,28 +565,44 @@ def run_test(self):
...
+            assert("value" not in final_decoded["vout"][1])
+            assert("asset" not in final_decoded["vout"][1])
+            assert(final_decoded["vout"][1]["commitmentnonce_fully_valid"])
+            assert("value" in final_decoded["vout"][2])
+            assert("asset" in final_decoded["vout"][2])
+            # check that it is accepted in the mempool
+            accepted = sidechain.testmempoolaccept([pegin_signed["hex"]])[0]
+            if not accepted["allowed"]:
+                print("Blinded pegin not allowed in mempool; reject-reason: %s" % accepted["reject-reason"])
+                raise Exception()
```
instagibbs: just stick the message inside the `Exception`

test/functional/feature_fedpeg.py
```diff
@@ -565,28 +565,44 @@ def run_test(self):
...
+            print("Blinded transaction looks ok!") # need this print to distinguish failures in for loop
+        # check if they get mined; since we're trying to mine two double spends, disconnect the nodes
+        disconnect_nodes(sidechain, 3)
+        disconnect_nodes(sidechain2, 2)
+        txid1 = sidechain.sendrawtransaction(pegin_signed1["hex"])
+        blocks = sidechain.generate(3)
+        assert_equal(sidechain.getrawtransaction(txid1, True, blocks[0])["confirmations"], 3)
+        txid2 = sidechain2.sendrawtransaction(pegin_signed2["hex"])
+        blocks = sidechain2.generate(3)
+        assert_equal(sidechain2.getrawtransaction(txid2, True, blocks[0])["confirmations"], 3)
```
instagibbs: I'd prefer you put the test back in a usable state after this. i.e., re-connect the two nodes at least.

instagibbs:
looks good otherwise, and nice to see the python stuff put to work :P

test/functional/feature_fedpeg.py
stevenroose: no idea :D it was done after the parent send in the rest of the test as well. I think it'll work without here, not sure Travis :D

test/functional/feature_fedpeg.py
stevenroose: haha, nice!

test/functional/feature_fedpeg.py
stevenroose: hmm yeah I was thinking that it was not needed but someone extending this test should do it. I'll connect them. I was looking for a way to discard the tx from the mempool so I could mine the first one, invalidate block, discard and mine the other one, but I couldn't find how.

test/functional/feature_fedpeg.py
stevenroose: removed it and pushed, let's see :) it works locally

stevenroose:
> and nice to see the python stuff put to work :P

Yeah nice to see that we can do all that stuff manually. I think the only thing that we can't do without liquid is the blinding, but that should be possible in Rust with rust-secp256k1-zkp. Very excited about that!

stevenroose:
@instagibbs build is fine it seems

instagibbs:
re:"Always return blind address for "blech32" type"

The help for getnewaddress should reflect this.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/663/commits/cc491065147276193e39095ee0f54b913a37180d

PR #674, opened 2019-07-08
[rpc]blindrawtransaction accepts more commitments

AkioNak:
In #510 / #550 a requirement that the blinding commitments were equal in amount to the inputs was introduced. This requirement is correct for the final case, but not for partial transactions (e.g. where party 1 blinds their inputs before even handing the transaction over to party 2 to add theirs). In order to hide party 1's input amounts from party 2, blindrawtransaction must allow >= the inputs, since the commitments are shared beforehand (if not, the blinding fails; this may need further investigating).

This PR restores the functionality to allow blinding commitments more than or equal to the amount of inputs, which makes the confidential assets demo https://github.com/ElementsProject/confidential-assets-demo functional again.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -5759,8 +5759,8 @@ UniValue blindrawtransaction(const JSONRPCRequest& request)
     std::vector<std::vector<unsigned char> > auxiliary_generators;
     if (request.params.size() > 2) {
         UniValue assetCommitments = request.params[2].get_array();
-        if (assetCommitments.size() != 0 && assetCommitments.size() != tx.vin.size()) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, "Asset commitment array must have exactly as many entries as transaction inputs.");
+        if (assetCommitments.size() != 0 && assetCommitments.size() < tx.vin.size()) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Asset commitment array must have at least as many entries as transaction inputs.");
         }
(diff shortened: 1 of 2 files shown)


instagibbs:
If we re-give this ability it should be documented in the RPCs that use it. It was hidden functionality IIRC.

kallewoof:
@instagibbs Without this, people have to provide placeholder inputs before they can start blinding stuff, and people have to know which entries to replace. How is that better?

instagibbs:
Sorry I must have been unclear: please change the help to make it clear
this functionality exists.

instagibbs:
e.g., `rawblindrawtransaction`

kallewoof:
Oh, I misunderstood. Sorry about that. You are saying the `rawblindtransaction` RPC docs need to be updated?

instagibbs:
Actually this is used in `blindrawtransaction`, currently text says 

> "An array of input asset generators. If provided, this list must be empty, or match the final input commitment list, including ordering, to make a valid surjection proof. This list does not include generators for issuances, as these assets are inherently unblinded."

I suppose this is technically correct but might want to make it even more explicit that this "final list" can include inputs not given in `hexstring`.

instagibbs:
I'll touch up the help myself, ACK https://github.com/ElementsProject/elements/pull/674/commits/d584b65c18d11af417048d86308cf2043e4d7654

AkioNak:
@instagibbs Oh, I'm sorry to bother you.

instagibbs:
No worries I just don't want to hold up this improvement as is.

PR #683, opened 2019-07-30
 TxToUniv: check vchRangeproof.empty() before secp256k1_rangeproof_info()

dgpv:
rangeproof may be empty, and then secp256k1_rangeproof_info will fail on `ARG_CHECK(proof != NULL);`

### DIFF
--- a/src/core_write.cpp
+++ b/src/core_write.cpp
@@ -327,7 +327,7 @@ void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry,
             uint64_t minv;
             uint64_t maxv;
             const CTxOutWitness* ptxoutwit = tx.witness.vtxoutwit.size() <= i? NULL: &tx.witness.vtxoutwit[i];
-            if (ptxoutwit && secp256k1_rangeproof_info(secp256k1_blind_context, &exp, &mantissa, &minv, &maxv, &ptxoutwit->vchRangeproof[0], ptxoutwit->vchRangeproof.size())) {
+            if (ptxoutwit && !ptxoutwit->vchRangeproof.empty() && secp256k1_rangeproof_info(secp256k1_blind_context, &exp, &mantissa, &minv, &maxv, &ptxoutwit->vchRangeproof[0], ptxoutwit->vchRangeproof.size())) {
                 if (exp == -1) {
                     out.pushKV("value", ValueFromAmount((CAmount)minv));
                 } else {


dgpv:
```
e1-cli sendrawtransaction 02000000010161616161616161616161616161616161616161616161616161616161616161610000000000ffffffff0100000000000000000000010100000000
```
results in
```

#0  0x00007ffff65686ab in raise () from /lib64/libc.so.6
#1  0x00007ffff6551539 in abort () from /lib64/libc.so.6
#2  0x0000555555e199bb in secp256k1_default_illegal_callback_fn (str=<optimized out>, data=<optimized out>) at src/secp256k1.c:52
#3  0x0000555555e2b8fc in secp256k1_callback_call (cb=<optimized out>, cb=<optimized out>, text=0x555555f35057 "proof != NULL")
    at src/util.h:24
#4  secp256k1_rangeproof_info (ctx=<optimized out>, exp=<optimized out>, mantissa=<optimized out>, min_value=<optimized out>, 
    max_value=<optimized out>, proof=<optimized out>, plen=0) at src/modules/rangeproof/main_impl.h:230
#5  0x0000555555cb2768 in TxToUniv (tx=..., hashBlock=..., entry=..., include_hex=false, serialize_flags=0) at core_write.cpp:330
#6  0x00005555557c3f42 in decoderawtransaction (request=...) at rpc/rawtransaction.cpp:684
```

This is because the transaction contains nValue in txout that have empty commitment. It is considered 'not explicit', but it is also not confidential. The code then proceeds to use non-existent rangeproof value.

To me, it seems that empty nValue (empty commitment) is not valid at all, and probably this should have been caught in the transaction decoding code.

instagibbs:
> To me, it seems that empty nValue (with empty commitment) should not be allowed at all, and probably this should have been caught in the transaction decoding code.

We're not going to change serialization rules at this point, just fix the mis-use.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/683/commits/981ac1b7ac9f775928aa5c6af145a6128b20619f

dgpv:
changed the check to use `!ptxoutwit->vchRangeproof.empty()` instead of `ptxoutwit->vchRangeproof.size()`

instagibbs:
was going to nit that, but thought against it ;) even better! re-utACK

instagibbs:
The recent PSBT PR IIRC fixed this. Can you close if you agree or rebase otherwise?

dgpv:
Right, https://github.com/ElementsProject/elements/commit/2848b520b2b169bc9e80a1b8198e3e5500004fa1 fixed this, closing.

PR #730, opened 2019-10-02
Change Proprietary types to use the method specified in BIP 174

achow101:
https://github.com/bitcoin/bips/pull/849 (not yet merged) adds to BIP 174 proprietary types. This PR changes the PSBT types used in Elements to be in line with the proposed changes to the BIP/

### DIFF
--- a/src/psbt.h
+++ b/src/psbt.h
@@ -665,7 +665,11 @@ struct PartiallySignedTransaction
     inline void Serialize(Stream& s) const {
 
         // magic bytes
-        s << PSBT_MAGIC_BYTES;
+        if (g_con_elementsmode) {
+            s << PSBT_ELEMENTS_MAGIC_BYTES;
+        } else {
+            s << PSBT_MAGIC_BYTES;
+        }
 
         // unsigned tx flag
         SerializeToVector(s, PSBT_GLOBAL_UNSIGNED_TX);
@@ -699,8 +703,14 @@ struct PartiallySignedTransaction
         // Read the magic bytes
         uint8_t magic[5];
         s >> magic;
-        if (!std::equal(magic, magic + 5, PSBT_MAGIC_BYTES)) {
-            throw std::ios_base::failure("Invalid PSBT magic bytes");
+        if (g_con_elementsmode) {
+            if (!std::equal(magic, magic + 5, PSBT_ELEMENTS_MAGIC_BYTES)) {
+                throw std::ios_base::failure("Invalid PSBT magic bytes");
+            }
+        } else  {
+            if (!std::equal(magic, magic + 5, PSBT_MAGIC_BYTES)) {
+                throw std::ios_base::failure("Invalid PSBT magic bytes");
+            }
         }
 
         // Read global data
--- a/src/psbt.h
+++ b/src/psbt.h
@@ -13,9 +13,11 @@
 
 // Magic bytes
 static constexpr uint8_t PSBT_MAGIC_BYTES[5] = {'p', 's', 'b', 't', 0xff};
+static constexpr uint8_t PSBT_ELEMENTS_MAGIC_BYTES[5] = {'p', 's', 'e', 't', 0xff};
 
 // Global types
 static constexpr uint8_t PSBT_GLOBAL_UNSIGNED_TX = 0x00;
+static constexpr uint8_t PSBT_GLOBAL_PROPRIETARY = 0xFC;
 
 // Input types
 static constexpr uint8_t PSBT_IN_NON_WITNESS_UTXO = 0x00;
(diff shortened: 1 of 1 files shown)


instagibbs:
concept ACK, will review

src/psbt.h
```diff
@@ -27,26 +28,30 @@ static constexpr uint8_t PSBT_IN_WITNESSSCRIPT = 0x05;
...
+static constexpr uint8_t PSBT_OUT_VALUE_BLINDER = 0x01;
+static constexpr uint8_t PSBT_OUT_ASSET_COMMITMENT = 0x02;
+static constexpr uint8_t PSBT_OUT_ASSET_BLINDER = 0x03;
+static constexpr uint8_t PSBT_OUT_RANGE_PROOF = 0x04;
+static constexpr uint8_t PSBT_OUT_SURJECTION_PROOF = 0x05;
+static constexpr uint8_t PSBT_OUT_BLINDING_PUBKEY = 0x06;
+static constexpr uint8_t PSBT_OUT_NONCE_COMMITMENT = 0x07;
+
+// Proprietary type identifer string
+static const std::string proprietary_id("elements");
```
instagibbs: non-blocking nit: `PSBT_ELEMENTS_ID` or something

stevenroose: I would also have made it a constant.

stevenroose:
Really cool! So the proprietary types are already accepted into the BIP, right?

stevenroose:
utACK, constant variable name nit.

stevenroose:
On a second thought. Are current Elements "PSBT"s compatible with Bitcoin's? Since the tx format is already different from Bitcoin's, an Elements "PSBT" won't be parseable by Bitcoin PSBT logic.

Irregardless of using the proprietary keys for out own special fields, shouldn't we update the magic value to something like `"pset"`, perhaps?

instagibbs:
> So the proprietary types are already accepted into the BIP, right?

The PR for it is still open IIRC, but I haven't heard any objections.

> shouldn't we update the magic value to something like "pset", perhaps?

Perhaps we should, also to make it clear that we're making breaking changes.

instagibbs:
Is this missing changes like the compact size wrapper here for non-proprietary fields? https://github.com/bitcoin/bitcoin/pull/17034/commits/6ccda5b023c6a35d49184c5d964d3c4e5171e6e3#diff-4d99ef1a2a2826595a92ac62c4162790R69

Should we work on upstream merge first to make this simpler?

achow101:
> On a second thought. Are current Elements "PSBT"s compatible with Bitcoin's? Since the tx format is already different from Bitcoin's, an Elements "PSBT" won't be parseable by Bitcoin PSBT logic.
> 
> Irregardless of using the proprietary keys for out own special fields, shouldn't we update the magic value to something like `"pset"`, perhaps?

Right, perhaps just change the magic, but keep everything else the same to reduce code drift? I have tried to decode an Elements PSBT with Core, and it fails on decoding the global tx.



> Is this missing changes like the compact size wrapper here for non-proprietary fields? [bitcoin/bitcoin@6ccda5b#diff-4d99ef1a2a2826595a92ac62c4162790R69](https://github.com/bitcoin/bitcoin/commit/6ccda5b023c6a35d49184c5d964d3c4e5171e6e3#diff-4d99ef1a2a2826595a92ac62c4162790R69)

Yes.  I noticed that the changes in https://github.com/bitcoin/bitcoin/pull/17034 were largely incompatible or unrelated to these changes, so I decided not to include them.

> Should we work on upstream merge first to make this simpler?

Perhaps, but I don't think they are related closely enough.

instagibbs:
> unrelated to these changes

If we're making changes to be near-compatible with new bip174, I feel like we should go "all the way" to reduce drift? Maybe I'm misunderstanding.

achow101:
> If we're making changes to be near-compatible with new bip174,

It is still compatible, the changes to BIP 174 were designed to be compatible with existing implementations.

> I feel like we should go "all the way" to reduce drift? Maybe I'm misunderstanding.

Those changes should merge cleanly with the changes (modulo the proprietary types) in Core, so a rebase would be sufficient.

It mostly just didn't make sense to me to PR the same commits to both Elements and Core when Elements will just rebase onto Core and pick up those commits later. They aren't particularly important.

instagibbs:
fair enough, so `pset` or whatever, plus my one nit should be enough

src/psbt.h
achow101: Changed. Not quite sure what you mean by "make it a constant", it's already const.

achow101:
Added a commit to use `pset` as the magic bytes when in elements mode.

instagibbs:
utACK https://github.com/ElementsProject/elements/pull/730/commits/34e1350114eee021618a4b71daaa8e131ac434af

src/psbt.h
stevenroose: "formatted as a constant" sorry :)

stevenroose:
LGTM! I'm gonna merge this despite the failure. It's an unrelated test and I don't have the error when run locally.

ACK  34e1350

PR #751, opened 2019-10-18
Peg-in fields for PSBT

achow101:
Adds peg-in fields for PSBT. `createpsbt` uses the new peg-in raw tx api to allow specifying the peg-in data and then adds it to the psbt. `decodepsbt`, `finalizepsbt` and `walletsignpsbt` are both updated to handle peg-in stuff too. Also tests.

`walletcreatefundedpsbt` still does not handle peg-in data. It will require handling external inputs which the wallet does not do. I think it would be best to first implement external inputs in `FundTransaction` upstream before allowing `walletcreatefundedpsbt` to accept peg-ins.


instagibbs:
concept ACK though:

```
  File "test/functional/feature_fedpeg.py", line 296, in run_test
    assert_equal(decoded_psbt['inputs'][0]['pegin_bitcoin_tx'], raw)
```
fails for the bitcoind parent run because `raw` here is the "segwit" serialization of the raw tx, and the version in the psbt is the witness-stripped(legacy) version. Not sure the cleanest way to deal with that and maintain coverage. You could I suppose compare the `decoderawtransaction` `pegin_witness` field and make sure the hex matches to make sure it's computed the same way at least.

achow101:
Fixed by sending both `raw` and `pegin_bitcoin_tx` through `decoderawtransaction` and comparing their txids.

src/rpc/rawtransaction.cpp
```diff
@@ -2062,6 +2062,9 @@ UniValue createpsbt(const JSONRPCRequest& request)
                                     {"txid", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "The transaction id"},
                                     {"vout", RPCArg::Type::NUM, RPCArg::Optional::NO, "The output number"},
                                     {"sequence", RPCArg::Type::NUM, /* default */ "depends on the value of the 'replaceable' and 'locktime' arguments", "The sequence number"},
+                                    {"pegin_bitcoin_tx", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "The raw bitcoin transaction (in hex) depositing bitcoin to the mainchain_address generated by getpeginaddress"},
+                                    {"pegin_txout_proof", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "A rawtxoutproof (in hex) generated by the mainchain daemon's `gettxoutproof` containing a proof of only bitcoin_tx"},
+                                    {"pegin_claim_script", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "The witness program generated by getpeginaddress."},
```
instagibbs: it's the script with version byte and witness program, not just witness program

PR #754, opened 2019-10-23
Update link from README.md to assets tutorial

philippem:


instagibbs:
utACK, the shell script now lives elsewhere

PR #788, opened 2019-11-21
gui: Make balances scrollable on overview

### LINKED ISSUE #685: elements-qt grows too large for screen once you have many assets
I have dozens of assets in my wallet, and now the window grows too large that it cannot stay on my screen and I cannot do anything that would normally be at the bottom of the window.

To solve this, we should have a scroll on the asset list on the overview.

achow101:
Makes the balances on the overview page scrollable when therer are lots of assets. It should be both vertically and horizontally scrollable.

Also Fixes the vertical layout so it uses the UI forms and looks better.

Fixes #685

### DIFF
--- a/src/qt/forms/overviewpage.ui
+++ b/src/qt/forms/overviewpage.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>596</width>
-    <height>419</height>
+    <height>781</height>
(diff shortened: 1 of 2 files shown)


instagibbs:
Seems to work, thanks

instagibbs:
got another out of bound tACK

PR #807, opened 2020-01-22
Replace Greg with Steven as security contact

stevenroose:


instagibbs:
utACK https://github.com/ElementsProject/elements/pull/807/commits/6ff3125fc48e3d0a5120bdd399b5684910a09abb fingerprint matches known public key

PR #808, opened 2020-01-28
Add DEFAULT chain and set to liquidv1

stevenroose:
This fixes f.e. that when arguments are set in CLI or the config file
without a network attached (`port` instead of `liquidv1.port`), they
will now be assigned to the liquidv1 instead of Bitcoin mainnet.

------

With 
```toml
port=9953
```
Before:
```
➜  elements git:(master) ✗ ./src/elementsd -chain=elementsregtest | rg 9953
Warning: Config setting for -port only applied on elementsregtest network when in [elementsregtest] section.
➜  elements git:(master) ✗ ./src/elementsd -chain=main | rg 9953           
➜  elements git:(master) ✗ ./src/elementsd -chain=liquidv1 | rg 9953   
Warning: Config setting for -port only applied on main network when in [liquidv1] section.
```
After:
```
➜  elements git:(master) ✗ ./src/elementsd -chain=elementsregtest | rg 9953
Warning: Config setting for -port only applied on elementsregtest network when in [elementsregtest] section.
➜  elements git:(master) ✗ ./src/elementsd -chain=main | rg 9953           
Warning: Config setting for -port only applied on main network when in [main] section.
➜  elements git:(master) ✗ ./src/elementsd -chain=liquidv1 | rg 9953
```

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -547,9 +547,11 @@ fs::path static StartupShortcutPath()
     std::string chain = gArgs.GetChainName();
     if (chain == CBaseChainParams::MAIN)
         return GetSpecialFolderPath(CSIDL_STARTUP) / "Bitcoin.lnk";
+    if (chain == CBaseChainParams::LIQUID1)
+        return GetSpecialFolderPath(CSIDL_STARTUP) / "Liquid.lnk";
     if (chain == CBaseChainParams::TESTNET) // Remove this special case when CBaseChainParams::TESTNET = "testnet4"
         return GetSpecialFolderPath(CSIDL_STARTUP) / "Bitcoin (testnet).lnk";
-    return GetSpecialFolderPath(CSIDL_STARTUP) / strprintf("Bitcoin (%s).lnk", chain);
+    return GetSpecialFolderPath(CSIDL_STARTUP) / strprintf("Elements (%s).lnk", chain);
 }
 
 bool GetStartOnSystemStartup()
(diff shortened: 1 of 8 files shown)


instagibbs:
concept ACK

src/qt/guiutil.cpp
```diff
@@ -673,10 +677,10 @@ bool SetStartOnSystemStartup(bool fAutoStart)
         // Write a bitcoin.desktop file to the autostart directory:
         optionFile << "[Desktop Entry]\n";
         optionFile << "Type=Application\n";
-        if (chain == CBaseChainParams::MAIN)
-            optionFile << "Name=Bitcoin\n";
+        if (chain == CBaseChainParams::LIQUID1)
+            optionFile << "Name=Liquid\n";
```
gwillen: Based on the other changes in this PR, I would expect this to be adding LIQUID1, not replacing MAIN. (But I don't really know what I'm looking for.) Is it correct as written here?

gwillen:
See my question in guiutil.cpp.

Some further notes, based on a global search for "CBaseChainParams::MAIN":

* Consider changing the comment on ArgsManager::GetChainName() in src/util/system.h, which is wrong for Elements, even though you didn't actually touch it in this PR. (I assume the hardcoded string constant in the function itself, in system.cpp, is not something in your scope to change at this time, even though it is a default that is not linked to your definition of DEFAULT.)
* I am confused that GetUnrecognizedSections in system.cpp does not misbehave in elements (also not touched by your PR.)
* Similar to the first point, in bitcoin-cli.cpp there is a list of chains, and a default of MAIN, which neither includes liquid nor uses the DEFAULT you're defining here.

Since you are probably more of an expert than I am on all the things I have listed here:
utACK, please take a look at my comments and then do what you want as appropriate.

src/qt/guiutil.cpp
stevenroose: Ah yeah sure, that makes sense. Leaving MAIN in place is the goal as it reduces diff. Good catch.

stevenroose:
Yeah so in Elements we allow for unrecognized chains. Any chain not in the list of known chains is just a random Elements chain with the given name. We always use f.e. `elementsregtest`, but that's not actually a known chain name, could just as well have had another name. Any name not in the set of known names is just Elements defaults + the parameters in CLI or config.

But yeah I agree we should leave MAIN around as much as possible.

I addressed your comments by
- putting back that MAIN usage you commented on
- changing the comment on `GetChainName` and using `::DEFAULT` as default chain, good catch as well!
- removing `GetUnrecognizedSections` because it's unused in Elements

gwillen:
utACK, thanks for the additional changes. Let me know once Travis is happy and you've updated #809  and I'll go do that one.

PR #811, opened 2020-02-04
Add link to RPC docs in readme

FrancisPouliot:
The RPC docs are a great way to understand the difference between Bitcoin and Elements. 

It should be highlighted along with the other differences from Bitcoin Core in the same section.

Alternatively, add them to Elements/Doc

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Elements deferred for additional research and standardization:
  * [Schnorr Signatures][schnorr-signatures]
 
 Additional RPC commands and parameters:
-* [RPC Docs](https://github.com/ElementsProject/elementsbp-api-reference/blob/master/api.md)
+* [RPC Docs](https://elementsproject.org/en/doc/)
 
 License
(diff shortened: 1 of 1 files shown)


instagibbs:
yeah looks like we lost the docs in their modern place: https://github.com/ElementsProject/elementsbp-api-reference/ see deprecated message. We need to update the docs, maybe starting with elements-specific calls.

stevenroose:
Could you perhaps update this PR by using this link to the documentation: https://elementsproject.org/en/doc/

FrancisPouliot:
@stevenroose hadn't seen that actually. Thanks for sharing! Would be nice if you also added the elements-specific changes perhaps by making the RPC calls bold or something

FrancisPouliot:
Done

stevenroose:
Hmm just notice that the README still says "additional RPC ..." while the linked docs show all.. Perhaps change that accordingly?

PR #817, opened 2020-02-14
WIP: Streaming SHA256 ops.

roconnor-blockstream:
My proposal is

* `OP_SHA256INITIALIZE` to pop a bytestring and push SHA256 context creating by adding the bytestring to the initial SHA256 context.
* `OP_SHA256UPDATE` to pop a SHA256 context and bytestring, and push an updated context by adding the bytestring to the data stream being hashed.
* `OP_SHA256FINALIZE` to pop a SHA256 context  and bytestring, and push a SHA256 hash value after adding the bytestring and completing the padding.

I propose defining the SHA256-context as a bytestring that is a 40 to 103 bytes in length where
* The first 32-bytes is the SHA256 midstate
* The next 8-bytes is the (little-endian) unsigned integer value containing the number of **bits** of data processed so far.  This value will always be a multiple of 8.
* The next 0 to 63 bytes is the store of data processed but not compressed.  The length of this "field" is `X/8 % 64` (equiv. `(X % 512) / 8`) where `X` is the integer value of the previous field.

Any invalid SHA256-context will cause the operations that process the context to fail.
If any input would cause the SHA256 bit counter to overflow, the operation will fail.

I could easily be convinced to use bytes instead of bits in the second field.  Bits is nicely aligned with how SHA-256 works, and the size is perfectly aligned.  OTOH Bytes is what the SHA256 internal state uses in practice and would be marginally faster to processes.

**Motivation**:  There already exists a `OP_SH256` opcode that performs computes SHA256 computations.  However because the `MAX_SCRIPT_ELEMENT_SIZE` is 520 bytes, that is the maximum message size that can `OP_SHA256` can operate on.  These proposed operations allow us to circumvent the `MAX_SCRIPT_ELEMENT_SIZE` limit on message size; however these opcode might be subject to abuse (e.g. [SHA-256 with freestart](https://crypto.stackexchange.com/questions/48580/fixed-point-of-the-sha-256-compression-function)).

### DIFF
--- a/src/crypto/sha256.cpp
+++ b/src/crypto/sha256.cpp
@@ -716,6 +716,60 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned 
     }
 }
 
+std::vector<unsigned char> CSHA256::Save() const {
+    size_t buf_size = bytes % 64;
+    std::vector<unsigned char> result(40 + buf_size);
+
+    WriteBE32(&result[ 0], s[0]);
+    WriteBE32(&result[ 4], s[1]);
+    WriteBE32(&result[ 8], s[2]);
+    WriteBE32(&result[12], s[3]);
+    WriteBE32(&result[16], s[4]);
+    WriteBE32(&result[20], s[5]);
+    WriteBE32(&result[24], s[6]);
+    WriteBE32(&result[28], s[7]);
+
+    WriteLE64(&result[32], bytes << 3);
+
+    memcpy(&result[40], buf, buf_size);
+
+    return result;
+}
+
+bool CSHA256::Load(const std::vector<unsigned char>& vch) {
+    if (vch.size() < 40) return false;
+
+    uint64_t bits = ReadLE64(&vch[32]);
+    size_t buf_size = (bits >> 3) % 64;
+
+    if ((bits & 0x07) != 0 || vch.size() != 40 + buf_size) return false;
+
+    // We want to leave the internal state of the object unchanged if false is returned.
+    // So no member variables can be modified until now.
+
+    s[0] = ReadBE32(&vch[ 0]);
+    s[1] = ReadBE32(&vch[ 4]);
+    s[2] = ReadBE32(&vch[ 8]);
+    s[3] = ReadBE32(&vch[12]);
+    s[4] = ReadBE32(&vch[16]);
+    s[5] = ReadBE32(&vch[20]);
+    s[6] = ReadBE32(&vch[24]);
+    s[7] = ReadBE32(&vch[28]);
+
+    bytes = bits >> 3;
+
+    memcpy(buf, &vch[40], buf_size);
+
+    return true;
+}
+
+bool CSHA256::SafeWrite(const unsigned char* data, size_t len) {
+  const uint64_t SHA256_MAX = 0x1FFFFFFFFFFFFFFF; // SHA256's maximum allowed message length in bytes.
+  if (SHA256_MAX < bytes || SHA256_MAX - bytes < len) return false;
+  Write(data, len);
+  return true;
+}
+
 CSHA256& CSHA256::Reset()
 {
     bytes = 0;
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -1432,6 +1432,65 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript& 
                 }
                 break;
 
+                case OP_SHA256INITIALIZE: // (in -- sha256_ctx)
+                {
+                    if (stack.size() < 1)
+                        return set_error(serror, SCRIPT_ERR_INVALID_STACK_OPERATION);
+
+                    CSHA256 ctx;
+                    valtype& vch = stacktop(-1);
+                    if (!ctx.SafeWrite(vch.data(), vch.size()))
+                        return set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
+
+                    popstack(stack);
+                    stack.push_back(ctx.Save());
+                }
+                break;
+
+                case OP_SHA256UPDATE: // (sha256_ctx in -- sha256_ctx)
+                {
+                    if (stack.size() < 2)
+                        return set_error(serror, SCRIPT_ERR_INVALID_STACK_OPERATION);
+
+                    CSHA256 ctx;
+                    valtype& vchCtx = stacktop(-2);
+                    if (!ctx.Load(vchCtx))
+                        return set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
+
+                    valtype& vch = stacktop(-1);
+                    if (!ctx.SafeWrite(vch.data(), vch.size()))
+                        return set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
+
+                    popstack(stack);
+                    popstack(stack);
+                    stack.push_back(ctx.Save());
+                }
+                break;
+
+                case OP_SHA256FINALIZE:
+                {
+                    // (sha256_ctx in -- hash)
+                    if (stack.size() < 2)
+                        return set_error(serror, SCRIPT_ERR_INVALID_STACK_OPERATION);
+
+                    valtype& vchCtx = stacktop(-2);
+                    CSHA256 ctx;
+                    if (!ctx.Load(vchCtx))
+                        return set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
+
+                    valtype& vch = stacktop(-1);
+                    if (!ctx.SafeWrite(vch.data(), vch.size()))
+                        return set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
+
+                    valtype vchHash(32);
+                    ctx.Finalize(vchHash.data());
+
+                    popstack(stack);
+                    popstack(stack);
+                    stack.push_back(vchHash);
+                }
+                break;
+
                 default:
                     return set_error(serror, SCRIPT_ERR_BAD_OPCODE);
             }
--- a/src/script/script.cpp
+++ b/src/script/script.cpp
@@ -130,6 +130,9 @@ const char* GetOpName(opcodetype opcode)
     case OP_DETERMINISTICRANDOM    : return "OP_DETERMINISTICRANDOM";
     case OP_CHECKSIGFROMSTACK      : return "OP_CHECKSIGFROMSTACK";
     case OP_CHECKSIGFROMSTACKVERIFY: return "OP_CHECKSIGFROMSTACKVERIFY";
+    case OP_SHA256INITIALIZE       : return "OP_SHA256INITIALIZE";
+    case OP_SHA256UPDATE           : return "OP_SHA256UPDATE";
+    case OP_SHA256FINALIZE         : return "OP_SHA256FINALIZE";
 
     // expansion
     case OP_NOP1                   : return "OP_NOP1";
--- a/src/script/script.h
+++ b/src/script/script.h
@@ -184,6 +184,9 @@ enum opcodetype
     OP_DETERMINISTICRANDOM = 0xc0,
     OP_CHECKSIGFROMSTACK = 0xc1,
     OP_CHECKSIGFROMSTACKVERIFY = 0xc2,
+    OP_SHA256INITIALIZE = 0xc4,
+    OP_SHA256UPDATE = 0xc5,
+    OP_SHA256FINALIZE = 0xc6,
 
     // expansion
     OP_NOP1 = 0xb0,
@@ -203,7 +206,7 @@ enum opcodetype
 };
 
 // Maximum value that an opcode can be
-static const unsigned int MAX_OPCODE = OP_SUBSTR_LAZY; // 0xc3
+static const unsigned int MAX_OPCODE = OP_SHA256FINALIZE; // 0xc6
 
 const char* GetOpName(opcodetype opcode);
 
--- a/src/crypto/sha256.h
+++ b/src/crypto/sha256.h
@@ -8,6 +8,7 @@
 #include <stdint.h>
 #include <stdlib.h>
 #include <string>
+#include <vector>
 
 /** A hasher class for SHA-256. */
 class CSHA256
@@ -26,7 +27,10 @@ public:
     //TODO: Midstate is a hack'ish speedup that probably should make way for something
     //akin to the SHA256D64 speedups
     void Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned char *buffer);
+    std::vector<unsigned char> Save() const;
+    bool Load(const std::vector<unsigned char>& vch);
     CSHA256& Reset();
+    bool SafeWrite(const unsigned char* data, size_t len);
 };
 
 /** Autodetect the best available SHA256 implementation.


src/crypto/sha256.cpp
```diff
@@ -716,6 +716,63 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned
     }
 }
 
+std::vector<unsigned char> CSHA256::Save() {
+   std::vector<unsigned char> result;
+   result.reserve(40 + (bytes % 64));
+
+   for(auto i = 0; i < 8; ++i) {
+     result.push_back(static_cast<unsigned char>(s[i] >> 24));
```
sipa: You can use `WriteBE32` here instead.

src/script/interpreter.cpp
```diff
@@ -1432,6 +1432,46 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript&
...
+                {
+                    if (stack.size() < (opcode == OP_SHA256_WRITE ? 2 : 1))
+                        return set_error(serror, SCRIPT_ERR_INVALID_STACK_OPERATION);
+
+                    CSHA256 ctx;
+                    if (opcode == OP_SHA256_WRITE) {
+                        if (!ctx.Load(stacktop(-2)))
+                            return set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
+                    }
+                    valtype& vch = stacktop(-1);
```
sipa: I think this line and the next must be inside the `(opcode == OP_SHA256_WRITE)` conditional.

src/crypto/sha256.cpp
```diff
@@ -716,6 +716,63 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned
...
+                  | static_cast<uint64_t>(vch[34]) << 16
+                  | static_cast<uint64_t>(vch[35]) << 24
+                  | static_cast<uint64_t>(vch[36]) << 32
+                  | static_cast<uint64_t>(vch[37]) << 40
+                  | static_cast<uint64_t>(vch[38]) << 48
+                  | static_cast<uint64_t>(vch[39]) << 56;
+
+    if ((bits & 0x07) != 0 || vch.size() != 40 + ((bits >> 3) % 64)) return false;
+
+    bytes = bits >> 3;
```
dgpv: nitpick: you can calculate `bytes` before the previous statement and then use `bytes` instead of `bits >> 3` in the size check

src/script/interpreter.cpp
roconnor-blockstream: While my original proposal was `OP_SHA256_IV`, at the last minute I decided that in practice almost every use of `OP_SHA256_IV` would be immediately followed by `OP_SHA256_WRITE`.  Therefore I've modified my proposal to define `OP_SHA256_INIT` which combines the behaviour of  `OP_SHA256_IV` and `OP_SHA256_WRITE`.  The original behaviour of `OP_SHA256_IV` can be simulated by the script `OP_0 OP_SHA256_INIT`.

src/script/interpreter.cpp
```diff
@@ -1432,6 +1432,46 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript&
                 }
                 break;
 
+                case OP_SHA256_INIT: // (in -- sha256_ctx)
```
dgpv: I think separating processing of these two opcodes in different code blocks might be beneficial: it will improve the readability and lessen the number of branches in the code, while not significantly increasing the size of the code (cache impact)

src/crypto/sha256.cpp
roconnor-blockstream: I'm torn on the issue, because it means that the 3 most significant bits of the `uint64_t` end up getting truncated and end up having no semantic value.  I don't know how I feel about that.  Depending on how people use these new opcodes, it could potentiality become a source of witness malleability.

Honestly I see myself being swayed to using bytes eventually, but I'm going to resist just a little while longer.

dgpv: the 3 least significant bits are checked with `bits & 0x07`, they are not ignored. What I had in mind is
```
bytes = bits >> 3;

if ((bits & 0x07) != 0 || vch.size() != 40 + (bytes % 64)) return false;
```

dgpv: Granted, it is slightly less readable because of bytes/bits visual similarity.

src/crypto/sha256.cpp
roconnor-blockstream: I would have to do an allocation rather than a reservation, but I think that would be a good idea.  That would also let me use `memcpy` below.

src/crypto/sha256.cpp
dgpv: might be more readable as two separate checks

```
if ((bits & 0x07) != 0) return false;

bytes = bits >> 3;

if (vch.size() != 40 + (bytes % 64)) return false;
```

src/crypto/sha256.cpp
```diff
@@ -716,6 +716,63 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned
     }
 }
 
+std::vector<unsigned char> CSHA256::Save() {
+   std::vector<unsigned char> result;
+   result.reserve(40 + (bytes % 64));
```
dgpv: nitpick: you could calculate `buf_size` here and use `40 + buf_size`

src/crypto/sha256.cpp
roconnor-blockstream: Ah.  I deliberately wanted to leave the internal state of the CSHA256 unchanged if the `false` is returned here.  Clearly this is worth adding a comment to the code.

src/script/interpreter.cpp
roconnor-blockstream: I agree.  Combining them turned out way worse than I was anticipating.

src/crypto/sha256.cpp
dgpv: definitely.

dgpv:
re `OP_SHA256_UPDATE_AND_FINALIZE`: Is finalize useful without a preceding write ? In a linear script without loops it seems that it is almost always will be preceded by some write, and if not, `OP_0 OP_SHA256_UPDATE_AND_FINALIZE` should be an equivalent

roconnor-blockstream:
I've updated the proposal so that `FINALIZE` also does an update, just like `INIT`.

dgpv:
code looks good to me, utACK https://github.com/ElementsProject/elements/pull/817/commits/fb6840a757186ea4afcc74aac17f3b0b035328d8

src/script/script.cpp
```diff
@@ -130,6 +130,9 @@ const char* GetOpName(opcodetype opcode)
     case OP_DETERMINISTICRANDOM    : return "OP_DETERMINISTICRANDOM";
     case OP_CHECKSIGFROMSTACK      : return "OP_CHECKSIGFROMSTACK";
     case OP_CHECKSIGFROMSTACKVERIFY: return "OP_CHECKSIGFROMSTACKVERIFY";
+    case OP_SHA256_INIT            : return "OP_SHA256_INIT";
```
roconnor-blockstream: Might as well call it `OP_SHA256_INITIALIZE`.

dgpv: sounds better to me, rhymes with `OP_SHA256_FINALIZE`

src/crypto/sha256.cpp
```diff
@@ -716,6 +716,53 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned
...
+    WriteBE32(&result[ 0], s[0]);
+    WriteBE32(&result[ 4], s[1]);
+    WriteBE32(&result[ 8], s[2]);
+    WriteBE32(&result[12], s[3]);
+    WriteBE32(&result[16], s[4]);
+    WriteBE32(&result[20], s[5]);
+    WriteBE32(&result[24], s[6]);
+    WriteBE32(&result[28], s[7]);
+
+    WriteLE64(&result[32], bytes << 3);
```
roconnor-blockstream: We should probably fail if the [sha256 counter overflows](https://twitter.com/pwuille/status/1232120323622891520).

roconnor-blockstream:
I know it is bikeshedding, but I kinda want rename `SHA256_WRITE` (back) to `SHA256_UPDATE`.

instagibbs:
(just saw this) I presume this is some sort of simplicity groundwork?

src/crypto/sha256.cpp
dgpv: In case we work on user-supplied midstate ?

roconnor-blockstream:
> (just saw this) I presume this is some sort of simplicity groundwork?

Nope.  Taproot is effectively a Simplicity prerequisite, and I'd somewhat like these opcodes added along with taproot to make Covenants and Oracles a little bit easier by not having to contend with the 520 (or whatever) stack item limit.

instagibbs:
Oh I misread the top post, nevermind.

src/crypto/sha256.cpp
roconnor-blockstream: Yes; there IMHO there is no reasonable way to prevent user-supplied midstates (there are proposals for sha256 registers or a sha256 alt-stack, but I personally don't think they are reasonable.  Others people disagree). Because these opcodes are consensus critical we have to perfectly specify their behavior even on absurd inputs.

There are only two reasonable design choices here:
* The status quo of having the bit counter add up bits modulo 2^64 (which still would need to be documented).
* The proposed amendment of having the various operations cause script to fail if the bit counter would overflow.

dgpv: I think checking the limit on `Load()` will be safe, if the limit is set low enough. Let's say `MAX_SIZE - MAX_SCRIPT_ELEMENT_SIZE`. This way, the counter could not go beyond `MAX_SIZE`, and you cannot have witness data larger than that even theoretically.

src/crypto/sha256.cpp
```diff
@@ -716,6 +716,60 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned
...
+    bytes = bits >> 3;
+
+    memcpy(buf, &vch[40], buf_size);
+
+    return true;
+}
+
+bool CSHA256::SafeWrite(const unsigned char* data, size_t len) {
+  const uint64_t SHA256_MAX = 0x1FFFFFFFFFFFFFFF; // SHA256's maximum allowed message length in bytes.
+  if (SHA256_MAX < bytes || SHA256_MAX - bytes < len) return false;
```
dgpv: Checking for `SHA256_MAX - MAX_SCRIPT_ELEMENT_SIZE` in `Load()` would allow to get rid of three `if`s, because Write() will be always safe in regard to this check: the size of data on the stack cannot be bigger than `MAX_SCRIPT_ELEMENT_SIZE`. Why you think that doing this in `Write()` is better ? Note that in `OP_SHA256_INITIALIZE` it is even impossible to fail this check, because the `CTX` is new and can only increase up to `MAX_SCRIPT_ELEMENT_SIZE`. I think an `if` check is due in `Load()` and here this condition can just be checked with `assert` to be safe.

dgpv: Is it because you want to allow for `UPDATE` with user-supplied midstate to be able to go up to `SHA256_MAX` without any other limits ?

src/crypto/sha256.cpp
```diff
@@ -716,6 +716,60 @@ void CSHA256::Midstate(unsigned char hash[OUTPUT_SIZE], uint64_t* len, unsigned
...
+    return result;
+}
+
+bool CSHA256::Load(const std::vector<unsigned char>& vch) {
+    if (vch.size() < 40) return false;
+
+    uint64_t bits = ReadLE64(&vch[32]);
+    size_t buf_size = (bits >> 3) % 64;
+
+    if ((bits & 0x07) != 0 || vch.size() != 40 + buf_size) return false;
```
dgpv: Maybe add a comment to state that bits & 0x07 check is to prevent malleability ?

src/crypto/sha256.cpp
dgpv: You cold make `Load()` take the length and then check `SHA_MAX - bytes < len` inside `Load()`.
Although this makes Load/Save api less generic. But if you want it to be more generic so it can be used in other places, you probably want that check inside `Load()` anyway, because in general you cannot prevent anyone to use just `Write()` instead of `SafeWrite()` after `Load()`

src/script/script.h
```diff
@@ -184,6 +184,9 @@ enum opcodetype
     OP_DETERMINISTICRANDOM = 0xc0,
     OP_CHECKSIGFROMSTACK = 0xc1,
     OP_CHECKSIGFROMSTACKVERIFY = 0xc2,
+    OP_SHA256_INITIALIZE = 0xc4,
+    OP_SHA256_UPDATE = 0xc5,
+    OP_SHA256_FINALIZE = 0xc6,
```
dgpv: Note that using two underscores in op name would mean breaking (useless?) convention that there's always just one underscore in `OP_<NAME>`.

src/crypto/sha256.cpp
roconnor-blockstream: I suppose I could write `Load` to return a `int64_t` that returns the number of bytes used so far (or alternatively the number of remaining bytes available to be processed), and return `-1` for failure.

Is this really all worthwhile just to avoid one `if` statement in `SHA256INITALIZE`?

src/crypto/sha256.cpp
roconnor-blockstream: It's a bit more than that: the `SHA256` class is simply incapable of processing sub-byte data as it stands.  But I can write a comment to that effect.

src/crypto/sha256.cpp
roconnor-blockstream: I suppose we also get rid of `SafeWrite` which is kinda nice.

src/crypto/sha256.cpp
dgpv: I mean you could just just do `bits &= ~0x07` and not do the check, but this would introduce non-determinism, and the comment could tell the reader why it is not a good idea

src/crypto/sha256.cpp
dgpv: all three `if` statement for `Write` is not needed if the `Write` is always safe,  if the check `SHA_MAX - bytes < len` is done inside Load() - at least in regard to manipulating the state from the script. The `Load()` can just fail if the subsequent `Write()` will be invalid. But my opinion is not strong on this - this is mostly bikeshedding. But I would like to add that check inside `Load()` would follow 'fail early' principle, which I favour.

roconnor-blockstream: Having `Load` take in an argument of `the number of bytes I intend to write later` is such an unusual API, that I am very hesitant to implement it that way.

dgpv: Just call it `Prepare()` then :-)

dgpv: As I said, I have no strong opinion on this, and returning number of bytes available from `Load` and then checking it might be more messy than `SafeWrite` approach, because you will need to have that check in two places (in UPDATE and FINALIZE)

src/crypto/sha256.cpp
roconnor-blockstream: I had hoped it goes without saying that one shouldn't mangle an invalidly serialized data blob into a valid one.

dgpv: validity of serialized data depends on the definition of the validity, but it is not explicitly specified (the serialization code in `Save()` is only one possible implementation of implicit specification). But, on the other hand, you can view the `(bits & 0x07) != 0` as the statement of the de-facto specification "low-order bits should be zero", and in this light, no extra comment seems necessary.

JeremyRubin:
See some discussion of the concept here: https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-October/002218.html There was some discussion on if it should be allowed to pass in midstates or not.

I think a lot of scripts can benefit from a batch update style API where you specify how many things to take off the stack.

dgpv:
One data point re the linked 2019 discussion is that I encountered the limitation of OP_CAT in practice, where it limited the number of possible outputs (520 byte stack item limit) and thus one option of a complex covenant was not possible. Some of the outputs were blinded, and thus much larger than normal outputs, and I could fit two blinded outputs and two non-blinded outputs plus a fee output in a transaction, but three blinded outputs was too much. With streaming SHA256 operations, this limitation does not exist.

dgpv:
This is of course specific to Elements where blinded outputs are possible, without them, the 520 bytes limit might not be a problem for practical purposes.

JeremyRubin:
It's true for checktemplateverify that it imposes a limitation on how many outputs/inputs you can add dynamically, so it's true beyond elements.

dgpv:
This particular limitation has no connection to CTV, because fixing outputs hash via CHECKSIGFROMSTACK does not impose any limits by itself, as outputs hash that is checked comes from just a normal sighash for segwit. This is about dynamically generating this outputs hash within the script from the fixed part (that enforces the covenant) and dynamic part (explicit fees and non-restricted outputs). CAT is used for that. Full outputs data need to be constructed on the stack before doing SHA256 on it, and this is where the 520 byte stack item datasize limit becomes the factor.

JeremyRubin:
@dgpv that's exactly what I'm saying; it imposes a limit if you're dynamically adding data to a CTV template using op_cat. Hence sha256stream being preferred there too

gwillen:
This is kind of ancient I guess, but: Rather than have "update" plus "initialize-and-update", just have "update" and "finalize" treat an empty stack element as an empty context. Then you can ditch initialize entirely.

roconnor-blockstream:
gwillen, if I understand your comment, that would entail requiring an extra OP_0 opcode to create the empty context in all typical uses of this functionality.

apoelstra:
Closing, superceded by #1020

PR #818, opened 2020-02-18
Place the mainchain RPC warmup message in a better place

stevenroose:
Now it also shows with `-validatepegin=0`.


instagibbs:
utACK https://github.com/ElementsProject/elements/pull/818/commits/c5cb7148472c9f4419619888953e0130d3844d4f

PR #834, opened 2020-03-16
Switch default bits of hiding to 52

stevenroose:
Replaces https://github.com/ElementsProject/elements/pull/793 and is basically that MR but rebased. (I can't rebase on Greg's branch.)

### DIFF
--- a/src/blind.cpp
+++ b/src/blind.cpp
@@ -178,8 +178,11 @@ bool GenerateRangeproof(std::vector<unsigned char>& rangeproof, const std::vecto
     memcpy(asset_message+32, asset_blindptrs[asset_blindptrs.size()-1], 32);
 
     // Sign rangeproof
+    int ct_exponent = std::min(std::max((int)gArgs.GetArg("-ct_exponent", 0), -1), 18);
+    int ct_bits = (int)gArgs.GetArg("-ct_bits", 52);
     // If min_value is 0, scriptPubKey must be unspendable
-    int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &nRangeProofLen, scriptPubKey.IsUnspendable() ? 0 : 1, &value_commit, value_blindptrs.back(), nonce.begin(), std::min(std::max((int)gArgs.GetArg("-ct_exponent", 0), -1),18), (int)gArgs.GetArg("-ct_bits", 51), amount, asset_message, sizeof(asset_message), scriptPubKey.size() ? &scriptPubKey.front() : NULL, scriptPubKey.size(), &gen);
+    uint64_t min_value = scriptPubKey.IsUnspendable() ? 0 : 1;
+    int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &nRangeProofLen, min_value, &value_commit, value_blindptrs.back(), nonce.begin(), ct_exponent, ct_bits, amount, asset_message, sizeof(asset_message), scriptPubKey.size() ? &scriptPubKey.front() : NULL, scriptPubKey.size(), &gen);
     rangeproof.resize(nRangeProofLen);
     return (res == 1);
 }
--- a/src/blind.cpp
+++ b/src/blind.cpp
@@ -179,7 +179,7 @@ bool GenerateRangeproof(std::vector<unsigned char>& rangeproof, const std::vecto
 
     // Sign rangeproof
     // If min_value is 0, scriptPubKey must be unspendable
-    int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &nRangeProofLen, scriptPubKey.IsUnspendable() ? 0 : 1, &value_commit, value_blindptrs.back(), nonce.begin(), std::min(std::max((int)gArgs.GetArg("-ct_exponent", 0), -1),18), std::min(std::max((int)gArgs.GetArg("-ct_bits", 36), 1), 51), amount, asset_message, sizeof(asset_message), scriptPubKey.size() ? &scriptPubKey.front() : NULL, scriptPubKey.size(), &gen);
+    int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &nRangeProofLen, scriptPubKey.IsUnspendable() ? 0 : 1, &value_commit, value_blindptrs.back(), nonce.begin(), std::min(std::max((int)gArgs.GetArg("-ct_exponent", 0), -1),18), (int)gArgs.GetArg("-ct_bits", 51), amount, asset_message, sizeof(asset_message), scriptPubKey.size() ? &scriptPubKey.front() : NULL, scriptPubKey.size(), &gen);
     rangeproof.resize(nRangeProofLen);
     return (res == 1);
 }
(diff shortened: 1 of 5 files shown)


src/blind.cpp
```diff
@@ -178,8 +178,11 @@ bool GenerateRangeproof(std::vector<unsigned char>& rangeproof, const std::vecto
     memcpy(asset_message+32, asset_blindptrs[asset_blindptrs.size()-1], 32);
 
     // Sign rangeproof
+    int ct_exponent = std::min(std::max((int)gArgs.GetArg("-ct_exponent", 0), -1), 18);
```
instagibbs: heh concept ACK

stevenroose: It's just taking the arguments out of the giant line. Nothing else changed. I don't fully understand the strange min max thing there. I suppose it's to get the values between -1 and 18.

apoelstra:
utACK

src/blind.cpp
```diff
@@ -178,8 +178,11 @@ bool GenerateRangeproof(std::vector<unsigned char>& rangeproof, const std::vecto
...
+    int ct_bits = (int)gArgs.GetArg("-ct_bits", 52);
     // If min_value is 0, scriptPubKey must be unspendable
-    int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &nRangeProofLen, scriptPubKey.IsUnspendable() ? 0 : 1, &value_commit, value_blindptrs.back(), nonce.begin(), std::min(std::max((int)gArgs.GetArg("-ct_exponent", 0), -1),18), std::min(std::max((int)gArgs.GetArg("-ct_bits", 36), 1), 51), amount, asset_message, sizeof(asset_message), scriptPubKey.size() ? &scriptPubKey.front() : NULL, scriptPubKey.size(), &gen);
+    uint64_t min_value = scriptPubKey.IsUnspendable() ? 0 : 1;
+    int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &nRangeProofLen, min_value, &value_commit, value_blindptrs.back(), nonce.begin(), ct_exponent, ct_bits, amount, asset_message, sizeof(asset_message), scriptPubKey.size() ? &scriptPubKey.front() : NULL, scriptPubKey.size(), &gen);
```
jonasnick: If there's doubt as to whether 4174 is correct for `DEFAULT_RANGEPROOF_SIZE`, you can check/print it here because `rangeproof_sign` will write the actual size of of the proof into nRangeProofLen. I can't do this myself because I can't compile elements because my boost version is too new (ArchLinux).

jonasnick:
Note that we should urge users to update. With this change and the default minfee/fallbackfee changes, wallet fingerprinting is trivial.

stevenroose:
Yeah I'm sure Allen is taking care of that. Plus we're coordinating with Green to update at the same time.

src/blind.cpp
stevenroose: I actually checked using the unit test what the size was that it found.

stevenroose: Using this line: https://github.com/ElementsProject/elements/pull/834/files#diff-d1a88d936748c5bd171ffa9e67824385R268

src/test/blind_tests.cpp
```diff
@@ -265,17 +265,17 @@ BOOST_AUTO_TEST_CASE(naive_blinding_test)
...
         int mantissa = 0;
         uint64_t min_value = 0;
         uint64_t max_value = 0;
         BOOST_CHECK(secp256k1_rangeproof_info(ctx, &exp, &mantissa, &min_value, &max_value, tx4.witness.vtxoutwit[2].vchRangeproof.data(), proof_size) == 1);
-        BOOST_CHECK(exp == 0);
-        BOOST_CHECK(mantissa == 36); // 36 bit default
-        BOOST_CHECK(min_value == 1);
-        BOOST_CHECK(max_value == 68719476736);
+        BOOST_CHECK_EQUAL(exp, 0);
+        BOOST_CHECK_EQUAL(mantissa, 52); // 36 bit default
```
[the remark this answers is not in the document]
stevenroose: fixed

src/blind.cpp
dgpv: ct_bits was also previously limited via this min/max construction. It was changed to just 52, I don't see why ct_exponent in this line could not be changed to just 0. The min/max limit for ct_bits was removed in #793 without any explanation, was it deemed useless ?

dgpv: I ask because I'll need to adjust my lib accordingly, but I hesitate to remove the limitations without fully understanding the consequences. Changing the number of bits result in the change in the size of the proof, and it looks like this only affects the tx size and therefore fee estimation. Setting ct_bits outside sane range would mean the blinding will just not work and fail with error, is this change (removing min/max limitation) added to help detect incorrect settings rather than work with limited settings ? If this is the logic behind this change, then ct_exponent should not be min/max restricted, too, and then `secp256k1_rangeproof_sign_impl` will just return 0 if incorrect `exp` was supplied

stevenroose: I don't know if your reference to #793 was a typo, but that PR was superseded with this one and it just sets the default number of bits from 38 to 52. We did this because the 38 bits covered only up to 10^-3 BTC or something. Because some assets are issuing assets using the BTC amounts instead of satoshi, we did this to increase the default blinding about up to 45*10^6 BTC.

It's true that it will about double the tx size. But given that we're also reducing the tx fee rate for Liquid to 0.1 sat/byte, that more than compensates for that increase in tx size.

stevenroose: Ah you're probably right about the `min(max(` construction of ct_exponent. Any wrong number will result in a failure on the secp side. It might be worth it to remove that construction, yes.

dgpv: reference to #793 was not a typo, the change of ct_bits was not that my question was about. I asked why the `std::min(std::max(...` is still applied to `ct_exponent`, but not to `ct_bits` anymore. If there was certain logic behind removing this for `ct_bits`, the same logic should apply to `ct_exponent`. I don't see why this min/max thing is needed for either of these two parameters, but I might be missing something, and I would like to know the reasoning why min/max was there in the first place. I have this code ported to python, and I will need to sync the code, but I would like the code to be consistent, and having different approaches to `ct_exponent` and `ct_bits` does not seem consistent to me.

stevenroose: Removed that construction in followup PR. You were right that it was not needed in either position and it's actually a bit annoying.

dgpv:
What were a reason to choose CT_BITS=52 instead of 51, given that 51 bits also cover amounts up to MAX_MONEY (2251799813685248), you can't have outputs > MAX_MONEY in Elements, and MAX_MONEY is the same as in Core, `21000000 * COIN` ?

stevenroose:
I got an e-mail with this comment from Dmitry, but I can't find it on
GitHub, was it removed?

> What were a reason to choose CT_BITS=52 instead of 51, given that 51 bits
also cover amounts up to MAX_MONEY (2251799813685248), you can't have
outputs > MAX_MONEY in Elements, and MAX_MONEY is the same as in Core, 21000000
* COIN ?

Well, the reason to do this was not really to cover all Bitcoins. We won't
ever have that many Bitcoins in Liquid anyway. But the reason was that
people are issuing assets with bitcoin-like amounts (with 8 decimals). So
we needed to update to a high number to cover tokens with high amounts of
issuance. And we opted for 52 instead of 51 because it's a multiple of 4
and that made it easier to reason about the rangeproof size, because part
of it grows with the bits divided by 4, so 52 made more sense.

dgpv:
I removed it because I got an answer from @instagibbs here: https://github.com/ElementsProject/elementsproject.github.io/pull/127#discussion_r397837690

dgpv:
Thanks for an explanation of the reasoning behind the decision!

PR #835, opened 2020-03-17
Reduce mintxfee to 0.1 sat/vbyte and fallback fee to 2 sat/vb

stevenroose:
This follows the reduction of the `minrelaytxfee` in the previous release to the same value. Using this value, the wallet will be able to make transactions with a 0.1 sat/vbyte fee by default.


apoelstra:
utACK. Did a cursory check that there are no additional limits in `miner.cpp` that should be reduced.

jonasnick:
utACK bfb77ad44d03985284b80048d41a8a87d71e698d Checked that this is really the same as setting -fallbackfee and -mintxfee

PR #837, opened 2020-03-18
WIP: Add gitian-build.sh script for Gitian builds

stevenroose:
To be reviewed.

### DIFF
--- a/contrib/gitian-build.sh
+++ b/contrib/gitian-build.sh
@@ -1,0 +1,81 @@
+#!/bin/bash
+
+# This script is to be used to generate Gitian builds for Elements releases.
+# Requirements:
+# - One needs the OSX SDK file in the following location:
+#   bec9d089ebf2e2dd59b1a811a38ec78ebd5da18cbbcd6ab39d1e59f64ac5033f  $HOME/blockstream/gitian/MacOSX10.11.sdk.tar.gz
+#   The file can be downloaded here:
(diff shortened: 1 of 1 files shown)


contrib/gitian-build.sh
```diff
@@ -0,0 +1,81 @@
...
+git clone https://github.com/ElementsProject/elements.git ${CI_PROJECT_DIR}
+cd ${CI_PROJECT_DIR}
+mkdir /tmp/gitian/inputs
+cp $HOME/blockstream/gitian/MacOSX10.11.sdk.tar.gz /tmp/gitian/inputs/MacOSX10.11.sdk.tar.gz
+
+cd /tmp/gitian
+
+bin/make-base-vm --docker --suite bionic --arch amd64
+
+bin/gbuild --num-make ${NUM_JOB} --memory 10000 --url elements=${CI_PROJECT_DIR} --commit elements=${CI_COMMIT_TAG} ${CI_PROJECT_DIR}/contrib/gitian-descriptors/gitian-linux.yml
```
philippem, line 49: this could be wrapped in a for loop

```
for platform_config in gitian-linux.yml gitian-win.yml gitian-osx.yml
do
...
end
```

stevenroose: Hmm not sure I like that better. Like this it's very easy to comment parts out if something goes wrong and it's also very easy to do the steps manually by copy pasting them one by one. Which is probably what one should do using this for the first time.

PR #840, opened 2020-03-24
Remove the min(max( construction for the ct_exponent argument

stevenroose:
As per @dgpv's suggestion here: https://github.com/ElementsProject/elements/pull/834#discussion_r396333202

@dgpv could you review?


dgpv:
Looks good to me, this makes the code consistent in that both `ct_bits` and `ct_exponent` are not restricted and the expectation being that blinding in secp256k1-zkp will just fail if incorrect values are set to these parameters.

stevenroose:
Thanks! Could you review the backport as well: https://github.com/ElementsProject/elements/pull/841

PR #843, opened 2020-03-25
Support fee estimation for fees lower than 1 sat/byte

stevenroose:
The fee estimation buckets also have a minimum of 1000 sat/kB.

This PR lowers that limit and also discards old estimates files.

### DIFF
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -901,7 +901,9 @@ bool CBlockPolicyEstimator::Write(CAutoFile& fileout) const
 {
     try {
         LOCK(m_cs_fee_estimator);
-        fileout << 149900; // version required to read: 0.14.99 or later
+        //ELEMENTS: We upped this from 0.14.99 in upstream because
+        // we changed the bucket sizes.
+        fileout << 180107; // version required to read: 0.18.1.7
         fileout << CLIENT_VERSION; // version that wrote the file
         fileout << nBestSeenHeight;
         if (BlockSpan() > HistoricalBlockSpan()/2) {
@@ -938,6 +940,9 @@ bool CBlockPolicyEstimator::Read(CAutoFile& filein)
 
         if (nVersionRequired < 149900) {
             LogPrintf("%s: incompatible old fee estimation data (non-fatal). Version: %d\n", __func__, nVersionRequired);
+        } else if (nVersionThatWrote < 180107) {
+            //ELEMENTS: we changed the buckets in 0.18.1.7
+            LogPrintf("%s: incompatible old fee estimation data (non-fatal). Version: %d\n", __func__, nVersionThatWrote);
         } else { // New format introduced in 149900
             unsigned int nFileHistoricalFirst, nFileHistoricalBest;
             filein >> nFileHistoricalFirst >> nFileHistoricalBest;
--- a/src/policy/fees.h
+++ b/src/policy/fees.h
@@ -174,8 +174,8 @@ private:
      * invalidates old estimates files. So leave it at 1000 unless it becomes
      * necessary to lower it, and then lower it substantially.
      */
-    static constexpr double MIN_BUCKET_FEERATE = 1000;
-    static constexpr double MAX_BUCKET_FEERATE = 1e7;
+    static constexpr double MIN_BUCKET_FEERATE = 100;
+    static constexpr double MAX_BUCKET_FEERATE = 1e6;
 
     /** Spacing of FeeRate buckets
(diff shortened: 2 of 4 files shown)


src/policy/fees.cpp
```diff
@@ -901,7 +901,8 @@ bool CBlockPolicyEstimator::Write(CAutoFile& fileout) const
 {
     try {
         LOCK(m_cs_fee_estimator);
-        fileout << 149900; // version required to read: 0.14.99 or later
+        //ELEMENTS: we changed the buckets in 0.18.1.7
+        fileout << 180107; // version required to read: 0.14.99 or later
```
gwillen: Comment on this line needs updated?

src/policy/fees.h
```diff
@@ -174,8 +174,8 @@ class CBlockPolicyEstimator
      * invalidates old estimates files. So leave it at 1000 unless it becomes
      * necessary to lower it, and then lower it substantially.
      */
-    static constexpr double MIN_BUCKET_FEERATE = 1000;
-    static constexpr double MAX_BUCKET_FEERATE = 1e7;
+    static constexpr double MIN_BUCKET_FEERATE = 100;
```
gwillen: Per comment above, does it make sense to go lower, or is this as low as we'd likely ever want it?

gwillen:
utACK, feel free to ignore my two nits.

src/policy/fees.h
dgpv: Lowering the fee 10x at the same time makes spamming UTXO set 10x cheaper. And due to blinding, filtering the spam-UTXO out will be difficult. My opinion is that lowering the fee should have come with excise on outputs, so that creating txs with a lot of outputs to spam the UTXO set would still be costly.

src/policy/fees.cpp
stevenroose: Hmm, I don't think so. I left it op purpose. That's the version to read the file structure that I wanted to leave intact. The new version requirement is for the bucket sizes. I can make that more explicit if you want.

src/policy/fees.h
stevenroose: @dgpv so this is just the bucket size of course. AJ has an (abandoned) PR in Core where he lowers them to 8. I talked with Lawrence about also doing that, but it seems like we can just as well lower this the next time we lower the minrelay fee. We need a new release to lower the relay fee anyhow, so.

stevenroose:
I bumped into some strange thing, and I looked at the fee calculation code for half a day until I realized I was not calculating my own numbers correctly. But I added a commit with some small size estimation optimizations that might be helpful. The fee calculation logic is more robust than I expected :)

src/policy/fees.h
dgpv: Sorry, my comment is probably misplaced -- it is about lowering fees in general, not about the fee estimator

src/policy/fees.cpp
gwillen: Hm, I think I don't really understand what's going on in this section of code, maybe we can chat about it.

src/wallet/wallet.cpp
```diff
@@ -3266,6 +3267,14 @@ bool CWallet::CreateTransaction(interfaces::Chain::Lock& locked_chain, const std
 
                 // vouts to the payees
                 coin_selection_params.tx_noinputs_size = 11; // Static vsize overhead + outputs vsize. 4 nVersion, 4 nLocktime, 1 input count, 1 output count, 1 witness overhead (dummy, flag, stack size)
+
+                // Account for the fee output in the tx.
+                if (g_con_elementsmode && nSubtractFeeFromAmount == 0) {
```
gwillen: Hmm why condition on the value of nSubtractFeeFromAmount? That doesn't change whether there is a fee output taking up space, right?

src/wallet/wallet.cpp
```diff
@@ -3305,12 +3312,18 @@ bool CWallet::CreateTransaction(interfaces::Chain::Lock& locked_chain, const std
                             strFailReason = _("Transaction amount too small");
                         return false;
                     }
+
+                    // Include the fee cost for outputs. Note this is only used for BnB right now
+                    coin_selection_params.tx_noinputs_size += ::GetSerializeSize(txout, PROTOCOL_VERSION);
```
gwillen: I'm confused that we weren't already doing this? Isn't this a significant fraction of the size?

gwillen: Oh it just moved, I should read more carefully. Nevermind.

gwillen:
utACK

Issue #667, opened 2019-06-17
Inconsistent naming in the application


nwoodfine:
1) The application is called Elements Core but shows Liquid logo, would benefit from standardisation to avoid confusion.
2) When running `Liquid Core` for first time, dialog says `welcome to Elements Core`, and tells me that it will download and store a copy of the `Bitcoin block chain` - pretty sure this is referring to Liquid blockchain - this is obviously very confusing. Liquid is not even mentioned.
3) Same dialog box says the `Bitcoin blockchain` is 1GB big, containing transaction data from 2009 - specific naming of blockchain size seems risky if not dynamically generated, will go out of date quick. 

I was downloading Liquid binaries, so expected to see application called Liquid Core, and dialog reference to the Liquid block chain (sidechain).

![image](https://user-images.githubusercontent.com/17946939/59611133-fb36ff80-9144-11e9-846e-072cd71a94df.png)

Version: liquid-0.17.0-osx-unsigned.dmg

stevenroose:
The Elements Core vs Liquid is I think as we intend it to be. "Elements Core" is the software, "Liquid" is the network. So `liquid-qt` or "Liquid Core" doesn't exist anymore.

But the mentions of the Bitcoin blockchain are an issue. We traverse the code often for mentions of Bitcoin, and we often let one slip here or there.

stevenroose:
There's plenty of unhandles Bitcoin name mentions and it would be quite intensive to go over all of them:
```
$ rg -tcpp "\".*Bitcoin.*\"" src/qt | wc -l
56
```
It might be worth it to go over all those within strings.

PR #844, opened 2020-03-30
Fix some Bitcoin mentions in Qt

stevenroose:
Solves https://github.com/ElementsProject/elements/issues/667.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -390,7 +390,7 @@ void BitcoinApplication::shutdownResult()
 
 void BitcoinApplication::handleRunawayException(const QString &message)
 {
-    QMessageBox::critical(nullptr, "Runaway exception", BitcoinGUI::tr("A fatal error occurred. Bitcoin can no longer continue safely and will quit.") + QString("\n\n") + message);
+    QMessageBox::critical(nullptr, "Runaway exception", BitcoinGUI::tr("A fatal error occurred. Elements can no longer continue safely and will quit.") + QString("\n\n") + message);
     ::exit(EXIT_FAILURE);
 }
 
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -149,8 +149,7 @@ Intro::Intro(QWidget *parent, uint64_t blockchain_size, uint64_t chain_state_siz
     }
     requiredSpace += m_chain_state_size;
     ui->sizeWarningLabel->setText(
-        //ELEMENTS: the Bitcoin mention is removed here which breaks translations
-        tr("%1 will download and store a copy of the block chain.").arg(tr(PACKAGE_NAME)) + " " +
+        tr("%1 will download and store a copy of the %2 block chain.").arg(tr(PACKAGE_NAME)).arg(tr("Liquid")) + " " +
         storageRequiresMsg.arg(requiredSpace) + " " +
         tr("The wallet will also be stored in this directory.")
     );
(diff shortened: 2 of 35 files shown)


greenaddress:
utack 8e02d7ee6953d134af80b1f6ce25d340c61533eb

wintercooled:
Looks good to me

src/qt/bitcoin.cpp
```diff
@@ -390,7 +390,7 @@ void BitcoinApplication::shutdownResult()
 
 void BitcoinApplication::handleRunawayException(const QString &message)
 {
-    QMessageBox::critical(nullptr, "Runaway exception", BitcoinGUI::tr("A fatal error occurred. Bitcoin can no longer continue safely and will quit.") + QString("\n\n") + message);
+    QMessageBox::critical(nullptr, "Runaway exception", BitcoinGUI::tr("A fatal error occurred. Elements can no longer continue safely and will quit.") + QString("\n\n") + message);
```
philippem: This is a key that is used to index into the translation. If we change the string here without changing the <source> elements in the locale specific translations, the localized builds may fail to display a string.
For example:

https://github.com/ElementsProject/elements/blob/master/src/qt/locale/bitcoin_es.ts#L576

philippem:
utNACK
If we change a string key we need to at least change the same key in the locale files

src/qt/bitcoin.cpp
stevenroose: Yeah it seemed that that's what we have been doing in the past.

stevenroose:
@philippem  Hmm, I think we discussed this in a private convo, can you ACK that you agree? Since we never updated any of the translation files in the past and just left them for what they are, this PR is not the place to do that. If we want to support localization, we should have a more wide effort for porting all the missing strings.

Fyi, `rg -i elements src/qt/locale` shows nothing (except for one phrase that just has the common word elements in it in a single language).

stevenroose:
@philippem If you ACK this one, can you also ACK the backport: https://github.com/ElementsProject/elements/pull/848.

achow101:
The second commit is unnecessary. Qt will fall back to the text in the source code if it is unable to find the string in the translation files. Old strings in the translation files will be removed during a translation file update.

src/qt/intro.cpp
```diff
@@ -130,7 +130,7 @@ Intro::Intro(QWidget *parent, uint64_t blockchain_size, uint64_t chain_state_siz
         .arg(tr(PACKAGE_NAME))
         .arg(m_blockchain_size)
         .arg(2009)
```
philippem: this should be 2018 which is when the liquidv1 chain started

stevenroose:
> The second commit is unnecessary. Qt will fall back to the text in the source code if it is unable to find the string in the translation files. Old strings in the translation files will be removed during a translation file update.

@achow101 Well, I do a bunch of actual translation updates. (Where it's just changing the Bitcoin->Elements name.) So in those cases, it should be slightly better, right?

src/qt/intro.cpp
stevenroose: I updated it to 2016 (the start of Elements Alpha) in an extra commit.

achow101:
> @achow101 Well, I do a bunch of actual translation updates. (Where it's just changing the Bitcoin->Elements name.) So in those cases, it should be slightly better, right?

It won't affect anything and potentially makes the automated stuff break. I've found that manually tinkering with translation files is a good way to have a bad day.

stevenroose:
> It won't affect anything and potentially makes the automated stuff break. I've found that manually tinkering with translation files is a good way to have a bad day.

I don't understand what you mean here. You're saying that I should just not touch the translation files at all, even if it's trivial to change (both the `<source>` and `<translation>` parts)?

@philippem so I just remove that commit _again_?

achow101:
> I don't understand what you mean here. You're saying that I should just not touch the translation files at all, even if it's trivial to change (both the `<source>` and `<translation>` parts)?

Yes. If there is no plan to support translations, then there's no point in touching the files. If there is a plan to support translations, then automated scripts will update and clean up everything for you. There's no reason to change the translation files as any manual changes will get cleaned up by other processes. It just adds noise to the diff.

Issue #842, opened 2020-03-25
issuance blinding key vector size check in BlindTransaction: why num of issuances added ? 


dgpv:
https://github.com/ElementsProject/elements/blob/ac5a435247b10b74fbaab9817ce7d199675b3d29/src/blind.cpp#L229-L230

Whenever `issuance_blinding_privkey` or `token_blinding_privkey` accessed, the index will be `< tx.vin.size()`, but the assert at the start of the function checks that the size of these vectors is less than `tx.vin.size()+GetNumIssuances(CTransaction(tx)`. Shouldn't it be just `tx.vin.size()`?

stevenroose:
Hmm, good point. That's very old code.

I removed it in https://github.com/ElementsProject/elements/issues/842, I'm asking the people that originally wrote that code to review.

dgpv:
#846 is merged, the issue is resolved

PR #846, opened 2020-03-31
[RFC] Remove GetNumIssuance from BlindTransaction assert

stevenroose:
This is in response to @dgpv 's https://github.com/ElementsProject/elements/issues/842.

I'm not 100% confident this is correct, but it seems so. It also passed all tests on my machine.


apoelstra:
utACK. Yeah, this makes sense. The original code looks like an artifact of a very old design.

stevenroose:
Travis has been acting out since yesterday. I can't restart jobs. I pushed an empty commit to trigger a build, will remove it and merge once jobs succeed.

PR #847, opened 2020-03-31
[0.18 backport] Support fee estimation for fees lower than 1 sat/byte

stevenroose:
Backport of #843.


gwillen:
Per discussion I am assuming that 0.18.1.7 is also the latest version on the branch you're backporting to, as well as on master. Given that, utACK.

PR #849, opened 2020-04-01
init: add SystemD elementsd.service file

Sosthene00:
I set up a Bitcoin Core and an Elements node on a Ubuntu machine for testing, and noticed that there was no `elementsd.service` file available in Elements repo as I expected, but a `bitcoind.service` file that seems to be an outdated version of the file in the Bitcoin Core repository.

Since I needed to create a systemd service, I modified the most recent version of `bitcoind.service` file for elementsd as it seems fit to me.

Besides replacing all occurrences of `bitcoin` and `bitcoind`, I also moved the `debug.log` file to another dir and the path to the bin is also different (`/usr/local/bin/elementsd`). This might be too specific to my project though.

I suggest to add it to the Elements repository along the existing files, as maybe some people might find it useful.

I'm pretty rookie with systemd too, I tried to stick to the example and recommandation in the Bitcoin Core repository, but I'd be glad if someone had suggestions to improve this.


stevenroose:
Hi @Sosthene00! :)

It would be better if you renamed the bitcoind service file to elementsd in one commit and then made the necessary modifications in a second commit, so that we can clearly review what changes have been made and so that upstream changes will be reflected to that file in the future.

So please stick as closely as possible to the original file apart from of course renaming bitcoin to elements.

Thanks a lot for your contribution!

Sosthene00:
Ok thanks for the answer, I wasn't sure if you want to keep the file as it is today or update it to the last version in Bitcoin Core, so I made 2 commits and made clear in the comment which are the changes in the most recent version.

contrib/init/elementsd.service
```diff
@@ -1,47 +1,53 @@
...
-                            -conf=/etc/bitcoin/bitcoin.conf \
-                            -datadir=/var/lib/bitcoind
+ExecStart=/usr/bin/elementsd -daemon \
+                            -pid=/run/elementsd/elementsd.pid \
+                            -conf=/etc/elements/elements.conf \
+                            -datadir=/var/lib/elementsd
+
+# Make sure the config directory is readable by the service user
+PermissionsStartOnly=true
+ExecStartPre=/bin/chgrp elements /etc/elements
```
stevenroose, line 24: Ah I see that Bitcoin Core now has these as well.

stevenroose:
Cool that's a lot better!

Any chance you could clean up the commits? (Ideally one commit moving the file and another making the changes.)

If not, I can merge it as is.

Sosthene00:
Hi, I agree my commits are a bit messy I'll clean that up don't worry

Sosthene00:
Ok it looks like I messed up with this branch, so I created a new, clean PR #865 
I also rebased to current master, let me know if there's anything else I can do.

PR #855, opened 2020-04-07
Remove another mention of Liquid-Qt

stevenroose:
This should be the last mention of `#if LIQUID` and `Liquid-Qt` that my grep can find.


philippem:
ACK

PR #860, opened 2020-04-10
Icon fixes

allenpiscitello:
The icon used previously had text in it, that made it unreadable on most screens.  This has been replaced with just the graphic part of the icon.

In addition, Gitian builds had an issue where the icon was not formatted correctly. Changes were made to the icon to be more similar of an svg format than the previous one that worked.

Lastly, the changing of the icon color based on testnet was removed since Elements has many networks and it would not make sense to change for each.


src/Makefile.am
```diff
@@ -718,7 +718,7 @@ EXTRA_DIST += $(RES_RENDERED_ICON_SRC) $(RES_ALL_RENDERED_ICONS)
 if CAN_RENDER_ICONS
 
 qt/res/rendered_icons/about.png: qt/res/rendered_icons/about.svg
-	$(RSVG_CONVERT) -f png -d 142 -p 142 < $< | $(IMAGEMAGICK_CONVERT) - -crop 64x78+0+31 -bordercolor transparent -border 7x0 -colorspace Gray -strip $@
+	$(RSVG_CONVERT) -f png -d 142 -p 142 < $< | $(IMAGEMAGICK_CONVERT) - -bordercolor transparent -colorspace Gray -strip $@
```
allenpiscitello, line 721: Since the image is already the full image, no need to crop and add a border.

src/Makefile.am
```diff
@@ -739,14 +739,14 @@ qt/res/rendered_icons/bitcoin.ico: qt/res/rendered_icons/bitcoin32d8.gif $(patsu
 	$(IMAGEMAGICK_CONVERT) $^ $@
 
 qt/res/rendered_icons/bitcoin_testnet.ico: qt/res/rendered_icons/bitcoin.ico
-	$(IMAGEMAGICK_CONVERT) $^ -modulate 100,87,9 $@
+	$(IMAGEMAGICK_CONVERT) $^ $@
```
allenpiscitello, line 742: No modulating of color since this both looks bad and doesn't really apply.

src/Makefile.am
```diff
@@ -739,14 +739,14 @@ qt/res/rendered_icons/bitcoin.ico: qt/res/rendered_icons/bitcoin32d8.gif $(patsu
...
 
 qt/res/rendered_icons/nsis-wizard.svg: qt/res/src/bitcoin.svg
 	@$(MKDIR_P) $(@D)
 	sed '/fill="#000"/d' < $< > $@
 
 qt/res/rendered_icons/nsis-wizard.bmp: qt/res/rendered_icons/nsis-wizard.svg
-	$(RSVG_CONVERT) -f png -d 360 -p 360 < $< | $(IMAGEMAGICK_CONVERT) - -crop 164x290+0+35 -border 0x12 -strip BMP3:$@
+	$(RSVG_CONVERT) -f png -d 360 -p 360 < $< | $(IMAGEMAGICK_CONVERT) - -strip BMP3:$@
```
allenpiscitello, line 749: Removing cropping since the image is already cropped.

src/qt/splashscreen.cpp
```diff
@@ -65,7 +65,7 @@ SplashScreen::SplashScreen(interfaces::Node& node, Qt::WindowFlags f, const Netw
 
     // draw the bitcoin icon, expected size of PNG: 1024x1024
     int _32 = 32 / devicePixelRatio;
-    QRect rectIcon(QPoint(_32,-_32/2), QSize(pixmap.size().height()/devicePixelRatio+_32*2,pixmap.size().height()/devicePixelRatio+_32*2));
+    QRect rectIcon(QPoint(0, 0), QSize(pixmap.size().height()/(2*devicePixelRatio)+_32*2, pixmap.size().height()/(2 * devicePixelRatio)+_32*2));
```
allenpiscitello, line 68: Moving icon to upper left corner.

stevenroose:
I see you remove most of the cropping. Does this add more restrictions to the nature of the source file? 

I think it would be good to have some kind of reference on what properties the source file should have in case we ever want/need to change it again.

allenpiscitello:
> I see you remove most of the cropping. Does this add more restrictions to the nature of the source file?

Any change to the icon invalidates the code to crop things, other than very trivial changes such as colors or font, so if anything this simplifies it - it just assumes this icon will be the same in all places. The only place that may make sense to be different is in the splash screen.

> 
> I think it would be good to have some kind of reference on what properties the source file should have in case we ever want/need to change it again.

I'm not certain what exactly caused the issue, but a couple of things I noticed:
- the build file was looking for 'fill=' code, but many editors to svg files instead use css-style attributes, such as using classes or using style="fill:#xyz".  Both break the automated processes to try to create different style icons.  This is another reason why it would be more ideal to just have the actual icons we want to use, and let designers design in whatever tool works, instead of trying to rework things backwards.

I would suggest we remove all of the generation of icons from the build process in the future to avoid these issues.

philippem:
tACK

stevenroose:
ACK 45fb11978

Issue #853, opened 2020-04-07
rawissueasset with invalid address results in fee outputs


shesek:
It appears like `rawissueasset` does not reject invalid addresses provided in `{asset,token}_address`, but instead sends the asset and the reissuance tokens to fee outputs:

```bash
$ issuetx=$(ecli fundrawtransaction $(ecli createrawtransaction '[]' '{"'$(ecli getnewaddress)'":0.00001}') '{"feeRate":0.00005}' | jq -r .hex)
$ issuetx=$(ecli rawissueasset $issuetx '[{"asset_amount":1,"token_amount":3,"asset_address":"foobar","token_address":"foobar"}]' | jq -r '.[0].hex')
$ txid=$(ecli sendrawtransaction $(ecli signrawtransactionwithwallet $(ecli blindrawtransaction $issuetx) | jq -r .hex))
$ ecli getrawtransaction $txid true | jq -c .vout[]

{"value-minimum":1e-08,"value-maximum":687.19476736,"ct-exponent":0,"ct-bits":36,"surjectionproof":"030007eb1b7e2096f8648657d17f63e0894e1e37d2d7afaf750079ae474373f7359c358e73b64e533ffe56672fe588ffaf0ceaa06fee87cd0716b20ded198e45e123bb765d5ecbccd27fec05173056f3c34a18acd35212b50956b879a27ada79097e5eea4d3f6e996ca934d961b327f08796a4e42b47c317b5af3e6a013b6e2aad87bf","valuecommitment":"094862037cb3307c9827ef2a4dad1da97878b2d46a668c3de25b419d8cef9fc765","assetcommitment":"0b5ea2fef2352dcb86afe5b1016a65cc363ad3f9f65571763bacf34d903e0fd734","commitmentnonce":"02580ea6b7d071ae24c14ad49468b89f1acfea39a3b1cbe96a08f3af9e90beccda","commitmentnonce_fully_valid":true,"n":0,"scriptPubKey":{"asm":"OP_HASH160 e937a357dcf66c67480e3552863397035e8d5281 OP_EQUAL","hex":"a914e937a357dcf66c67480e3552863397035e8d528187","reqSigs":1,"type":"scripthash","addresses":["XYcNweAGbbWKmNZoCmaEqzcJmomgxbxUat"]}}
[...]

PR #861, opened 2020-04-13
Make issuance fail for invalid addresses 

stevenroose:
Fixes https://github.com/ElementsProject/elements/issues/853.

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -2808,13 +2808,10 @@ struct RawIssuanceDetails
 
 // Appends a single issuance to the first input that doesn't have one, and includes
 // a single output per asset type in shuffled positions.
-void issueasset_base(CMutableTransaction& mtx, RawIssuanceDetails& issuance_details, const CAmount asset_amount, const CAmount token_amount, const std::string& asset_address_str, const std::string& token_address_str, const bool blind_issuance, const uint256& contract_hash)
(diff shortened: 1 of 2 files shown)


stevenroose:
Please also review backport: https://github.com/ElementsProject/elements/pull/867.

src/rpc/rawtransaction.cpp
```diff
@@ -2808,13 +2808,10 @@ struct RawIssuanceDetails
...
 // a single output per asset type in shuffled positions.
-void issueasset_base(CMutableTransaction& mtx, RawIssuanceDetails& issuance_details, const CAmount asset_amount, const CAmount token_amount, const std::string& asset_address_str, const std::string& token_address_str, const bool blind_issuance, const uint256& contract_hash)
+void issueasset_base(CMutableTransaction& mtx, RawIssuanceDetails& issuance_details, const CAmount asset_amount, const CAmount token_amount, const CTxDestination& asset_dest, const CTxDestination& token_dest, const bool blind_issuance, const uint256& contract_hash)
 {
-
-    CTxDestination asset_address(DecodeDestination(asset_address_str));
-    CTxDestination token_address(DecodeDestination(token_address_str));
-    CScript asset_destination = GetScriptForDestination(asset_address);
-    CScript token_destination = GetScriptForDestination(token_address);
+    CScript asset_script = GetScriptForDestination(asset_dest);
```
gwillen, line 2813: You don't need to change this now if you don't feel like it, but in the future it makes review easier not to BOTH rename AND move code in the same commit.

gwillen:
utACK 4256916.

gwillen:
Looks like the test failure is a timeout, I assume you'll kick it to try again.

src/rpc/rawtransaction.cpp
stevenroose: Yeah I could have cut the change into more than one commit. I'd be happy to do that. But I see you ack'ed it.

PR #862, opened 2020-04-14
[WIP] added createrawtransaction rpc

tomt1664:
Inputs same format as createrawtransaction (an array of outpoints), but with that RPC you also supply a map of addresses to amounts and then another map of addresses to asset IDs (meaning you cannot have two outputs with the same address). createrawtxoutputs you specify all outputs explicitly (via the second array). 

Several use cases where users want to send different assets to the same address in the same transaction.


src/qt/qrc_bitcoin_build.qrc
```diff
@@ -0,0 +1,96 @@
+<!DOCTYPE RCC><RCC version="1.0">
```
dgpv, line 1: This file seems to be not relevant to the changes in RPC

test/functional/rpc_createrawtxoutputs.py
```diff
@@ -0,0 +1,72 @@
...
+        node2.sendrawtransaction(signtx["hex"])
+
+        self.sync_all()
+        node0.generate(1)
+        self.sync_all()
+
+        txdec = node2.getrawtransaction(sendtx,True)
+
+        assert txdec["confirmations"] == 1
+        assert len(txdec["vout"]) == 5
```
dgpv, line 68: could check assets and amounts, too

test/functional/rpc_createrawtxoutputs.py
```diff
@@ -0,0 +1,72 @@
...
+        inputs.append({"txid":issue2["txid"],"vout":vout2})
+        outputs.append({"address":addr1,"amount":10.0,"asset":issue1["asset"]})
+        outputs.append({"address":addr1,"amount":5.0,"asset":issue2["asset"]})
+        outputs.append({"address":addr2,"amount":4.99,"asset":issue2["asset"]})
+        outputs.append({"address":"fee","amount":0.01,"asset":issue2["asset"]})
+        outputs.append({"data":"deadbeef","amount":0.0,"asset":issue2["asset"]})
+
+        # create a transaction with same address outputs
+        node2.createrawtxoutputs(inputs,outputs)
+        node2.signrawtransaction(rawtx)
```
dgpv, line 58: I am most likely missing something, but where are `rawtx` and `signtx` are assigned ?

stevenroose:
Hmm, I agree that this way of inputting things is a lot better than the current `createrawtransaction` way. The latter is really hacked to be backwards compatible with Core.

We're discussing how we can get rid of the intrinsic 8-decimal Bitcoin notation and one of the potential solutions is to create alternatives for a lot of the RPC methods, so this might fit into that picture.

apoelstra:
I think a better approach would be to just change `ConstructTransaction` in rpc/rawtransaction_util.cpp so that the duplicate check checks (destination, asset) pairs instead of just destination. This would be a much smaller change and then would be consistent across `createrawtransaction` `sendmany`, `send`, etc

PR #865, opened 2020-04-14
Init: add elementsd.service file

Sosthene00:
(This is a cleaned up version of #849 )
I set up a Bitcoin Core and an Elements node on a Ubuntu machine for testing, and noticed that there was no elementsd.service file available in Elements repo as I expected, but a bitcoind.service file that seems to be an outdated version of the file in the Bitcoin Core repository.

Since I needed to create a systemd service, I modified the most recent version of bitcoind.service file for elementsd as it seems fit to me.

Besides replacing all occurrences of bitcoin and bitcoind, I also moved the debug.log file to another dir and the path to the bin is also different (/usr/local/bin/elementsd). This might be too specific to my project though.

I suggest to add it to the Elements repository along the existing files, as maybe some people might find it useful.

I'm pretty rookie with systemd too, I tried to stick to the example and recommandation in the Bitcoin Core repository, but I'd be glad if someone had suggestions to improve this.


stevenroose:
Thanks! ACK 536939e5e

Issue #868, opened 2020-04-17
Windows Setup installs into Bitcoin subfolder; icon says BitCoin Core


philippem:
When installing from the win setup built using gitian on tag `elements-0.18.1.7-rc1` on a Windows 10 64 bit (MSEdge-Win10-VMware VM), the default install folder is "C:\Program Files(x86)\Bitcoin". Expected folder would be "C:\Program Files(x86)\Elements". In addition, the dialog displays "BitCoin Core". See attached screenshots.

![image](https://user-images.githubusercontent.com/565792/79622266-6d60d380-80cb-11ea-8ab7-cb0d6000b098.png)

philippem:
This also happens with the win64-setup.

Setup files tested are `elements-0.18.1.7-win32-setup-unsigned.exe` and `elements-0.18.1.7-win64-setup-unsigned.exe`. 

![image](https://user-images.githubusercontent.com/565792/79622472-3f2fc380-80cc-11ea-9801-46f91a21ad17.png)

stevenroose:
Yeah so grepping for `Program Files` gives no result, so I'll have to dig a bit deeper for where this string is actually defined :/

dgpv:
https://github.com/ElementsProject/elements/blob/f0d97cdf70097de15a4aaf1321368a6a9f0f008a/share/setup.nsi.in#L53-L57

stevenroose:
Clever grepping. I found it as well, grepping for "installer" IIRC :)
See https://github.com/ElementsProject/elements/pull/869. But thanks!

PR #869, opened 2020-04-20
[need-test] Fix Bitcoin name usage in Windows installer

stevenroose:
Fixes https://github.com/ElementsProject/elements/issues/868.

I can't test by lack of Windows machine. @philippem how hard is it for you to run the Gitian build Windows only and test this? I replaced all the capital-B Bitcoin mentions in the file.

### DIFF
--- a/share/setup.nsi.in
+++ b/share/setup.nsi.in
@@ -51,7 +51,7 @@ Var StartMenuGroup
 !insertmacro MUI_LANGUAGE English
 
 # Installer attributes
-InstallDir $PROGRAMFILES64\Bitcoin
+InstallDir $PROGRAMFILES64\Elements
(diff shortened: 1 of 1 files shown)


philippem:
Yes I will test this, but the installation dir should be "Elements Core" not "Elements"

stevenroose:
Just rebased

psgreco:
We'll need to add this to the release notes, but we have to merge it eventually, and a major release seems like a good moment to do it.
The only undesired effect is that updating to a version with this patch leaves the old version still installed in ProgramFiles\Bitcoin, it doesn't seem that bad

ACK de75222ef70218e94cbdf80d280f30765bbe5702

PR #871, opened 2020-04-21
fixing windows installer icon

allenpiscitello:
The icon was improperly clipped and stretched during the last icon fixes. This makes it the correct size.


psgreco:
Code upstream has changed and doesn't apply anymore, we'll deal with this in another PR

Issue #885, opened 2020-06-17
Elements daemon crashes when rawreissue a non-funded raw transaction


louisinger:
**Config**: Regtest local Elements network. (See: https://github.com/louisinger/minimal-regtest-elements-network)

**issue**:  
Create a raw transaction with one input: a valid UTXO of a reissuance token:

`RTX=$(e1-cli createrawtransaction "[{\"txid\": \"3dd648b1b3b1f30c710a98c31966d3f3ea8774c6f9cfe6c32e4245215f10bf6e\", \"vout\": 1}]" '{"data":"1170bda35de764c0713f31c1aa8256bcd223441f7755f3c838688ed7c7a81f2d"}')`

Then, directly use this transaction inside `rawreissueasset`:

`e1-cli rawreissueasset $RTX "[{\"asset_amount\": 0.00000001, \"asset_address\": \"CTEq9QWdEgzhheGjVYaFDPxiJZVz8j2x2cf5z1VpyjG2TRAL3icqh1d3LDNM2EBrDfMVfMWQEZifNEGb\", \"input_index\": 0, \"asset_blinder\": \"7e8016819034f5e47810c1ae4844a20d115211a5f3d6194da85459b9bedd570f\", \"entropy\": \"fdb2b1bd8dc186d4f30d4eb51b6b35572adf6d04d500697dc969be0232c51ee9\"}]"`

The command crashes the daemon.

stevenroose:
I could reproduce this.

gwillen:
@louisinger I see that your input transaction doesn't appear to have any fee -- the fee in Elements transactions is an explicit output that has to be provided when constructing the transaction. Is that intentional? I think part of the reason you're hitting this issue is that the code is assuming (not unreasonably, although apparently unjustifiably here due to a missing check) that a fee output will be present, meaning that the reasonable minimum number of outputs is two.

louisinger:
@gwillen Yeah, that's intentional. I made the mistake of rigging the transaction. It surprised me that the daemon crashed. I saw the commit of jhfrontz, I close the issue.

PR #887, opened 2020-06-24
Handle degenerate case of a single output

jhfrontz:
Solves #885

`GetRandInt()` requires parameter be non-zero.  Otherwise, it returns a large random number-- this is what is happening in #885 (where the code was attempting to randomize the order of a single output).


jhfrontz:
Closed in lieu of https://github.com/ElementsProject/elements/pull/896

PR #888, opened 2020-06-29
build: Fix search for brew-installed BDB 4 on OS X (backport of upstream fix)

gwillen:
Backport of my upstream fix https://github.com/bitcoin/bitcoin/pull/19356 .

Net effect is to change the include order slightly for macOS homebrew users with BDB 4.8 installed privately (not linked into system directories) -- our configure scripts will prefer BDB 4.8 if they find it there, even if it is not system default. This fixes a build break that would occur in this situation when some _other_ BDB _is_ installed in system directories. The previous search order would use headers from system directories, but libs from 4.8, causing a crash.

----
On OS X, when searching Homebrew keg-only packages for BDB 4.8, if we find it,
use BDB_CPPFLAGS and BDB_LIBS instead of CFLAGS and LIBS for the result. This
is (1) more correct, and (2) necessary in order to give this location
priority over other directories in the include search path, which may include
system include directories with other versions of BDB.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -563,9 +563,10 @@ case $host in
            PKG_CONFIG_PATH="$openssl_prefix/lib/pkgconfig:$PKG_CONFIG_PATH"
            export PKG_CONFIG_PATH
          fi
-         if test x$bdb_prefix != x; then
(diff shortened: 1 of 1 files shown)


stevenroose:
ACK 011cea3e8

PR #889, opened 2020-06-30
Improve logging of failures during block signature validation

gwillen:
Pass in a ValidationState object to CheckProof and friends, so that we can
provide helpful error messages when validation fails.

@stevenroose This came out of the time I spent figuring out the problem we were seeing with Elements and Liquid regtest syncing.

### DIFF
--- a/src/block_proof.cpp
+++ b/src/block_proof.cpp
@@ -6,9 +6,12 @@
 #include <pow.h>
 
 #include <chain.h>
+#include <consensus/validation.h>
 #include <primitives/block.h>
 #include <script/interpreter.h>
 #include <script/generic.hpp>
+#include <util/system.h>
+#include <validation.h>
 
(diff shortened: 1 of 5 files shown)


gwillen:
Ugh, the circular dependency lint is really mad about this. @stevenroose, do you think this approach (extending the use of ValidationState into CheckProof) is worth trying to rescue, or I should just add some simple log messages and call it a day?

stevenroose:
It seems to me that the main differentiator it adds is "signature too large" vs "invalid signature" which could be valid information. IIRC when either of this happens, we ban the peer, right? So I'm not super worried about excessive logging if you'd just add another line indicating those 2 different scenarios..

gwillen:
@stevenroose This is kind of ancient, do you have an opinion on whether to close it or try to rescue it? I will default to closing.

PR #892, opened 2020-07-03
Change liquid PAK list for 0.18

gwillen:
Change Liquid PAK list and bump to 0.18.1.8.


apoelstra:
lol this needs to be forward-ported and brought into rc2.

Should also review all other 0.18 PRs

PR #896, opened 2020-07-15
Fix logic issues with randomization of outputs in [re]issueasset_base; minor refactoring.

### LINKED ISSUE #885: Elements daemon crashes when rawreissue a non-funded raw transaction

gwillen:
Fixes #885. This is a more comprehensive alternative to #887.

One note: The existing code claimed to assume (and mostly did assume) that fee outputs always go at the end. Other code does not necessarily appear to assume that. I have added explicit checks for it, so it should be obvious enough if it causes problems in practice (and with a small fix, the integration tests do pass.) If we're concerned about whether the assumption is valid, I'm willing to relax the assumption (but then the old code makes even less sense.)

@stevenroose for review? (@jhfrontz also, since #887 was yours so maybe you have opinions.)

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -2964,11 +2964,11 @@ UniValue rawissueasset(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have at least one output.");
     }
     if (!mtx.vout[mtx.vout.size() - 1].IsFee()) {
-        throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have exactly one fee output, which must be last");
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Last transaction output must be fee.");
     }
     for (size_t i = 0; i < mtx.vout.size() - 1; i++) {
         if (mtx.vout[i].IsFee()) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have exactly one fee output, which must be last");
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction can only have one fee output.");
         }
     }
 
@@ -3092,11 +3092,11 @@ UniValue rawreissueasset(const JSONRPCRequest& request)
 
     // Validate fee output location, required by the implementation of reissueasset_base
     if (!mtx.vout[mtx.vout.size() - 1].IsFee()) {
-        throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have exactly one fee output, which must be last");
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Last transaction output must be fee.");
     }
     for (size_t i = 0; i < mtx.vout.size() - 1; i++) {
         if (mtx.vout[i].IsFee()) {
-            throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have exactly one fee output, which must be last");
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction can only have one fee output.");
         }
     }
 
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -2807,9 +2807,13 @@ struct RawIssuanceDetails
 };
 
 // Appends a single issuance to the first input that doesn't have one, and includes
-// a single output per asset type in shuffled positions.
+// a single output per asset type in shuffled positions. Requires at least one output
+// to exist (the fee output, which must be last).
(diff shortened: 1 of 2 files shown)


src/rpc/rawtransaction.cpp
```diff
@@ -3076,9 +3090,17 @@ UniValue rawreissueasset(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have at least one output.");
     }
 
-    UniValue issuances = request.params[1].get_array();
+    // Validate fee output location, required by the implementation of reissueasset_base
+    if (!mtx.vout[mtx.vout.size() - 1].IsFee()) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Transaction must have exactly one fee output, which must be last");
```
jhfrontz, line 3095: Might it make any sense to differentiate this exception text from the next (e.g., substituting or adding "Last output is not a fee output" vs. "fee output is not the last output")?  I don't have enough context/background to know for sure, but it seems like giving the user additional/discriminating information might be helpful for debugging.

stevenroose: Hmm yeah I might agree with that. Exact same error messages for different pieces of code is often confusing. Even though I think Glenn carefully made sure the message should give clarity if you did either of the 2 wrong scenarios.

gwillen: Fair point, done.

stevenroose:
ACK 3386cae45bad5aecb4a5a9c501e81d8d9091d1f9

src/rpc/rawtransaction.cpp
```diff
@@ -3104,31 +3126,24 @@ UniValue rawreissueasset(const JSONRPCRequest& request)
             throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Invalid asset address provided: %s", asset_address_uni.get_str()));
         }
 
-        int input_index = -1;
+        size_t input_index = -1;
```
apoelstra, line 3129: what did you mean by this

gwillen: ... that is a good question. Let me look at that.

gwillen: I believe the answer is this:

- The initialization of `input_index` was irrelevant before, and is still irrelevant, because all codepaths either set it or throw an exception, so the initialization value is never used anywhere;
- I believe that what I did here (unintentionally) does have defined semantics, and will result in the value of `i` being the maximum representable `size_t` (irrelevant due to the above, except for the fact that it is not undefined behavior.)
- Making this unsigned also renders the `if (input_index < 0) {` below into dead code; I'm surprised there wasn't a warning about that.
- But it also means that negative values returned from `input_index_o.get_int();` should trip the following check `else if (input_index >= mtx.vin.size())`, and so should still fail.

So I think the resulting code is correct but slightly stupid.

What I was _thinking_ here is that I have a general preference for using `size_t` rather than `int` for array indices, but obviously that is stupid here if `input_index_o.get_int()` returns `int` and can return a negative value on failure.

So probably `input_index` should be changed back to `int` (or to `ssize_t` if we believe in those.)

apoelstra: There is a warning about the `< 0` in current gcc, FYI, which is how I noticed this.

Ok, cool, we can change it back to int in a followup PR. I'll leave the warning in place for the remainder of the rebase since we're almost done anyway.

PR #897, opened 2020-08-03
Documentation of Version 1 asset contracts (CBOR) format

gwillen:
This document describes:
* A brief history of Elements asset contracts
* A new format for user-supplied asset contracts, based on the RFC 7049 CBOR standard
* Some other tools and interfaces for dealing with assets and asset contracts

Left for another document which will come separately:
* Changes to the way asset values are handled in the Elements RPC interface, to
  cleanly handle assets with different levels of precision (i.e., number of digits after the
  decimal point)

### DIFF
--- a/doc/asset-contracts-cbor.md
+++ b/doc/asset-contracts-cbor.md
@@ -1,0 +1,189 @@
+# Asset Contracts Version 1 (CBOR format)
+
+## Overview
+
+This document describes:
+* A brief history of Elements asset contracts
+* A new format for user-supplied asset contracts, based on the RFC 7049 CBOR standard
+* Some other tools and interfaces for dealing with assets and asset contracts
+
+The document `asset-values-rpc.md` contains:
+* Changes to the way asset values are handled in the Elements RPC interface, to
+  cleanly handle assets with different levels of precision (i.e., number of digits after the
+  decimal point)
+
+## Glossary
+
+* Asset issuer: The party issuing the asset and writing the asset contract (and the software used for producing and registering the asset contract.)
+* Asset (contract) consumer: The party using the information contained in the asset contract (and the software used for this)
+* Issuance transaction: The transaction in which a new asset is initially defined.
+* Reissuance token: A special asset type, created (optionally) along with another asset, and which can be used to create new units of that asset.
+* Atom: The smallest unit of an asset (equivalent to a Satoshi in Bitcoin.)
+* Precision: The number of digits after the decimal point in the "decimal" representation of an asset value (e.g. 8 for Bitcoin).
+* Known asset: An asset for which a given consumer has seen/processed a valid version 1 contract.
+* Whitelisted asset: An asset for which a given consumer has the contract data from an out-of-band source.
+
+## History and motivation
+
+When creating an asset in Elements, some user-supplied data -- the "Ricardian contract" -- is
+included in the hash that defines the asset ID. This allows for an asset issuance transaction to
+commit to some external information that defines what the asset is.
+
+Although the Elements blockchain consensus (and standardness) rules do not enforce any structure
+to the user-supplied data, it was by convention
+the hash of a JSON document, with certain fields defining various aspects of the asset
+such as its name, issuer, ticker symbol, and precision. This was "version 0" of the asset contract
+format (indicated by '"version": 0' in the JSON object.) For example:
+
+```
+{
+    "entity": {
+        "domain": "store.blockstream.com"
+    },
+    "issuer_pubkey": "023c239fd39ae5fc8b88454fe36cae6a65a10c5b637a28dbcbc423d1e7f3bcc25e",
+    "name": "Hat",
+    "nonce": "171716",
+    "precision": 0,
+    "ticker": "HAT",
+    "version": 0
+}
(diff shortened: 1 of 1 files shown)


stevenroose:
ACK 63f70564bdedb59151aa6875896ad9190124269d

allenpiscitello:
ACK

doc/asset-contracts-cbor.md
```diff
@@ -0,0 +1,189 @@
...
+## Contract format
+
+A version 1 asset contract consists of: a single-byte version field (which must be 1), followed by a CBOR-encoded array with 3 elements:
+* The precision, encoded as an unsigned integer from 0 to 8.
+* The ticker, encoded as a string, 3-5 characters from the character set "[a-zA-Z.-]" (that is, upper and lowercase letters, dot, and dash.)
+* A map containing other fields, all of which are OPTIONAL:
+  * (The field names `"precision"`, `"ticker"`, and `"entity"` are reserved for backwards compatibility, and may not be used.)
+  * `"name"`, a string containing ASCII characters, length 1 to 255 (inclusive)
+  * `"issuer_pubkey"`, a byte-string containing a secp256k1 public key in compressed form.
+  * `"domain"`, a domain name belonging to the issuer
```
dgpv, line 117: Maybe also specify how the accompanying real-world contract can be found on this domain, with the hash of the text of this real-world contract to be used as 'contract hash' in the issuance ?

dgpv: This way wallets will be able to locate the real-world contract governing the off-chain conditions, verify that it is indeed referenced in the issuance, and display the contract to the user

gwillen: We mention the use of the 'domain' field a bit later under 'Registry rules' -- it's only used to prove the identity of the asset issuer. The use of "contract" here is not referring to a legal contract, but just to this structure with the parameters of the asset in it. (There is some discussion below of how these "contracts" are distributed.)

If the issuer wanted to add a reference to some other document, they would do that by adding another field, but it would be up to them to define the meaning of it.

doc/asset-contracts-cbor.md
```diff
@@ -0,0 +1,189 @@
...
+  * Consumers who trust and rely on an asset registry MAY assume that version 1 contracts downloaded from the registry are well-formed.
+  * Consumers who get their contracts from an untrusted source SHOULD refuse to use assets whose contracts are not well-formed, but are not required to do so.
+  * Consumers (such as hardware wallets) who wish to minimize parsing complexity may rely on the fact that a valid version 1 contract will always begin as follows: `0x01 0x83 <1 byte precision, value 0-8> 0x73-0x75 <3-5 byte ticker, ASCII text> ...`
+    * Of course, such consumers MUST NOT crash or display exploitable behavior if a contract received from an untrusted source turns out to be invalid.
+
+Registry rules:
+* A registry may impose additional restrictions on contracts, beyond those described above.
+* In particular: The Liquid Asset Registry requires the `name`, `issuer_pubkey`, and `domain` fields to be present, and the `domain` field must be controlled by the issuer, which must be proven by the existence of a specific file served over https from that domain.
+* The domain validation rule helps protect registry users against assets which are intended to spoof other assets.
```
dgpv, line 160: What is the domain validation rule exactly ?

doc/asset-contracts-cbor.md
dgpv: Why not define the standard way to do this, especially if there _is_ a need to pull some file from that domain to prove the identity of the issuer ?

dgpv: and if standard way is defined, then there's no need for adding another field

gwillen: Added the details for the way the Liquid Asset Registry does this right now.

doc/asset-contracts-cbor.md
gwillen: Added more info!

doc/asset-contracts-cbor.md
```diff
@@ -0,0 +1,194 @@
...
+    * For issuers who wish to make use of the onchain contract distribution feature in `OP_RETURN` outputs, attention should be paid to the size of the contract. For ancillary data which should be verifiable, but need not be distributed to all users in-band, it is recommended to store a hash of the data in a field of bytestring type, rather than the data itself.
+
+Further restrictions:
+* The entire contract, including the version byte and the entire CBOR structure, MUST be at most 256 bytes in length.
+  * This will be enforced as a standardness/policy rule, for contracts encoded in `OP_RETURN` outputs.
+  * Consumers SHOULD reject overlong contracts, but are not required to do so.
+* Some additional restrictions are imposed on the CBOR data, for security, ease of parsing, and (limited) JSON compatibility:
+  * The CBOR must be "well-formed" and "valid", and decoding must be done in "strict mode" if possible, as defined in RFC 7049:
+    * Maps MUST NOT have duplicate keys.
+    * Strings MUST be valid UTF-8.
```
jgriffiths, line 128: Domain names should be ASCII, punycode etc to mitigate https://en.wikipedia.org/wiki/IDN_homograph_attack

PR #898, opened 2020-08-09
Make 'librsvg2-2_2.40.13-3_amd64.deb' a required input to gitian build scripts

maaku:
Building elements on gitian requires an internet connection because the file 'librsvg2-2_2.40.13-3_amd64.deb' is fetched from the network.  This can be avoided if the file is listed as one of the required dependencies.  Be sure to put it in the 'inputs' directory prior to executing gbuild.

### DIFF
--- a/contrib/gitian-descriptors/gitian-linux.yml
+++ b/contrib/gitian-descriptors/gitian-linux.yml
@@ -37,7 +37,8 @@ packages:
 remotes:
 - "url": "https://github.com/ElementsProject/elements.git"
   "dir": "elements"
-files: []
+files:
+- "librsvg2-2_2.40.13-3_amd64.deb"
 script: |
 
   WRAP_DIR=$HOME/wrapped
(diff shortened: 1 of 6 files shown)


jhfrontz:
@maaku I'm confused -- how are you not needing an internet connection for the rest of the dependencies?  Maybe a local cache?

jhfrontz:
@maaku does this provide the internet independence you're looking for?  https://github.com/ElementsProject/elements/pull/899

maaku:
I don’t know. I assume there was a reason the gitian script was written to install this specific version? Does it impact reproducibility?

jhfrontz:
@maaku asks:
> I don’t know. I assume there was a reason the gitian script was written to install this specific version? Does it impact reproducibility?

That's just it-- the `wget` (over http, no less) looks suspect because gitian has infrastructure to request a specific version (or even a specific repo/URL). So I suspect it may have been added by someone who was unfamiliar with gitian.

apoelstra:
The stated reason for the specific version was to make icons render correctly, not anything to do with reproducibality.

I think this is supeceded by #899. Recommend close.

PR #899, opened 2020-08-10
Use gitian infrastructure to install librsvg2-2 (in lieu of hard-coded wget)

jhfrontz:
This removes the hardcoded `wget` of libsrvg2-2 package from the internal gitian `script` and instead uses the `packages` capability (the `wget`-based implementation causes unintended dependencies on internet infrastructure; it is also brittle).

### DIFF
--- a/contrib/gitian-descriptors/gitian-linux.yml
+++ b/contrib/gitian-descriptors/gitian-linux.yml
@@ -25,6 +25,7 @@ packages:
 - "git"
 - "icnsutils"
 - "imagemagick"
+- "librsvg2-2"
 - "librsvg2-bin"
 - "pkg-config"
 - "autoconf"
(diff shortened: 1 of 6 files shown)


stevenroose:
Yeah so I am thinking the reason @luke-jr added this specific version `2.40.13-3` must have been because the then-default version didn't render icons correctly. So an actual testing of all the icons (@philippem ?) in the OSX, Windows and Linux binaries and installers will be necessary.

maaku:
Or at least merge #898 which solve the problem of requiring internet connectivity for a build without changing the installed version of librsvg2.

jhfrontz:
@maaku if that *exact* version (vs. just 2-2) is required, there is a gitian way to do that.  But first let's check if the latest is fine.

PR #900, opened 2020-08-13
Redesign PSET

achow101:
The original PSET had some issues with leaking private information and there were issues with the design of the various roles. This PR replaces it with a new PSET design that should be able to provide all of the things that people expect PSETs to do while preserving privacy.

Additionally, an actual specification for this new design has been written up. It describes all of the new fields, their serialization, and how roles should handle the PSET data.

This is still a work in progress as not everything with the blinder role has been worked out yet.


doc/pset.mediawiki
```diff
@@ -332,24 +457,69 @@ Using the transaction format involves many different responsibilities. Multiple
...
+
+For the Blinder's outputs that are to be blinded (i.e. they have a blinding pubkey), the Blinder will create value and asset commitments and put them in their respective fields.
+When they do so, the <tt>PSET_OUT_VALUE</tt> and <tt>PSET_OUT_ASSET</tt> fields must be removed.
+The Blinder will create the Value Rangeproof and Asset Surjection Proof and put them in their respective fields.
+It will also add the ephemeral pubkey used for ECDH of the nonce for the rangeproof to the <tt>PSET_OUT_ECDH_PUBKEY</tt> field.
+The blinder will then subtract the sum of their inputs' blinding factors from the sum of their output's blinding factors.
+The result must be added as a <tt>PSET_GLOBAL_SCALAR</tt> field.
+
+For the Blinder who blinds the last output to be blinded, it will subtract all of the scalar offsets from the value blinding factor for the last output.
+The result will be the amount blinding factor for that last output. The creation of the asset commitment, proofs, and other fields proceeds as usual.
```
jgriffiths, line 507: s/amount blinding factor/value blinding factor/

doc/pset.mediawiki
```diff
@@ -21,6 +20,9 @@ signatures for an input while the input does not have a complete set of signatur
 The signer can be offline as all necessary information will be provided in the
 transaction.
 
+This is a modification of BIP 174 for the Elements sidechain. It includes changes necessary
+for Confidential Transactions, Confidential Assets, and Peg-in transactions.
+
```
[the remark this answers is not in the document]
achow101: Commit 0b3b9edcfe5bcf74c1239b073e00a1ea3f788682 is just the PSET changes.

doc/pset.mediawiki
```diff
@@ -138,6 +140,17 @@ The currently defined global types are as follows:
...
+Note that previous implementations of PSET used a different prefix string and had different fields.
+The use of a new identifier prefix is to avoid conflicts with the previous implementations.
+
+The currently defined elements global proprietary types are as follows:
+
+* Type: Scalar Offset <tt>PSBT_GLOBAL_SCALAR = 0x00</tt>
+** Key: A 32 Scalar Offset computed by a blinder.
+*** <tt>{0x00}|{offset}</tt>
+** Value: None. The value must have a length of zero.
+
```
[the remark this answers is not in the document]
achow101: There can be multiple scalars. Putting it in the value then requires a unique key per scalar.

The original suggestion was to attach a scalar to an output, however they aren't really related to a particular output, so I opted to go with a global field.

doc/pset.mediawiki
```diff
@@ -332,24 +457,69 @@ Using the transaction format involves many different responsibilities. Multiple
...
+Such an output can be a 0 value OP_RETURN output.
+
+The Creator decides whether and which inputs are issuance inputs.
+For those inputs, the issuance data must be added to the input in the global unsigned transaction, except for the issuance amount.
+Such inputs must also have the issuance flag set in the outpoint.
+This means that the asset blinding nonce, asset entropy, and asset inflation keys need to be set, while the asset issuance amount is 0 for now.
+The amount is added to the PSET input as <tt>PSET_IN_ISSUANCE_VALUE</tt>.
+
+The Creator decides whether and which inputs are Peg-in inputs.
+Peg-in inputs must have the Peg-in flag set in their outpoints.
```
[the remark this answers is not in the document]
achow101: I am under the impression that they are not mutually exclusive. I don't really see why they would be.

doc/pset.mediawiki
```diff
@@ -206,6 +219,68 @@ The currently defined per-input types are defined as follows:
...
+* Type: Issuance Value <tt>PSET_IN_ISSUANCE_VALUE = 0x00</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x00}</tt>
+** Value: The explicit little endian 64-bit integer for the value of this issuance.
+*** <tt>{value}</tt>
+
+* Type: Issuance Value Commitment <tt>PSET_IN_ISSUANCE_VALUE_COMMITMENT = 0x01</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x01}</tt>
+** Value: The 33 byte Value Commitment. Adding this field requires the removal of <tt>PSBT_IN_ISSUANCE_VALUE</tt> from the PSET
```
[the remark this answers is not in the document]
achow101: Done.

doc/pset.mediawiki
```diff
@@ -332,24 +457,69 @@ Using the transaction format involves many different responsibilities. Multiple
...
 ===Updater===
 
-The Updater must only accept a PSBT.
-The Updater adds information to the PSBT that it has access to. If it has the UTXO for an input, it should add it to the PSBT.
+The Updater must only accept a PSET.
+The Updater adds information to the PSET that it has access to. If it has the UTXO for an input, it should add it to the PSET.
+Such UTXOs must contain all commitments, rangeproofs, and surjection proofs if they are present.
 The Updater should also add redeemScripts, witnessScripts, and BIP 32 derivation paths to the input and output data if it knows them.
 
+For Peg-in inputs, if available, the updaater should add the Peg-in transaction, the txout proof, the genesis hash, and the claim script.
```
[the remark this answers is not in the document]
achow101: Fixed.

doc/pset.mediawiki
achow101: Done

doc/pset.mediawiki
```diff
@@ -304,10 +435,10 @@ the key.
...
-will still need to handle events where keys are duplicated when combining transactions with duplicated fields. In this event, the software may choose
+will still need to handle events where keys are duplicated when combining transactions with duplicated fields. In this event, with the exception of duplicate output commitments and proofs, the software may choose
 whichever value it wishes.<ref>'''Why can the values be arbitrarily chosen?''' When there are duplicated keys, the values that can be chosen will either be
 valid or invalid. If the values are invalid, a signer would simply produce an invalid signature and the final transaction itself would be invalid. If the
-values are valid, then it does not matter which is chosen as either way the transaction is still valid.</ref>
+values are valid, then it does not matter which is chosen as either way the transaction is still valid.</ref>. For <tt>PSBT_OUT_VALUE_COMMITMENT</tt>, <tt>PSBT_OUT_ASSET_COMMITMENT</tt>, <tt>PSBT_OUT_VALUE_RANGEPROOF</tt>, <tt>PSBT_OUT_ASSET_SURJ_PROOF</tt>, and <tt>PSBT_OUT_ECDH_PUBKEY</tt>, the PSET should be considered invalid if these are duplicated.
```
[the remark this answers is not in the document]
achow101: Simplified to "All PSET outputs"

doc/pset.mediawiki
```diff
@@ -234,6 +309,62 @@ determine which outputs are change outputs and verify that the change is returni
...
+** Value: The 33 byte blinding pubkey to be used when blinding this output.
+*** <tt>{blinding pubkey}</tt>
+
+* Type: Ephemeral ECDH Pubkey <tt>PSET_OUT_ECDH_PUBKEY = 0x07</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x07}</tt>
+** Value: The 33 byte ephemeral pubkey used for ECDH in the blinding of this output.
+*** <tt>{ephemeral pubkey}</tt>
+
+* Type: Blinder Index <tt>PSET_OUT_BLINDER_INDEX</tt>
```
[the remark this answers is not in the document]
achow101: Fixed.

doc/pset.mediawiki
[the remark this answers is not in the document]
jgriffiths: Ah. This wasn't clear from the docs. given multiple scalars in a psbt, how do i know in which context to use one versus another? It seems like more data than just the scalar value will be needed?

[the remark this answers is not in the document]
achow101: The scalars are only used by the last blinder and they are used all together at that time. They're just used in the last blinding step to make sure that everything balances.

src/psbt.cpp
```diff
@@ -38,8 +42,23 @@ bool PartiallySignedTransaction::Merge(const PartiallySignedTransaction& psbt)
 
 bool PartiallySignedTransaction::IsSane() const
 {
-    for (PSBTInput input : inputs) {
+    for (unsigned int i = 0; i < inputs.size(); ++i) {
+        const PSBTInput& input = inputs[i];
         if (!input.IsSane()) return false;
+        // Check issuance is empty in CTxOut but set in input
```
dgpv, line 48: This comment is confusing. The "Check issuance is empty in CTxOut" happens later, but it is referred to as it it happens just here. I think better wording would be along the lines of "Check that issuance is empty in the input, because later we check that issuance is empty in CTxOut"

src/psbt.cpp
```diff
@@ -38,8 +42,23 @@ bool PartiallySignedTransaction::Merge(const PartiallySignedTransaction& psbt)
...
+        if (!tx->vin[i].assetIssuance.IsNull()) {
+            if (!tx->vin[i].assetIssuance.nAmount.IsNull()) return false;
+            if (!input.issuance_value && input.issuance_value_commitment.IsNull()) return false;
+        }
+        // Check whether any input is blinded
```
dgpv, line 53: The comment is at the end of the loop. Usually the comment describes the code that is below (or left of) the comment. Does this comment refer to the code that is supposed to be here, but removed ?

src/psbt.cpp
```diff
@@ -209,23 +228,77 @@ bool PSBTOutput::IsNull() const
...
-    if (nonce_commitment.IsNull() && !output.nonce_commitment.IsNull()) nonce_commitment = output.nonce_commitment;
     if (range_proof.empty() && !output.range_proof.empty()) range_proof = output.range_proof;
     if (surjection_proof.empty() && !output.surjection_proof.empty()) surjection_proof = output.surjection_proof;
+
+    return true;
+}
+
+bool PSBTOutput::IsBlinded() const
+{
+    return blinding_pubkey.IsValid();
```
dgpv, line 277: Nor `IsPartiallyBlinded` neither `IsFullyBlinded` check `blinding_pubkey.IsValid()`. That means that there is a logical contradiction that an output can be 'FullyBlinded', but not 'Blinded'. This can create confusion that can result in mistakes. Maybe better wording can help, like 'Blindable' instead of 'Blinded' ? Or `IsPartiallyBlinded`/`IsFullyBlinded` should imply `IsBlinded`

src/psbt.cpp
```diff
@@ -329,6 +402,7 @@ bool FinalizeAndExtractPSBT(PartiallySignedTransaction& psbtx, CMutableTransacti
         result.witness.vtxinwit[i].scriptWitness = psbtx.inputs[i].final_script_witness;
         PSBTInput& input = psbtx.inputs[i];
 
+        /*
```
dgpv, line 405: Commented-out (and not removed) code without any further comment with justification of not removing it can be confusing and distracting. Is this commented out just temporary ?

src/psbt.h
```diff
@@ -174,49 +159,83 @@ struct PSBTInput
             SerializeToVector(s, final_script_witness.stack);
         }
 
+        // Issuance value
+        // We shouldn't have both the value and value commitment, but maybe we do
```
dgpv, line 163: The comments should make the code more clear, rather than introduce doubt.

"but maybe we do" is not connected to the action in the code, and should either be dropped, or expanded, "but maybe we do, and in this case we ignore the value"

src/psbt.h
```diff
@@ -528,47 +574,56 @@ struct PSBTOutput
...
+            SerializeToVector(s, PSBT_PROPRIETARY, PSBT_ELEMENTS_ID, PSET_OUT_VALUE);
+            SerializeToVector(s, *value);
+        }
 
-            if (!asset_commitment.IsNull()) {
-                SerializeToVector(s, PSBT_OUT_PROPRIETARY, PSBT_ELEMENTS_ID, PSBT_OUT_ASSET_COMMITMENT);
-                SerializeToVector(s, asset_commitment);
-            }
+        // Asset
+        // We shouldn't have both asset and asset commitment, but if we do, wrte only the asset commitment
```
dgpv, line 590: mistype: "wrte" instead of "write"

doc/pset.mediawiki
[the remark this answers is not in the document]
dgpv: The people who will be implementing support for PSET in their code have to use the full document, they cannot use commit as the reference. I concur with @jgriffiths that the doc should be an addendum rather than edit. I think that the described problems (need to read diff PSET/PSBT instead of the document, need to sync with PSBT as it is updated) are significant. It is the additions to the standard that are proprietary, not the standard itself. Therefore the additions should be described in the spec.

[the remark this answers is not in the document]
dgpv: *proprietary in a sense that they use the `PSBT_PROPRIETARY ` type designed for proprietary extensions, not that the changes are somehow not open.

src/psbt.cpp
achow101: Yes, it is just temporary. The comment is to just make the code compile so that the action that this code does can be reimplemented later in this PR.

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+* Type: Value <tt>PSET_OUT_VALUE = 0x00</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x00}</tt>
+** Value: The explicit little endian 64-bit integer value for this output.
+*** <tt>{value}</tt>
+
+* Type: Value Commitment <tt>PSET_OUT_VALUE_COMMITMENT = 0x01</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x01}</tt>
+** Value: The 33 byte Value Commitment for this output. Adding this field requires the removal of <tt>PSET_OUT_VALUE</tt>.
```
dgpv, line 323: Adding PSET_OUT_VALUE requires remove of commitment, too. I think better say that the fields are mutually exclusive, as said for issuance value/commitment fields. (the same can be said for asset and asset commitment)

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+key that has no data except for the type specifier must only have the type specifier in
+the key.
+
+===Handling Duplicated Keys===
+
+Keys within each scope should never be duplicated; all keys in the format are unique. PSBTs containing duplicate keys are invalid. However implementors
+will still need to handle events where keys are duplicated when combining transactions with duplicated fields. In this event, with the exception of duplicate output commitments and proofs, the software may choose
+whichever value it wishes.<ref>'''Why can the values be arbitrarily chosen?''' When there are duplicated keys, the values that can be chosen will either be
+valid or invalid. If the values are invalid, a signer would simply produce an invalid signature and the final transaction itself would be invalid. If the
+values are valid, then it does not matter which is chosen as either way the transaction is still valid.</ref>. For <tt>PSBT_OUT_VALUE_COMMITMENT</tt>, <tt>PSBT_OUT_ASSET_COMMITMENT</tt>, <tt>PSBT_OUT_VALUE_RANGEPROOF</tt>, <tt>PSBT_OUT_ASSET_SURJ_PROOF</tt>, and <tt>PSBT_OUT_ECDH_PUBKEY</tt>, the PSET should be considered invalid if these are duplicated.
```
dgpv, line 441: I think what is meant here is that 'the combined PSET should be considered invalid'. Wouldn't it be more clear to say that PSETs cannot be combined if these fields are present in both instances ?

Also I believe that if values of these fields are equal for the outputs of two PSETs, they can be combined. For example, two PSETs that was modified from one common ancestor could have these fields in outputs that are equal.

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+
+The Creator creates a new PSET. It must create an unsigned transaction and place it in the PSET.
+
+For all outputs, the Creator must set the values and assets in the unsigned transaction to 0 or NULL.
+It will then attach the output value and asset type as explicit values in the <tt>PSET_OUT_VALUE</tt> and <tt>PSET_OUT_ASSET</tt> fields respectively.
+
+For any output which is going to be blinded, the Creator must add the blinding pubkey in the <tt>PSET_OUT_BLINDING_PUBKEY</tt> field.
+
+If any input is blinded, at least one of the outputs must also be blinded, i.e. it must have a <tt>PSET_OUT_BLINDING_PUBKEY</tt>.
+Such an output can be a 0 value OP_RETURN output.
```
dgpv, line 474: Creator can be unaware that Updater will add a blinded input afterwards. Then the updater have to enforce this rule, too.

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+
+The Creator decides whether and which inputs are issuance inputs.
+For those inputs, the issuance data must be added to the input in the global unsigned transaction, except for the issuance amount.
+Such inputs must also have the issuance flag set in the outpoint.
```
dgpv, line 478: I would add "in the outpoint of the prevout". Just "outpoint" can be mistakenly read as "output" and confuse the reader (I myself misread and was confused, like "how an output can be related here at all ?")

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x08}</tt>
+** Value: The little endian 64-bit value of the peg-in for the Peg-in Transaction.
+*** <tt>{value}</tt>
+
+* Type: Peg-in Witness <tt>PSET_IN_PEG_IN_WITNESS = 0x09</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x09}</tt>
+** Value: The Peg-in witness for the Peg-in Transaction.
+*** <tt>{script}</tt>
```
dgpv, line 282: There should be `PSET_IN_ASSET_COMMITMENT` to supply asset commitment of the prevout, so Creator/Updater can provide blinding information for the inputs when they will not be the Blinder

dgpv: (this info is supplied in `auxiliary_generators` argument to `BlindTransaction()`

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+The Blinder can identify which outputs it should blind by checking the <tt>PSET_OUT_BLINDER_INDEX</tt>.
+Using that index, the Blinder should inspect the input at that same index.
+If that input belongs to the Blinder, then the output is to be blinded by it.
+
+For the Blinder who blinds the last output to be blinded, it will subtract all of the scalar offsets from the value blinding factor for the last output.
+The result will be the value blinding factor for that last output. The creation of the asset commitment, proofs, and other fields proceeds as usual.
+Once all outputs are blinded, all <tt>PSET_GLOBAL_SCALAR</tt> fields must be removed from the PSET.
+
+A single entity is likely to be a Creator, Updater, and Blinder.
+In that case, the PSET should never be output with <tt>PSET_IN_ISSUANCE_VALUE</tt>, <tt>PSET_OUT_VALUE</tt>, or <tt>PSET_OUT_ASSET</tt>.
```
dgpv, line 521: Creator must set output values to 0 and assets to NULL. So for unblinded outputs, `PSET_OUT_VALUE` and `PSET_OUT_ASSET` need to remain after blinding. Or there should be a rule that unblinded output values should have their value/asset specified in embedded transaction, and these `PSET_*` fields used only for blindable outputs.

doc/pset.mediawiki
```diff
@@ -0,0 +1,803 @@
...
+Change outputs would then be identified as being the outputs which have the same script policy as the inputs that are being signed.
+
+===Combiner===
+
+The Combiner can accept 1 or many PSBTs.
+The Combiner must merge them into one PSBT (if possible), or fail.
+The resulting PSBT must contain all of the key-value pairs from each of the PSBTs.
+The Combiner must remove any duplicate key-value pairs, in accordance with the specification. It can pick arbitrarily when conflicts occur.
+For all PSET output fields, the combiner must fail to combine if these fields are not identical.
+If the PSETs to be combined have incomplete blinded outputs but the combined result would only complete blinded outputs, then the Combiner must fail unless it is also a Blinder and can reblind its outputs.
```
dgpv, line 620: This wording is unclear. Does it says that if the two outputs that are combined are both incomplete-blinded, but the resulting output will have all the fields to consider it complete, there is a big chance that these fields are incompatible because different randomness was used to produce them ?

doc/pset.mediawiki
```diff
@@ -234,6 +309,62 @@ determine which outputs are change outputs and verify that the change is returni
...
+** Value: The 33 byte Asset Commitment for this output. Adding this field requries the removal of <tt>PSET_OUT_ASSET</tt>.
+*** <tt>{asset commitment}</tt>
+
+* Type: Value Rangeproof <tt>PSET_OUT_VALUE_RANGEPROOF = 0x04</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x04}</tt>
+** Value: The rangeproof for the value of this output.
+*** <tt>{rangeproof}</tt>
+
+* Type: Asset Surjection Proof <tt>PSET_OUT_ASSET_SURJ_PROOF = 0x05</tt>
```
dgpv, line 344: if `RANGEPROOF` is spelled without underscore, it would be more consistent to spell `SURJPROOF` or `SURJECTIONPROOF` also without underscore. Or use `RANGE_PROOF` instead.

achow101:
I've updated pset.mediawiki to just be PSET and not contain PSBT.

dgpv:
I'd like to note that I find PSET prefix confusing if used in implementation (not so much in the specification) that will most likely be based on PSBT. Unless BIP174 is used in prefix for the names rather than PSBT, I would say that it is better to use `PSBT_Elements`, `PSBT_ELTS` and the like prefixes for the names, and say that 'PSBT' stands for 'partially-signed bitcoin-like transaction'.

dgpv:
Especially when used along with `PSBT_`-prefixed names, `PSET_` will cause frequent paint dues to mixups, especially because `B` and `E` are visually close.

doc/pset.mediawiki
dgpv: I personally prefer `PSET_OUT_VALUE_RANGE_PROOF` and `PSET_OUT_ASSET_SURJECTION_PROOF`

dgpv: Although in the Elements code they are referred to as `rangeproof` and `surjectionproof`, without underscores.

dgpv: Also 'value' and 'asset' prefixes can be implicit in these names

doc/pset.mediawiki
achow101: That would violate the constraint that the unsigned tx cannot change. Maybe we should assume (or stipulate) that creators know which/whether inputs are blinded?

doc/pset.mediawiki
achow101: This should be in the UTXO already.

doc/pset.mediawiki
achow101: Yes.

doc/pset.mediawiki
achow101: In the various papers on CT and CA, `rangeproof` as one word is the typical usage while `surjection proof` is 2 words. So following that convention, it would be `RANGEPROOF` and `SURJECTION_PROOF`.

doc/pset.mediawiki
achow101: I clarified this to say that unblinded outputs should have these fields.

doc/pset.mediawiki
achow101: Done

doc/pset.mediawiki
achow101: I've tried to clarify and generalize this.

doc/pset.mediawiki
achow101: Done

src/psbt.h
achow101: Fixed

src/psbt.h
achow101: Clarified

src/psbt.cpp
achow101: Enforced it.

src/psbt.cpp
achow101: Removed

src/psbt.cpp
achow101: Reworded

dgpv:
> In the various papers on CT and CA, rangeproof as one word is the typical usage while surjection proof is 2 words. So following that convention, it would be RANGEPROOF and SURJECTION_PROOF.

Implementors are more likely to look into reference code than the papers, and more likely to get the conventions from there. Elements code mainly uses no-underscore convention.

```
elements$ git grep surjectionproof|wc -l
226
elements$ git grep surjection_proof|wc -l
24
elements$ git grep rangeproof|wc -l
272
elements$ git grep range_proof|wc -l
16
```

doc/pset.mediawiki
```diff
@@ -0,0 +1,289 @@
...
+* Type: Peg-in Witness <tt>PSBT_ELEMENTS_IN_PEG_IN_WITNESS = 0x09</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x09}</tt>
+** Value: The Peg-in witness for the Peg-in Transaction.
+*** <tt>{script}</tt>
+
+* Type: Issuance Inflation Keys Amount <tt>PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS</tt>
+** Key: None. The key must only contain the 1 byte type.
+*** <tt>{0x0a}</tt>
+** Value: The number of inflation keys to set in this issuance. This is mutually exclusive with <tt>PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS</tt>.
```
sanket1729, line 114: Should be exclusive to `PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS_COMMITMENT` instead of `PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS`

achow101: whoops yes.

achow101:
This has been reimplemented to be based on PSBTv2. Also rebased onto master.

stepansnigirev:
Running `walletcreatefundedpsbt` always uses non-confidential change address, even if I set `changeAddress` to confidential in options (blinding pubkey is not added to the change psbt output).

achow101:
> Looks like several compilation failures in CI.

Hopefully those have been fixed

>     * Running `walletcreatefundedpsbt` always uses non-confidential change address, even if I set `changeAddress` to confidential in options (blinding pubkey is not added to the change psbt output).
> 
>     * `combinepsbts` crashes `elementsd`

Both should be fixed.

stepansnigirev:
A few questions regarding hardware wallet integration:
- Is there a way to get psbt with asset blinding factor and value blinding factor from the RPC (and ideally with the rangeproof nonce)? Otherwise HW can't verify what amount is sent to which address. Maybe we could add an option that allows to keep the blinding factors in the psbt?
- How can I sign psbt with sighash `ALL | RANGEPROOF`? Looks like only normal bitcoin sighashes are supported by `walletprocesspsbt` command

stepansnigirev:
This PSET crashes `elementsd` (it might be not a valid PSET though):

```
cHNldP8BAgQCAAAAAQMEAAAAAAEEAQEBBQEDAfsEAgAAAAABAP1UAQIAAAAAAQWsi8gzbkpIFg1JGOfwbwHB+zwTvDqdu62PxxRlZTEbAAAAAAD9////Aws+xB40+bU6wxMzsaHj2AyfBGtDgfJxv2qg1MQNeeFo9Aj6EetUwJqXyQMHU9hntovlxLe/I7o9+hKplEc5mk7AmANU7QZl+xlHN2aK8pkizkqRFOZ7qEjdeIuZSOh5RakfFRYAFORruriE5312Qp8/LMr61EJs5lPdC7G7TKYKP9DFrt4NwW88DNK3GacbU8+dJdSbXUT9HKciCAswmY1J6UgHkmijiuEAZnnfL8LrvW1CgqkQysTizqoNAlAIkSvDDhYGo3Mn2d5KeliK1L+3YCBrco6kqPqAtDhmFgAUH3AHqFUpakDGc4Zhvtn3VLoH/HcBTAKfSA9a0YQ+mUPM+JelqJGC43Gq+3scLUjjPPuNC4oBAAAAAAAABV4AAAAAAAABAXoLPsQeNPm1OsMTM7Gh49gMnwRrQ4Hycb9qoNTEDXnhaPQI+hHrVMCal8kDB1PYZ7aL5cS3vyO6PfoSqZRHOZpOwJgDVO0GZfsZRzdmivKZIs5KkRTme6hI3XiLmUjoeUWpHxUWABTka7q4hOd9dkKfPyzK+tRCbOZT3SIGAp72ZIN+Q3KR138cZ7FK5t0DWT1R60J4cZSKrU5x0fzHGGbQm7VUAACAAQAAgAAAAIABAAAACAAAAAEOIC0BA7G+v4yC4HdoIf40hP8ryuKuCQYyDOXFG5GqWYpRAQ8EAAAAAAEQBP3///8H/ARwc2V0Dv1OEGAzAAAAAAAAAAFtNcYBAUm9N+ZAvpXinP35edvTxRYr1LMugUTWB63StcsPXob+1KKapUdivDq7E9tR0DmT5xt8IzufjJqG8yWTU5N/IurdBPslkesixTFpg1ox6tgXsDDkeWkuyPJobFmY4R7vJ6JIVSV9bVudIYOHGrDwCGZx7aRxbqWw8XNpgdaTA/zBZRIMKpjyOh6jgI1KPDImGMkS6OfoEKSV08mW3sbT6EsA+2ZXoVe4l3I7APB5LR6vtGzllWNPVsKBD1RIpuM1EuO5xF3+IpNW4ILQkrKNwCyX/nP/nZ+gwHTEJsPcNWheXRzc6NbDB1O5zF02VIYd0bvClE9gc7QJphJbbjiMpCbroFMqnePZxXUtd2TNrsUZNRslf5mYGA8+Y3qNymV0uZ/WNZcUM4kya6UMxB0cqe9tnlngjDj1HjUJ+kyL6hez45ndYHfPRCyeQ4JvuyWfvtzU2LmYWrAU/MqxNNGTPRspVa76/ChOpawHKjihz0EZtQvjSkgal4EBiFL5gSd8v9o5swevhNFcNWyiv7gqLap9WTK3nTjzKSM0MGtBBUUMsAbI1izMFQdcSAuB9F9wZsQuZTdG/SykpoL6ecsKOhgnycAzhLkPuE8GeehT1CHGyj2iWqXNGIMHiy9BvjXJW/K1ChMnZ089f5/Z7JDYg43cPbB7JLBXVcrX4QyvR6yWaGSNZlh9qYcKd2oey37mUVdEqPfCU+Y75Gz8KyuFuFq42vecOwFYeRNNCjcAT9kuRq1OV/r/u7PvaaYM2iTsMwxozv1oMGtDcR3sk4Hd/eg0bNG8ARP1V4/ZxzJBQZGGQI4n6pCJL4GDdQfcdbZ9SO2cOVjEI9Uth0oVY0uXZ9IynqnFpy3s2RAHrqjoUlAPdYZCdcs+SpZpb06F5eHHybCpB5ky1Mla2mKcdbY4exXCZYsqlOGknyLQh4sCvL0nYZX2RR/aE+E318Kd/N9nflBHewE0kCFUHxP30ftET+qN5yH97Hns7TmmCKpSmsS8YG6n9GS1GqAAd7WVXiVJ7DcoeqnBgcZYzX6KSCU9sb1g+hJf5tHQoRBrifsZHhmyJDZmv7bSiV6zHGWE+JuwpFc5QKwGOuqCPSQ4hXyg1iYhtOVN18xeoSnHUCguzOhkGF7+PuMVi9TnmlybeEEMUyobpRcnQA4FOwagnig5yGgW9OqjemOgxiB8EYBC3AIHbDnISEC99b5RMSwng5LgjpcfJJenz8lAwoXPrf8Sj0bKcGHBZTTasUhApopvznDrBFARuJe2SHxQ5mv97FOZQn3EamfLDTdGAjLPQZYAPcr240Y53n8PSGfNRleQzWiNJxxWx9OmlnQZpTQoEQkhGb+AxSShQ/IRXcSHfWrA/e8aMvjQe2qrZepj7qG6Uzt3tFSKCRWQaoCEV557aDjEeNT44CtFaaIUoEMDGO/OKmev2ghio8NJay6F8gojJBftLi2AM4UKv9kEtK+Wq9eGeor1DlrwvKWh4Ooh3VBPbjIxBqat1UyY2TDasrdu91AdIKaANc4UI6WeHoMCwmTGg7NPdSQ6lF6ZFukNz7od7ccXLaYxAwFci1WS7+9+28kK0JbJjAzyP0ml4BQ+CBpymmLnRoK8SjuMCtzqvWrYvxbymHDwnPZ6ofv/PAyv3Z7MpbQQaNQkf8Y9OkCjhm6PCPKHgPvHCj7mdSAi8WhTCQSTf8LG0cLqSu3/uVO0XJoI2JnwzbbUBMIs4ePAILcHtIJKtxMqXlxjS0Q1dORS69oPEudfjEoPe1WwElLyUQE5YYtri4VZKPUyImwogsmux2kNW4jFC1HGDIjFVSMuS92TdZaH8JW+h1bFSnEdUtPGcgo53D8+E6N2qWewqLwa6Edklc9Qh9IymZOK8E16GfXXnjhoa+OiLGqYrnYGJrl18xti9v1TA319Glv7kBVYx6DTwYJIqKjV2MOB+C2FcRRNgdBjgJk71M6V9dT/8uyxoPhGRGYo1Tdi2uhYjZf8d5U0ti/E7z+CTEpSZg+nfF500K6gmNUX6tv1xSl24NbImVzuDrmxZcuab3gqXK+IvwIKS4viOMWhDWrt0g57rTZ/F3ff5yz7Iegn2Iv3pw6ZHWoWq4DLepIN1gghimcS9OelpqPtSYFPXcMzsreIumhZSgHV2cNXnHPjRq4D53ErN/VewqGExSuv50VYWyX8xEOTpb+PZSHq6UpS6KKul1eTDPoomTY0fbSjAcXScIxakA3Kk2XZpMjVJX7eB7J0ia16UOr7SvG/9YPH+q9UvfczZTYjMglj7mRGtQF8kuFYmPGAk7bnaROOwobH4kOJUvAavjqzSeXuxHvV96IuuxIVBZH9wLut4hlQoRMDnzv17l/iJM/5RB1hD7wozf3FJFV5GERx7QTIxF+XWY13nylXL96u5oy7E2EbYcBXMDh2eCxtG9A6Y4fpsONB/Os9ccKObAY4C3odl2cOm3jxWnQi08ZSazMUWR3PBxlKC4yfVOABLtQxd374Cqu4rMkzJc0zUpk6WMt8FPyk/cMqAj2TBnt2/cobJP0lTu3/fUOBAR9CPC+Ju3M9dkVttDDcdtsnA+XeJzoTNkU+bsJilQ5/84lUkfeknC2Ay4kQ0eEtlVvXcsI2gXUXyvGZZ4VMGhz+O24BXVPYVbjkOxc+LagRzZrK3RMEgilSNS3yd0WZ48T+Wa8Es/etEGjo5VqVMP+9ELdoEC997qgrHdELCijj7CaVXBonzwPsmP03CPKxKbHLKl27p7y9aiS1Ki3SR0yW/liEyx5PltOr90aeC6Bp1NCqXuQt3rJxBZQ0/3fLH9MWkWxrORkPzw6sJjwl+H6Irw9ZNvXWMsxzQHEe3aeQp06LVqvqQNPe1UNSuB4jrd/7h4d/q4mHeHVEfSDFYyR0RorkjDZV0NFl60KIaNyqF+L01m6eY+37IZQe77CooxiE7E5G0Z1T4C+u5WnOd15twoxHewQ/1tp1k7InJgZIjuIjydiG7+L8PYV1dIbGpEgKbIvdyNjhoXLTAjtJ2i0HBRgqs7iNQyluuggiNQ3kGRZZyUfjc0zrHUYgq9CEC2lfon3kUki81BZqO/nLByO7Yx5FuBKy0NaTODuE5bMN6ZZCXcyeGAJvEndKysB4PX66U5R8weyXYreXcW4RQnDm0QOxYrCCbXvsDwWm2K+Q9xDUmXJKul85Z6Mg249suoPavErmdDSLYpy5qgk+s13y36wYqeoSZcwLSKTbRjh1E8+WGYtUL7eaopRSV3JieMlpgAuQWLajfnEnlbxB1voPYr7rUlZlM24jJyOhazA3NGE0/JgrzV5mu2Sd5SHrkOmld/3sdn54pKXhfLBNZKUeybNR7CBBm0C7QRN5+j+pvfnX4L5itiSmTu08/7QV2rQWz7dcC8lqd2ZtzhhqdYoNY21HUjTGlzQ8wr/eOwHxoE11+++eHcsc5YJoVq6TR+nTHzziISRwNNdRLZJgBrzxNewvESlbkXD8IhhrAgcoW88v1O/XSatleeoZq2zPKvAnz4a63N3EFjPETOFaN6DS+FOosSM2EGFCMBhgk14TJZWUiLDECFwoy//ZPNRvE/dewmDOI1d19AzYo71AmB8U04W7zdG4BToIzCdZC7gfoPOrQeMdXejreZ3e3vRETGbXE2HuOJcgF+SFVg5Uw0cxDp2XIo4rBLRFPLxCznfUSXpdmwSlX2VFjZlQQp82vc4R11L5oteCUcOsjlp5FcPSDRZC2rsSQLyvrrLwsAylJqmCmElK8zOdXfQ6hAUvmD45RpzIIrzNMX9/G9HS4K85J0Zeh1DVrmnMAutOBsIAy3TYj2TgjdZAy/byTN5eWePY2aWn0M9yWz8uMITxfmUVi9mxAI3gxeqHX942OSZ1FqxuBVz1U/PIvhuaDMn+PTYKuUY9nMUxD3djP06A5oc2PhONRmg7IdQEJXx9F0C0nDSA77RssvzzvSOdStZ4QCd5KpmmH84rUmIysFlNJqVj68D0Qb8viOxULTfWYqq9/hWUlgP4wmK7Bu9koQ2VzD76YYkZpm3qIR0Vos9YbRaN1Zzcbmvbjlh09v6NlJ5z+Cl8KletHFGkd8IaTJe7QEb0s6Fv6exmuAkNiKzcdoCToKEF2Edfjp4vGIAaSHsJbz2sRex2GQCbz+juqrdB5vCPW+PzZF7hiUQKw2eYwc7IkTont++vE3ND9l6nkKCVDZY/1NOEEa9MSTPRgDdJXQ5uU1dkLV423Qlv9AnG/KF7rUOC4Sav+JNSUnKMOEJKScaBqChO4bYKA34L1HGxkBXCp1zUdCnHGh8jww3FcZbjZSkLw5KpRnVyxWaNtmfuesLg8aKhVZ+xVVtjBN3uKgSeXjC+udXYl1zfW+Y59D0Mg2bnyc2vAzoOh/IbvcetjgQyKqWO5MTzFN6rKef28y+VML+DOmlbaK42juiifX/BX4PqL3Hlhs3bxIlqMVGP4Tl5Nhk1xli7+6czqaFwY7ZSuN/Rz2IgQXxkAix6ZHzgxbpmNHlrCebASv/oBfan/QjCWRGCd1OPRwC4+BmX1O445AOOfcSVrok2eLBQODsyu2ODFRB0yVSDUeEsrrPgOnWxDR4h5LJNH+PWxTHipcMLMQFZ3ii030AaGfrrkZptlmtJNs5i//+4utnflcqSBYZZvHxQPBmhZTv2ulUoupdy14HJwd9YZqpO6f8uPT/sNXlm4OrhIGWvfMLXqC2tXmMOln86ON5nrBfA8lUTFRw5m2Yauc5dpUJRsza9PKlZhBZ/rywLy2IgPKFHkwSZEb1GfKMG75d53MfJEsgXtTtybsW46hNYaCLKTcuCAAqJELT97g4s9cVabQgrci7mhpAod+CeEYLagFFrsSIu5qI0zX4auy5yAxE16mw8euKYVcXgr4AFOldt0st18mWlXF7/Kp/GBoSBLFbmF+Sd4Of885RLhu6xdDmyz/SF3CT+x2otYRi0AGvcbctYE/G/DxIm29vDOBuCfoZBPBfQWesuz5vBSawcjTJA32iVn2Ieymv3M8oOIiIkmcnsvjPJR/N12ZAxCiIKZ+MMwP5D+6+pQkEK6BM6rLzaksfhhuqvzdUJ444lTQwTcR6lzMqFTo6DJ/D/ySDZEmbU5eeylDsJHHuRD1cJ71QyJiFwZpq7ApaO+9IWrZJOIVocSFQVkgxrb6flOn9ANMI6kvCh5B6VkUDGYm7UQF6S8qF+tMyg8tXWTTWx2p1dYh5PYEQz7a2G6HSA9JaJA+q6f9TWBjBoG/TBC8/d4m19VJX5lMzyRsoLU+OMRxUayKzfXCZ7lnePnbJtmifmKWDskM/wECZHWfXMGx7SwQA24VhojrnqzkWyWpJ+D99k+IMEGR2V9fSW2CU6Y6t0rBKQxaEGyJIWhU9nFDMBxLgNa5anfz46sh47Cdv2cZ7F3WnjfbVtbSiW82g5V2pFusA/1Z7M4keNdsJZQccrgngBDQW6Tudvcqrq01f3E4yzPFB5w8VuG4j3tYRgwyeaRyYpIAJeIvdfRMx9uJb81Xu8/vtQLs4CNN7gFR0nTK8Z0xwI9zv6swbpXhlBo5VT4Z3fDbYAAQQWABQfcAeoVSlqQMZzhmG+2fdUugf8dwf8BHBzZXQIBAAAAAAH/ARwc2V0AiBMAp9ID1rRhD6ZQ8z4l6WokYLjcar7exwtSOM8+40Ligf8BHBzZXQBIQlRlTRhpoOmzthRKQq/hBcxYjnkmgpAPux9E1xzDJycgQf8BHBzZXQDIQu0HzHRH3lqNkvPc+KH0UxzGlYTL1EbXSkMwZboj84ENwf8BHBzZXQGIQIcBGAzCoZahwGSu2dVIGK9QeXANk6tgQMn42O13wj0eQf8BHBzZXQFQwEAAaRAtPTkSqNxS0Go4ifRVS3MhH1uUe0YTKmJnFnlpbFmNjZocK+Gt3aaAz01Oz0MbIRd8MpGrGOX8kKkTLgKeZ8AIgICr1av/3qQPQSOBFDewV0d0yjVGRtZJiNZS4xZ9URU/BoYZtCbtVQAAIABAACAAAAAgAEAAAAKAAAAAQQWABSexsZF9BWn0s1JKJ/neL4SjIptyQf8BHBzZXQIBAAAAAAH/ARwc2V0AiBMAp9ID1rRhD6ZQ8z4l6WokYLjcar7exwtSOM8+40Ligf8BHBzZXQBIQgGY8MymMoR77NyNE6PxvhAIz/V+AIECv5uEKvR53K8WQf8BHBzZXQDIQs55PwsAcBf2apYu4msuXUHP0GE2rBLR7l/4NHNz5ewnQf8BHBzZXQGIQOgRrE7l/r0uqik2WM9ykQ8BB1z8zp70a5VasRnboomAwf8BHBzZXQFQwEAAREyiORynry6/f+A8ubFRI1MhXGY25waVKQxPjd0/Gr6pNtslqE9PtKHmfP8Tj/dXwX/c9FSPQzOE8AHG1pBTa0AAQMIXgUAAAAAAAABBAAH/ARwc2V0CAQAAAAAB/wEcHNldAIgTAKfSA9a0YQ+mUPM+JelqJGC43Gq+3scLUjjPPuNC4oA
```

```
elementsd: /usr/include/boost/optional/optional.hpp:1206: boost::optional<T>::reference_const_type boost::optional<T>::get() const [with T = long int; boost::optional<T>::reference_const_type = const long int&]: Assertion `this->is_initialized()' failed.
Aborted (core dumped)
```

tiero:
AFAIK the current psbt commands in elements are not reliable/broken, so not in sync yet with new PSET spec merged. @stepansnigirev

stepansnigirev:
> AFAIK the current psbt commands in elements are not reliable/broken, so not in sync yet with new PSET spec merged. @stepansnigirev

I am testing @achow101's pset branch (this PR), and I thought it implements exactly that.

stepansnigirev:
`changeAddress` option of `walletcreatefundedpsbt` doesn't work if the user is sending more than one asset.
Would it make sense to change it to `changeAddresses` and accept a list there so Elements Core can pick one of these addresses for every asset?

achow101:
> Is there a way to get psbt with asset blinding factor and value blinding factor from the RPC (and ideally with the rangeproof nonce)? Otherwise HW can't verify what amount is sent to which address. Maybe we could add an option that allows to keep the blinding factors in the psbt?

The encrypted blinding factors are included in an output's rangeproof via a side channel. The rangeproof for the output being spent is included in a `PSBT_ELEMENTS_IN_UTXO_RANGEPROOF` field. So a HWW should be able to extract the encrypted blinding factor from that rangeproof.

> How can I sign psbt with sighash `ALL | RANGEPROOF`? Looks like only normal bitcoin sighashes are supported by `walletprocesspsbt` command

The parser was missing `|RANGEPROOF` options. I've added a commit that implements this.

> `changeAddress` option of `walletcreatefundedpsbt` doesn't work if the user is sending more than one asset.
> Would it make sense to change it to `changeAddresses` and accept a list there so Elements Core can pick one of these addresses for every asset?

This is currently expected behavior. Since it involves more than just PSET (it also affects `fundrawtransaction` and coin selection in general), I think that change can be left for a followup.

> This PSET crashes `elementsd` on `decodepsbt` (it might be not a valid PSET though)

This PSET is not valid. Your outputs have an explicit asset and an asset commitment, but this is not allowed. The outputs also have a surjection proof but no rangeproof, and this is an invalid state to be in. I've added further error checking so that these become parsing errors instead of crashing elementsd.

apoelstra:
Commit abc27a5484962f8690988914baff07d7bb5d2abd does not build

src/rpc/rawtransaction.cpp
```diff
@@ -1844,12 +1844,53 @@ static RPCHelpMan combinepsbt()
...
+            has_fully_blinded = true;
+            break;
+        }
+    }
+
+    // Swap the psbt we want to use as base with the 0'th psbt which is the position for the base psbt
+    if (base_psbt_index > 0) {
+        PartiallySignedTransaction psbt = psbtxs[base_psbt_index];
+        psbtxs.insert(psbtxs.begin(), psbt);
+        psbtxs.erase(psbtxs.begin() + base_psbt_index + 1);
```
apoelstra, line 1749: nit: surely we could swap more efficiently than this?

src/pegins.cpp
```diff
@@ -533,3 +533,46 @@ CScriptWitness CreatePeginWitness(const CAmount& value, const CAsset& asset, con
...
     return CreatePeginWitnessInner(value, asset, genesis_hash, claim_script, tx_ref, merkle_block);
 }
+
+bool DecomposePeginWitness(const CScriptWitness& witness, CAmount& value, CAsset& asset, uint256& genesis_hash, CScript& claim_script, boost::variant<boost::blank, Sidechain::Bitcoin::CTransactionRef, CTransactionRef>& tx, boost::variant<boost::blank, Sidechain::Bitcoin::CMerkleBlock, CMerkleBlock>& merkle_block)
+{
+    if (witness.stack.size() < 5) return false;
+
+    const auto& stack = witness.stack;
+
+    CDataStream stream(witness.stack[0], SER_NETWORK, PROTOCOL_VERSION);
```
apoelstra, line 543: Should this just be `stack` (and elsewhere)?

src/pegins.h
```diff
@@ -42,5 +44,7 @@ std::vector<std::pair<CScript, CScript>> GetValidFedpegScripts(const CBlockIndex
 CScriptWitness CreatePeginWitness(const CAmount& value, const CAsset& asset, const uint256& genesis_hash, const CScript& claim_script, const CTransactionRef& tx_ref, const CMerkleBlock& merkle_block);
 CScriptWitness CreatePeginWitness(const CAmount& value, const CAsset& asset, const uint256& genesis_hash, const CScript& claim_script, const Sidechain::Bitcoin::CTransactionRef& tx_ref, const Sidechain::Bitcoin::CMerkleBlock& merkle_block);
 
+/** Break out the individual parts of the peg-in witness stack */
+bool DecomposePeginWitness(const CScriptWitness& witness, CAmount& value, CAsset& asset, uint256& genesis_hash, CScript& claim_script, boost::variant<boost::blank, Sidechain::Bitcoin::CTransactionRef, CTransactionRef>& tx, boost::variant<boost::blank, Sidechain::Bitcoin::CMerkleBlock, CMerkleBlock>& merkle_block);
```
apoelstra, line 48: This method needs to be visible from psbt.cpp, missing `#include` (also this commit needs to appear much earlier as `DecomposePeginWitness` is first used several commits ago, see my other comment about not compiling)

src/rpc/rawtransaction.cpp
```diff
@@ -2042,48 +2047,57 @@ static RPCHelpMan createpsbt()
...
-            VectorReader vr_proof(SER_NETWORK, PROTOCOL_VERSION, pegin_witness.stack[5], 0);
-            if (Params().GetConsensus().ParentChainHasPow()) {
-                Sidechain::Bitcoin::CTransactionRef tx_btc;
-                vr_tx >> tx_btc;
-                psbtx.inputs[i].peg_in_tx = tx_btc;
-                Sidechain::Bitcoin::CMerkleBlock tx_proof;
-                vr_proof >> tx_proof;
-                psbtx.inputs[i].txout_proof = tx_proof;
+        if (!rawTx.vin[i].assetIssuance.IsNull()) {
+            const UniValue& blind_reissuance_v = find_value(request.params[0].get_array()[i].get_obj(), "blind_reissuance");
```
apoelstra, line 1927: We should probably have an `else` on this if statement where we also check this field, and if it's present, we should return an error rather than silently failing to issue the asset.

apoelstra: Should check all the issuance-related fields*

src/rpc/rawtransaction.cpp
```diff
@@ -3078,6 +3078,108 @@ static RPCHelpMan calculateasset()
     };
 }
 
+static RPCHelpMan updatepsbtpegin()
+{
+    return RPCHelpMan{"updatepsbtpegin",
+            "\nFill in Peg-in input data for a particular input in a PSBT. Data is filled if provided.\n"
+            "The Peg-in witness will be constructed and finalized if everything is available.\n",
```
apoelstra, line 2962: This comment is a bit unclear ... it seems to suggest that this RPC will construct the pegin witness and include it in the PSET somewhere, but what you mean is that **during signing** (or finalization? finalization seems to make more sense but in a later commit you implement this in `SignPSBT`) a pegin witness will be constructed only if all fields are available.

src/blindpsbt.cpp
```diff
@@ -0,0 +1,430 @@
...
+};
+
+static Blind_PSBT_ECC_Init ecc_init_on_load;
+
+// Create surjection proof
+bool CreateAssetSurjectionProof(std::vector<unsigned char>& output_proof, const std::vector<secp256k1_fixed_asset_tag>& fixed_input_tags, const std::vector<secp256k1_generator>& ephemeral_input_tags, const std::vector<uint256>& input_asset_blinders, const uint256& output_asset_blinder, const secp256k1_generator& output_asset_tag, const CAsset& asset)
+{
+    int ret;
+    // 1 to 3 targets
+    size_t inputs_to_select = std::min((size_t)3, fixed_input_tags.size());
```
apoelstra, line 66: Can you replace this with a new constant in blind.h (and find the other place in the wallet that we hardcode 3). We'd like to increase this number at some point, it was chosen to copy Monero in 2015, and they have since increased their number to 11..

src/blindpsbt.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+#include <hash.h>
+#include <primitives/transaction.h>
+#include <primitives/confidential.h>
+#include <psbt.h>
+#include <issuance.h>
+#include <random.h>
+#include <util/system.h>
+
+static secp256k1_context* secp256k1_blind_context = NULL;
```
apoelstra, line 15: Is there anyway we can reuse the static context from blind.cpp? Each of these costs like 5Mb, which starts to add up in CI when running 4 multi-daemon functional tests in parallel. (Though I guess we have only one or two tests which use PSET so maybe it'll be ok..)

src/blindpsbt.cpp
```diff
@@ -0,0 +1,430 @@
...
+    // Find correlation between asset tag and listed input tags
+    if (secp256k1_surjectionproof_initialize(secp256k1_blind_context, &proof, &input_index, &fixed_input_tags[0], fixed_input_tags.size(), inputs_to_select, &fixed_output_tag, 100, randseed) == 0) {
+        return false;
+    }
+    // Using the input chosen, build proof
+    ret = secp256k1_surjectionproof_generate(secp256k1_blind_context, &proof, &ephemeral_input_tags[0], ephemeral_input_tags.size(), &output_asset_tag, input_index, input_asset_blinders[input_index].begin(), output_asset_blinder.begin());
+    assert(ret == 1);
+    // Double-check answer
+    ret = secp256k1_surjectionproof_verify(secp256k1_blind_context, &proof, &ephemeral_input_tags[0], ephemeral_input_tags.size(), &output_asset_tag);
+    assert(ret != 0);
```
apoelstra, line 82: For consistency, we should assert `== 1` like above

src/blindpsbt.cpp
```diff
@@ -0,0 +1,430 @@
...
+    int ret;
+    conf_value.vchCommitment.resize(CConfidentialValue::nCommittedSize);
+    ret = secp256k1_pedersen_commit(secp256k1_blind_context, &value_commit, value_blinder.begin(), amount, &asset_gen);
+    assert(ret != 0);
+    secp256k1_pedersen_commitment_serialize(secp256k1_blind_context, conf_value.vchCommitment.data(), &value_commit);
+    assert(conf_value.IsValid());
+}
+
+// Subtract b from a in place
+bool SubtractScalars(uint256& a, const uint256& b)
```
apoelstra, line 146: Probably best to do this in a followup PR, but Sanket found a simpler way to do all this arithmetic by calling `secp256k1_pedersen_blind_generator_blind_sum` in secp-zkp.

src/wallet/wallet.cpp
```diff
@@ -2834,103 +2834,111 @@ BlindingStatus CWallet::WalletBlindPSBT(PartiallySignedTransaction& psbtx) const
...
-                pegin_witness = CreatePeginWitness(*input.value, input.asset, input.genesis_hash, input.claim_script, btc_peg_in_tx, btc_txout_proof);
+                pegin_witness = CreatePeginWitness(*input.m_peg_in_value, Params().GetConsensus().pegged_asset, input.m_peg_in_genesis_hash, input.m_peg_in_claim_script, boost::get<Sidechain::Bitcoin::CTransactionRef&>(input.m_peg_in_tx), boost::get<Sidechain::Bitcoin::CMerkleBlock>(input.m_peg_in_txout_proof));
             } else {
-                const CTransactionRef& elem_peg_in_tx = boost::get<CTransactionRef>(input.peg_in_tx);
-                const CMerkleBlock& elem_txout_proof = boost::get<CMerkleBlock>(input.txout_proof);
-                pegin_witness = CreatePeginWitness(*input.value, input.asset, input.genesis_hash, input.claim_script, elem_peg_in_tx, elem_txout_proof);
+                input.m_peg_in_witness = CreatePeginWitness(*input.m_peg_in_value, Params().GetConsensus().pegged_asset, input.m_peg_in_genesis_hash, input.m_peg_in_claim_script, boost::get<CTransactionRef&>(input.m_peg_in_tx), boost::get<CMerkleBlock>(input.m_peg_in_txout_proof));
```
apoelstra, line 2895: I'm having trouble understanding this. In the PoW case we assign the pegin witness to `pegin_witness` and it is later put into the transaction; meanwhile we leave `input.m_peg_in_witness` alone. In the non-PoW case we assign the pegin witness to `input.m_peg_in_witness` and leave `pegin_witness` blank, so we then blank the pegin witness in the transaction.

If this is intended it at least needs a comment explaining the discrepancy between the PoW and non-PoW behavior.

src/rpc/rawtransaction_util.cpp
```diff
@@ -280,102 +274,101 @@ CMutableTransaction ConstructTransaction(const UniValue& inputs_in, const UniVal
...
-            for (size_t i = 0; i < vdata.size(); i++) {
-                std::vector<unsigned char> data = ParseHexV(vdata[i].get_str(), "Data");
-                datascript << data;
-            }
+        bool is_fee = false;
+        for (const std::string& name_ : output.getKeys()) {
+            if (name_ == "data") {
+                if (has_data) {
+                    throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, duplicate key: data");
+                }
```
apoelstra, line 298: Might be better to just check that no keys are duplicated rather than special-casing this. (I realize that you are just copying the existing logic here, but the existing logic seems wrong/incomplete.)

src/rpc/rawtransaction_util.cpp
```diff
@@ -355,28 +348,31 @@ CMutableTransaction ConstructTransaction(const UniValue& inputs_in, const UniVal
...
+            psbt_outs.push_back(psbt_out);
         }
     }
 
     // Add fee output in the end.
     if (!fee_out.nValue.IsNull() && fee_out.nValue.GetAmount() > 0) {
         rawTx.vout.push_back(fee_out);
-        if (output_pubkeys_out) {
-            output_pubkeys_out->push_back(CPubKey());
+        psbt_outs.emplace_back(2);
```
apoelstra, line 370: Can you add a comment here (and several lines above at `PSBTOutput psbt_out(2);`) that "2" represents a dummy PSBT output with version 2?

src/primitives/confidential.h
```diff
@@ -201,7 +201,7 @@ class CAssetIssuance
     SERIALIZE_METHODS(CAssetIssuance, obj) { READWRITE(obj.assetBlindingNonce, obj.assetEntropy, obj.nAmount, obj.nInflationKeys); }
 
     void SetNull() { nAmount.SetNull(); nInflationKeys.SetNull(); }
-    bool IsNull() const { return (nAmount.IsNull() && nInflationKeys.IsNull()); }
+    bool IsNull() const { return (nAmount.IsNull() && nInflationKeys.IsNull()) && assetBlindingNonce.IsNull() && assetEntropy.IsNull(); }
```
apoelstra, line 204: lol I think this might be a hardfork

src/rpc/rawtransaction_util.cpp
```diff
@@ -280,102 +274,101 @@ CMutableTransaction ConstructTransaction(const UniValue& inputs_in, const UniVal
...
-            if (!output.isObject()) {
-                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, key-value pair not an object as expected");
-            }
-            if (output.size() != 1) {
-                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, key-value pair must contain exactly one key");
-            }
-            outputs_dict.pushKVs(output);
-        }
-        outputs = std::move(outputs_dict);
-    }
```
apoelstra, line 263: Has this been done in Core as well? It feels weird to have a breaking RPC change in `createrawtransaction` for Elements 0.21 that isn't reflected in Core 0.21. I think we sholud revert this for 0.21.

src/psbt.cpp
```diff
@@ -496,6 +496,12 @@ bool PSBTOutput::Merge(const PSBTOutput& output)
     return true;
 }
 
+CTxOut PSBTOutput::GetTxOut() const
+{
+    assert(amount != nullopt);
+    return CTxOut(*amount, script);
```
apoelstra: `*script` (`script` is an optional too)

apoelstra:
Done review. This looks great! Most of my comments were fairly minor, except:

* Many of the commits don't compile; need to rebase to fix this
* The change to `AssetIssuance::IsNull` is dangerous because this function is called many times during transaction verification. Can we drop this? Otherwise we need to reason very carefully about the consquences.
* I'd prefer not to break `createrawtransaction` this release unless Bitcoin Core has the same breakage.

src/rpc/rawtransaction_util.cpp
achow101: No, but because we need this extra data for elements, it gets really messy to maintain the other way too. The other way to specify outputs is supposed to be deprecated anyways, but it hasn't been removed.

src/psbt.cpp
achow101: script is a `CScript`, not an optional. Should it be?

apoelstra: I get a compile error about this when it is introduced, but it is fixed several commits later (in the commit that adds `nullopt` initializers to all the options). It may be that I am misinterpreting the compiler error and that it's somehow about initialization rather than about types, but regardless, this commit doesn't build.

apoelstra: Specifically a compiler error about not finding an instance of `CTxOut()` which works for the given parameter types. I forget the exact error.

src/rpc/rawtransaction_util.cpp
apoelstra: Ok, fair enough

src/psbt.cpp
achow101: Oh, script is a CScript originally, then made into an optional in a later commit.

src/rpc/rawtransaction_util.cpp
achow101: I would prefer for a followup to do that.

src/rpc/rawtransaction.cpp
achow101: Changed to use `std::swap`

src/pegins.cpp
achow101: Yes, done.

src/pegins.h
achow101: Reordered and added missing include.

src/rpc/rawtransaction.cpp
achow101: Done in `ConstructTransaction`.

src/rpc/rawtransaction.cpp
achow101: Removed this line.

src/blindpsbt.cpp
achow101: Done in a new commit.

src/blindpsbt.cpp
achow101: Done (I think).

src/blindpsbt.cpp
achow101: Done here and elsewhere.

src/wallet/wallet.cpp
achow101: That's a mistake. Changed to set `pegin_witness`.

src/rpc/rawtransaction_util.cpp
achow101: Done.

src/primitives/confidential.h
achow101: Removed that commit.

apoelstra:
utack e3a40815a5f95b910b56396071559d0615a6add7

I need an hour or two to build-test all the commits and run some tests on the tip, and we need to merge #1012 to fix CI, then we should be good to go!

apoelstra:
ack f8553c571b49ca8ee7f1a731455581c30bfec06f

apoelstra:
CI failure due to #1012

PR #911, opened 2020-09-30
When decoding or signing a PSBT, check blinded values.

gwillen:
When we decode or sign a PSBT given to us via RPC, first check that, if it contains blinded output values, they verifiably match the unblinded output values contained in the original transaction proposal. Otherwise fail.

### DIFF
--- a/src/blind.cpp
+++ b/src/blind.cpp
@@ -36,6 +36,74 @@ public:
 
 static Blind_ECC_Init ecc_init_on_load;
 
+bool VerifyConfidentialPair(const CConfidentialValue& conf_value, const CConfidentialAsset& conf_asset, const CAmount& claimed_value, const CAsset& claimed_asset, const uint256& value_blinding_factor, const uint256& asset_blinding_factor) {
+    if (conf_value.IsNull() || conf_asset.IsNull() || claimed_asset.IsNull()) {
+        return false;
+    }
+
+    if (conf_value.IsExplicit()) {
+        // Match behavior of UnblindConfidentialPair
+        return false;
+    }
+    if (conf_asset.IsExplicit() && conf_asset.GetAsset() != claimed_asset) {
+        return false;
+    }
+
+    // Just to be safe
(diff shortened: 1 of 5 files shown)


src/blind.cpp
```diff
@@ -36,6 +36,74 @@ class Blind_ECC_Init {
...
+    if (conf_value.IsExplicit()) {
+        // Match behavior of UnblindConfidentialPair
+        return false;
+    }
+    if (conf_asset.IsExplicit() && conf_asset.GetAsset() != claimed_asset) {
+        return false;
+    }
+
+    // Just to be safe
+    if (claimed_value > (int64_t)MAX_MONEY || !MoneyRange(claimed_value)) {
```
dgpv, line 53: claimed_value is already CAmount, and MoneyRange checks `nValue >= 0 && nValue <= MAX_MONEY`, is `claimed_value > (int64_t)MAX_MONEY` needed here ?

It would be needed if claimed_value would be `u_int64_t`, but I think here this extra check is not needed

gwillen: Hmm good question, I think I copied and pasted this check from elsewhere.

@stevenroose, thoughts? I will try to take a closer look tomorrow morning (my time).

dgpv: A similar code is at https://github.com/ElementsProject/elements/blob/ec9b28e0897c270bc551031741adb7361cec34b5/src/blind.cpp#L93-L95
but the difference is that `amount` there is of type uint64_t as received from  `secp256k1_rangeproof_rewind`. IIRC, this check is needed because unsigned to signed conversions when unsigned is outside of the signed range is implementation-defined (up until some recent C++ version, I don't remember which)

stevenroose: Makes sense. The claimed value indeed comes from a CAmount from the unblinded tx, so I agree it doesn't need the extra MAX_MONEY check.

gwillen: Fixed.

stevenroose:
ACK 8a73f5d24

PR #913, opened 2020-10-06
[0.18 backport] Backport #896: Fix logic issues with randomization of outputs in [re]issueasset_base; minor refactoring

stevenroose:


gwillen:
backport ACK 91ec3dd16 (2 commits)

gwillen:
(this convention, which I have just now made up, means that I have verified that the last two commits here match the last two commits from the original PR, according to `git range-diff`.)

PR #905, opened 2020-09-02
Update shellcheck URL

gwillen:
Update shellcheck URL per  per https://github.com/koalaman/shellcheck/issues/1871 to unbreak CI.

PR #919, opened 2020-10-06
[0.18 backport] Backport #905: Update shellcheck URL per https://github.com/koalaman/shellcheck/issu…

stevenroose:


philippem:
utACK

PR #932, opened 2020-12-02
Update TestFramework segwit SigHash to include issuance

sanket1729:
c++ code link for review aid and comparison: https://github.com/ElementsProject/elements/blob/88067ddbe4223e43a4c86912928b28b834bbf69e/src/script/interpreter.cpp#L1622

### DIFF
--- a/test/functional/test_framework/script.py
+++ b/test/functional/test_framework/script.py
@@ -696,9 +696,11 @@ def SegwitVersion1SignatureHash(script, txTo, inIdx, hashtype, amount):
 
     if not (hashtype & SIGHASH_ANYONECANPAY):
         serialize_issuance = bytes()
-        # TODO actually serialize issuances
-        for _ in txTo.vin:
-            serialize_issuance += b'\x00'
+        for i in txTo.vin:
+            if i.assetIssuance.isNull():
+                serialize_issuance += b'\x00'
+            else:
+                serialize_issuance += i.assetIssuance.serialize()
         hashIssuance = uint256_from_str(hash256(serialize_issuance))
 
     if ((hashtype & 0x1f) != SIGHASH_SINGLE and (hashtype & 0x1f) != SIGHASH_NONE):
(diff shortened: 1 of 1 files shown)


apoelstra:
Can you update the `feature_issuance.py` test to test this?

sanket1729:
Reading the file `feature_issuance.py`, I don't think this belongs there. It probably belongs in `p2p_segiwt.py` along with the other sighash checks.

apoelstra:
Sure :) whatever test makes sense, it would just be good to hit this code.

sanket1729:
This is more challenging to make a functional test than I initially presumed because:
1) I could not find support for creating range proofs or surjection proofs in python testframework. So, I cannot manually create transactions and send it to node to accept since I cannot create the proofs.
2) Another thing I tried is getting raw issue transaction by rpc, looking at transaction created and trying to manually verify. But there is some tooling missing for doing this. For example, parsing script sig hex into parts etc. 
3) test with hard-coded transactions across elements, but then this is a functional test, but a unit test. 

1) is probably a longer-term goal and 2) Is relatively simpler to do, but still a bit annoying.

dgpv:
Gregory Sanders asked me once how hard it would be to port the blinding code from https://github.com/Simplexum/python-elementstx/ to the test framework. I've looked and estimated that there will be a lot of work, and it might be more economical to just use python-elementstx to create a testing suite that would run separately from the main tests (so the main Elements repo won't have that dependency). If this is not an acceptable solution, python-elementstx can be used to get some bits of code out of it, but a large portion of the code can't be directly taken because the core classes differ significantly in many aspects

apoelstra:
We can probably use the `blindrawtransaction` or `rawblindrawtransaction` RPCs to do the blinding, if we can generate the transactions.

FWIW, the existing sighashes don't even cover the rangeproofs so I'm not sure this is necessary..

stevenroose:
Yeah I think Andrew is right that the raw blind RPC should be able to do the job. I'm not 100% but I haven't ran into a job it couldn't do yet in testing AFAIR. And I did the sighash rangeproof test :sweat_smile:

sanket1729:
Yeah, you are probably right. I did not sufficient understanding of elements back then, so maybe I was just confused. I will try to do this again.

apoelstra:
Closing in favor of #1045

PR #933, opened 2020-12-02
doc: add document describing Taproot sighashes

apoelstra:
Thanks to @roconnor-blockstream for writing the first draft of this document.

### DIFF
--- a/doc/taproot-sighash.mediawiki
+++ b/doc/taproot-sighash.mediawiki
@@ -15,6 +15,8 @@ The parameter ''hash_type'' is an 8-bit unsigned value. The <code>SIGHASH</code>
 
 '''''NEW''''' If the input under consideration is a pegin input, the fields ''nAsset'', ''nValue'' and ''scriptPubKey'', where they appear, are taken from the pegin witness data. The asset used is the asset ID on the sidechain, not that on the parent chain (if any).
 
+'''''NEW''''' The fields ''nNonce'', ''nAsset'' and ''nValue'' are serialized in a fixed-length format, which consists of their ordinary 1/9/33 byte encoding followed by sufficiently many 0 to pad the length out to 33. When we refer to ''extended <code>CTxOut</code> format'' below, we mean the fields ''nAsset'', ''nValue'', ''nNonce'' serialized in that order in fixed-length format, followed by the ordinary length-prefixed ''scriptPubKey''.
+
 The parameter ''ext_flag'' is an integer in range 0-127, and is used for indicating (in the message) that extensions are added at the end of the message<ref>'''What extensions use the ''ext_flag'' mechanism?''' [https://github.com/bitcoin/bips/blob/master/bip-0342.mediawiki BIP-0342] reuses the same common signature message algorithm, but adds BIP-0342-specific data at the end, which is indicated using ''ext_flag = 1''.</ref>.
 
 If the parameters take acceptable values, the message is the concatenation of the following data, in order (with byte size of each item listed in parentheses). Numerical values in 2, 4, or 8-byte are encoded in little-endian.
@@ -34,28 +36,28 @@ If the parameters take acceptable values, the message is the concatenation of th
 *** '''''NEW''''' ''sha_amounts_assets'' (32): the SHA256 of the serialization of all output ''nAsset'' fields followed by ''nValue'' fields
 *** ''sha_scriptpubkeys'' (32): the SHA256 of the serialization of all spent output ''scriptPubKey''s.
 *** ''sha_sequences'' (32): the SHA256 of the serialization of all input ''nSequence''.
-*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of all input ''assetIssuance'' or '0x00' for inputs with no issuance. (Note: this is the same messaged hashed that used in Elements Segwit V0 signatures.)
+*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of all input ''assetIssuance'' or 130 '0x00' bytes for inputs with no issuance
 ** If ''hash_type & 3'' does not equal <code>SIGHASH_NONE</code> or <code>SIGHASH_SINGLE</code>:
-*** ''sha_outputs'' (32): the SHA256 of the serialization of all outputs in <code>CTxOut</code> format.
+*** ''sha_outputs'' (32): the SHA256 of the serialization of all outputs in extended <code>CTxOut</code> format.
 *** '''''NEW''''' ''sha_output_witnesses'' (32): the SHA256 of the serialization of all output witnesses (rangeproof and surjection proof) in <code>CTxWitness</code> format.
 * Data about this input:
 ** ''spend_type'' (1): equal to ''(ext_flag * 2) + annex_present'', where ''annex_present'' is 0 if no annex is present, or 1 otherwise (the original witness stack has two or more witness elements, and the first byte of the last element is ''0x50'')
 ** If ''hash_type & 0x80'' equals <code>SIGHASH_ANYONECANPAY</code>:
 *** '''''NEW''''' ''outpoint_flag'' (1): the input's outpoint flags shifted right by 24 bits. (Compare ''sha_outpoint_flags'' above.)
 *** ''outpoint'' (36): the <code>COutPoint</code> of this input (32-byte hash + 4-byte little-endian).
-*** '''''NEW''''' ''nAsset'' (33): (possibly confidential) assetID of the previous output spent by this input.
-*** '''''NEW''''' ''nValue'' (9--33): (possibly confidential) amount of the previous output spent by this input.
+*** '''''NEW''''' ''nAsset'' (33): assetID of the previous output spent by this input, in fixed-length format
+*** '''''NEW''''' ''nValue'' (33): (possibly confidential) amount of the previous output spent by this input, in fixed-length format
 *** ''scriptPubKey'' (35): ''scriptPubKey'' of the previous output spent by this input, serialized as script inside <code>CTxOut</code>. Its size is always 35 bytes.
 *** ''nSequence'' (4): ''nSequence'' of this input.
-*** '''''NEW''''' ''asset_issuance'' (0 or 74--130): if ''outpoint_flag & 0x80 == 0x80'': asset issuance data of this input.
+*** '''''NEW''''' ''asset_issuance'' (130): if ''outpoint_flag & 0x80 == 0x80'': asset issuance data of this input; otherwise 130 zero bytes
 ** If ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
 *** ''input_index'' (4): index of this input in the transaction input vector. Index of the first input is 0.
 ** If an annex is present (the lowest bit of ''spend_type'' is set):
 *** ''sha_annex'' (32): the SHA256 of ''(compact_size(size of annex) || annex)'', where ''annex'' includes the mandatory ''0x50'' prefix.
 * Data about this output:
 ** If ''hash_type & 3'' equals <code>SIGHASH_SINGLE</code>:
-*** ''sha_single_output'' (32): the SHA256 of the corresponding output in <code>CTxOut</code> format.
+*** ''sha_single_output'' (32): the SHA256 of the corresponding output in extended <code>CTxOut</code> format.
 *** '''''NEW''''' ''sha_single_output_witness'' (32): the SHA256 of the serialization of the corresponding output witnesses (rangeproof and surjection proof) in <code>CTxWitness</code> format.
 
-The total length of ''SigMsg()'' is at most ''491'' bytes. Note that this does not include the size of sub-hashes such as ''sha_prevouts'', which may be cached across signatures of the same transaction.
+The total length of ''SigMsg()'' is ''442'' bytes for <code>ANYONECANPAY</code> sighashes, ''366'' bytes for non-<code>ANYONECANPAY</code> sighashes, and both numbers are reduced by 64 bytes for <code>SIGHASH_NONE</code> sighashes. Note that this does not include the size of sub-hashes such as ''sha_prevouts'', which may be cached across signatures of the same transaction.
 
--- a/doc/taproot-sighash.mediawiki
+++ b/doc/taproot-sighash.mediawiki
@@ -36,7 +36,8 @@ If the parameters take acceptable values, the message is the concatenation of th
 *** '''''NEW''''' ''sha_asset_amounts'' (32): the SHA256 of the serialization of ''nAsset||nValue'' for each output. As mentioned above, for pegin inputs an explicit value (from the parent chain) and asset (the peg asset) will be used.
 *** ''sha_scriptpubkeys'' (32): the SHA256 of the serialization of all spent output ''scriptPubKey''s.
 *** ''sha_sequences'' (32): the SHA256 of the serialization of all input ''nSequence''.
-*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of all input ''assetIssuance'' or 130 '0x00' bytes for inputs with no issuance
+*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of asset issuance data for inputs with an issuance or ''0x00'' for inputs with no issuance. (This matches the Segwit v0 encoding of this hash.)
+*** '''''NEW''''' ''sha_issuance_rangeproofs'' (32): the SHA256 of the serialization of the concatenation of ''issuanceAmountRangeproof||inflationKeysRangeproof'' for all inputs, where each field is encoded as ''0x00'' if absent (either if there is no issuance, or if there is an explicit issuance)
 ** If ''hash_type & 3'' does not equal <code>SIGHASH_NONE</code> or <code>SIGHASH_SINGLE</code>:
 *** ''sha_outputs'' (32): the SHA256 of the serialization of all outputs in <code>CTxOut</code> format.
 *** '''''NEW''''' ''sha_output_witnesses'' (32): the SHA256 of the serialization of all output witnesses (rangeproof and surjection proof) in <code>CTxOutWitness</code> format.
@@ -45,11 +46,12 @@ If the parameters take acceptable values, the message is the concatenation of th
 ** If ''hash_type & 0x80'' equals <code>SIGHASH_ANYONECANPAY</code>:
 *** '''''NEW''''' ''outpoint_flag'' (1): the input's outpoint flags shifted right by 24 bits. (Compare ''sha_outpoint_flags'' above.)
 *** ''outpoint'' (36): the <code>COutPoint</code> of this input (32-byte hash + 4-byte little-endian) where the output index excludes the outpoint flags.
-*** '''''NEW''''' ''nAsset'' (33): (possibly confidential) assetID of the previous output spent by this input, in fixed-length format
-*** '''''NEW''''' ''nValue'' (33): (possibly confidential) amount of the previous output spent by this input, in fixed-length format
+*** '''''NEW''''' ''nAsset'' (33): (possibly confidential) assetID of the previous output spent by this input
+*** '''''NEW''''' ''nValue'' (9-33): (possibly confidential) amount of the previous output spent by this input
 *** ''scriptPubKey'' (35): ''scriptPubKey'' of the previous output spent by this input, serialized as script inside <code>CTxOut</code>. Its size is always 35 bytes.
 *** ''nSequence'' (4): ''nSequence'' of this input.
-*** '''''NEW''''' ''asset_issuance'' (130): if ''outpoint_flag & 0x80 == 0x80'': asset issuance(fields for ''nInflationKeys'' and ''nAmount'' are serialized in fixed-length format) data of this input; otherwise 130 zero bytes
+*** '''''NEW''''' ''asset_issuance'' (1-130): if ''outpoint_flag & 0x80 == 0x80'': asset issuance data for this input, or ''0x00'' otherwise
+*** '''''NEW''''' ''sha_single_issuance_rangeproofs'' (0-32): if ''outpoint_flag & 0x80 == 0x80'': the SHA256 of the serialization of the concatenation of ''issuanceAmountRangeproof||inflationKeysRangeproof'' for this input, where each field is encoded as ''0x00'' if absent
 ** If ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
 *** ''input_index'' (4): index of this input in the transaction input vector. Index of the first input is 0.
 ** If an annex is present (the lowest bit of ''spend_type'' is set):
--- a/doc/taproot-sighash.mediawiki
+++ b/doc/taproot-sighash.mediawiki
@@ -17,8 +17,6 @@ The parameter ''hash_type'' is an 8-bit unsigned value. The <code>SIGHASH</code>
 
 '''''NEW''''' The epoch field prepended before signature hash is completely dropped. If there are new updates to the taproot signature hashes in elements, they will use new tagged hashes instead of incrementing epochs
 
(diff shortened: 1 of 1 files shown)


doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+** If ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** ''input_index'' (4): index of this input in the transaction input vector. Index of the first input is 0.
+** If an annex is present (the lowest bit of ''spend_type'' is set):
+*** ''sha_annex'' (32): the SHA256 of ''(compact_size(size of annex) || annex)'', where ''annex'' includes the mandatory ''0x50'' prefix.
+* Data about this output:
+** If ''hash_type & 3'' equals <code>SIGHASH_SINGLE</code>:
+*** ''sha_single_output'' (32): the SHA256 of the corresponding output in <code>CTxOut</code> format.
+*** '''''NEW''''' ''sha_single_output_witness'' (32): the SHA256 of the serialization of the corresponding output witnesses (rangeproof and surjection proof) in <code>CTxWitness</code> format.
+
+The total length of ''SigMsg()'' is at most ''491'' bytes. Note that this does not include the size of sub-hashes such as ''sha_prevouts'', which may be cached across signatures of the same transaction.
```
sanket1729, line 60: One general point for discussion. (Fixed length sighash with positions for fields)

- We observed that having a fixed length sighash with a fixed location for which field occurs where is super helpful in creating complex scripts easily. 
- Even though we will new opcodes for getting these directly, it might still be useful to have a structured sighash. 
- The only fields with variable length are `assetIssuance` and `nValue`( `assetIssuance` internally uses confidential value serialization). So, everything can be addressed if we fix the serialization of explicit in `nValue`. 

Note that we don't need to change the way serialization is done for these fields in p2p network, but only when we do sighash computation. 
Downsides:
- more code complexity
- more divergence from core

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,61 @@
...
+* Data about this input:
+** ''spend_type'' (1): equal to ''(ext_flag * 2) + annex_present'', where ''annex_present'' is 0 if no annex is present, or 1 otherwise (the original witness stack has two or more witness elements, and the first byte of the last element is ''0x50'')
+** If ''hash_type & 0x80'' equals <code>SIGHASH_ANYONECANPAY</code>:
+*** '''''NEW''''' ''outpoint_flag'' (1): the input's outpoint flags shifted right by 24 bits. (Compare ''sha_outpoint_flags'' above.)
+*** ''outpoint'' (36): the <code>COutPoint</code> of this input (32-byte hash + 4-byte little-endian).
+*** '''''NEW''''' ''nAsset'' (33): (possibly confidential) assetID of the previous output spent by this input.
+*** '''''NEW''''' ''nValue'' (9--33): (possibly confidential) amount of the previous output spent by this input.
+*** ''scriptPubKey'' (35): ''scriptPubKey'' of the previous output spent by this input, serialized as script inside <code>CTxOut</code>. Its size is always 35 bytes.
+*** ''nSequence'' (4): ''nSequence'' of this input.
+*** '''''NEW''''' ''asset_issuance'' (0 or 74--130): if ''outpoint_flag & 0x80 == 0x80'': asset issuance data of this input.
```
sanket1729, line 50: Should be (82--130) instead of (74--130)? The minimum size is 2 explicit values + 2 32 bytes = 18 + 62 = 82

sanket1729 (commented):
We support sidechains to sidechains? I was not aware of this. If not, what is `parent_pegin_asset` and how is that supposed to be used?

apoelstra:
Yes, we support sidechains to sidechains, and I believe some of our functional tests depend on this fact.

doc/taproot-sighash.mediawiki
apoelstra: The `inflationKeys` field can be null, in which case it is one byte.

doc/taproot-sighash.mediawiki
apoelstra: We could add zeros to the hash after nNonce nAmount and nAsset such that they'd always total out to 33 bytes. And same for the fields in the issuance.

I don't think this is too much added complexity and it would give you, as you say, fixed-offset fields in the sighash. I can certainly imagine that being useful for a covenant which is trying to fix one piece of the sighash while not touching anything else.

apoelstra: We would still have a variable-length encoding based on the sighash flag, and I don't think there's any reasonable way to avoid that (the different flags mean you're hashing different data). But it's easy to fix/check this flag since it appears on the siganture itself, not only inside the sighash.

Do you think we should 0-pad so that the issuance data field is always 130 bytes even if there is no issuance? This adds two full sha2 compressions (at about 200ns/shot on a desktop CPU) and will rarely be nonzero.

sanket1729: We should compare this overhead to normal signature validation?

sanket1729: If we still want to maintain length index prefixing, we could change the order in which we do the sighash. We could all the input specific things at the end of the sighash.

apoelstra: Normal signature validation is 50us. So an added .5us is a 1% performance hit.

I'd rather not change the order of the sighash too much to minimize the cognitive load for people coming from Bitcoin.

apoelstra:
@sanket1729 updated doc to make confidential fields all fixed-length, including issuance which is now always 130 bytes.

I don't think we care about 1% perf hit when validating ANYONECANPAY signatures. If we can get a taproot sig in a blockheader, that saves us 10 signatures' worth of validation which means that even if the block has 100 such inputs (unlikely, nobody uses ACP :)) Taproot will still be a net win.

roconnor-blockstream:
the `nValue` is in big endian, so it should probably get 0 byte prefixes instead of suffixes.

apoelstra:
Like, between the `01` byte indicating that it is explict, and the 8 value bytes? I think that would complicate the mapping between the normal format and the extended one.

roconnor-blockstream:
Good point.

apoelstra:
Rebased. What needs to be done to get this in?

sanket1729:
ACK 169af2d0345d12a63ad469d5a8d34b57fc08025f

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,65 @@
...
+** ''genesis_block'' (32): the genesis hash again
+* Control:
+** ''hash_type'' (1).
+* Transaction data:
+** ''nVersion'' (4): the ''nVersion'' of the transaction.
+** ''nLockTime'' (4): the ''nLockTime'' of the transaction.
+** If the ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** '''''NEW''''' ''sha_outpoint_flags'' (32): the SHA256 of the serialization of the concatenation of one byte per input of the input's outpoint flags shifted right by 24 bits. (The byte for an pegin input would be 0x40. The byte for an issuance input would be 0x80. The byte for both a pegin and issuance would be 0xc0.)
+*** ''sha_prevouts'' (32): the SHA256 of the serialization of all input outpoints. Each prevout output index is the masked version containing outpoint flags
+*** '''''NEW''''' ''sha_asset_amounts'' (32): the SHA256 of the serialization of all output ''nAsset'' fields followed by ''nValue'' fields
```
roconnor-blockstream, line 36: - This is a bit ambiguously described, is it pair of all assets and values or all pairs or assets and values? (yes I realize these words are mine own).

apoelstra:
Removed the input `nNonce` field and cleaned up the ambiguity that @roconnor-blockstream identified. Good to merge now?

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,64 @@
...
+*** ''sha_prevouts'' (32): the SHA256 of the serialization of all input outpoints. Each prevout output index is the masked version containing outpoint flags
+*** '''''NEW''''' ''sha_asset_amounts'' (32): the SHA256 of the serialization of ''nAsset||nValue'' for each output
+*** '''''NEW''''' ''sha_nonces'' (32): the SHA256 of the serialization of all output ''nNonce'' fields.
+*** ''sha_scriptpubkeys'' (32): the SHA256 of the serialization of all spent output ''scriptPubKey''s.
+*** ''sha_sequences'' (32): the SHA256 of the serialization of all input ''nSequence''.
+*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of all input ''assetIssuance'' or 130 '0x00' bytes for inputs with no issuance
+** If ''hash_type & 3'' does not equal <code>SIGHASH_NONE</code> or <code>SIGHASH_SINGLE</code>:
+*** ''sha_outputs'' (32): the SHA256 of the serialization of all outputs in sighash <code>CTxOut</code> format.
```
apoelstra, line 42: Whoops, this should not be "sighash `CTxOut` format", this one needs to include `nNonce`

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+* '''''NEW''''' Blockchain data: (Rationale: Signatures cannot be reused across different Elements instances even if pubkeys are reused.)
+** ''genesis_block'' (32): the hash of the blockchain's genesis block
+** ''genesis_block'' (32): the genesis hash again
+* Control:
+** ''hash_type'' (1).
+* Transaction data:
+** ''nVersion'' (4): the ''nVersion'' of the transaction.
+** ''nLockTime'' (4): the ''nLockTime'' of the transaction.
+** If the ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** '''''NEW''''' ''sha_outpoint_flags'' (32): the SHA256 of the serialization of the concatenation of one byte per input of the input's outpoint flags shifted right by 24 bits. (The byte for an pegin input would be 0x40. The byte for an issuance input would be 0x80. The byte for both a pegin and issuance would be 0xc0.)
```
apoelstra, line 34: I wonder if we should replace this "shifted by 24 bits" language with an explicit list of cases. In both Elements Core and rust-elements, the prevout stored internally does not actually have these bits set; they are extracted during deserialization and recorded separately.

roconnor-blockstream: However you want is fine.  I phrased it this way to be forward compatible with new flags that might be added in the future.

apoelstra: Yeah, that sounds reasonable. Let's keep the existing text.

doc/taproot-sighash.mediawiki
roconnor-blockstream: There is also variable length introduced by the presence/absence of the annex.  If you are interested in resolving that, then you should pad the annex hash in the case that there is no annex.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+
+==== Common signature message ====
+
+The function ''SigMsg(hash_type, ext_flag)'' computes the message being signed as a byte array. It is implicitly also a function of the spending transaction and the outputs it spends, but these are not listed to keep notation simple.
+
+The parameter ''hash_type'' is an 8-bit unsigned value. The <code>SIGHASH</code> encodings from the legacy script system are reused, including <code>SIGHASH_ALL</code>, <code>SIGHASH_NONE</code>, <code>SIGHASH_SINGLE</code>, and <code>SIGHASH_ANYONECANPAY</code>, plus the default ''hash_type'' value ''0x00'' which results in signing over the whole transaction just as for <code>SIGHASH_ALL</code>. The following restrictions apply, which cause validation failure if violated:
+* Using any undefined ''hash_type'' (not ''0x00'', ''0x01'', ''0x02'', ''0x03'', ''0x81'', ''0x82'', or ''0x83''<ref>'''Why reject unknown ''hash_type'' values?''' By doing so, it is easier to reason about the worst case amount of signature hashing an implementation with adequate caching must perform.</ref>).
```
roconnor-blockstream, line 13: we can remove the `<ref>...<ref>` here and below.  Alternative we could add a reference section below, but I don't think that is preferable.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+* Using <code>SIGHASH_SINGLE</code> without a "corresponding output" (an output with the same index as the input being verified).
+
+'''''NEW''''' If the input under consideration is a pegin input, the fields ''nAsset'', ''nValue'' and ''scriptPubKey'', where they appear, are taken from the pegin witness data. The asset used is the asset ID on the sidechain, not that on the parent chain (if any).
```
roconnor-blockstream, line 16: I think it is preferable for all these NEW statements to be their own paragraph.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+** ''genesis_block'' (32): the hash of the blockchain's genesis block
+** ''genesis_block'' (32): the genesis hash again
+* Control:
+** ''hash_type'' (1).
+* Transaction data:
+** ''nVersion'' (4): the ''nVersion'' of the transaction.
+** ''nLockTime'' (4): the ''nLockTime'' of the transaction.
+** If the ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** '''''NEW''''' ''sha_outpoint_flags'' (32): the SHA256 of the serialization of the concatenation of one byte per input of the input's outpoint flags shifted right by 24 bits. (The byte for an pegin input would be 0x40. The byte for an issuance input would be 0x80. The byte for both a pegin and issuance would be 0xc0.)
+*** ''sha_prevouts'' (32): the SHA256 of the serialization of all input outpoints. Each prevout output index is the masked version containing outpoint flags
```
roconnor-blockstream, line 35: masked version **excluding** outpoint flags.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+*** ''scriptPubKey'' (35): ''scriptPubKey'' of the previous output spent by this input, serialized as script inside <code>CTxOut</code>. Its size is always 35 bytes.
+*** ''nSequence'' (4): ''nSequence'' of this input.
+*** '''''NEW''''' ''asset_issuance'' (130): if ''outpoint_flag & 0x80 == 0x80'': asset issuance(fields for ''nInflationKeys'' and ''nAmount'' are serialized in fixed-length format) data of this input; otherwise 130 zero bytes
+** If ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** ''input_index'' (4): index of this input in the transaction input vector. Index of the first input is 0.
+** If an annex is present (the lowest bit of ''spend_type'' is set):
+*** ''sha_annex'' (32): the SHA256 of ''(compact_size(size of annex) || annex)'', where ''annex'' includes the mandatory ''0x50'' prefix.
+* Data about this output:
+** If ''hash_type & 3'' equals <code>SIGHASH_SINGLE</code>:
+*** ''sha_single_output'' (32): the SHA256 of the corresponding output in extended <code>CTxOut</code> format.
```
roconnor-blockstream, line 59: needs a (**NEW**) before the word extended.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+*** '''''NEW''''' ''sha_outpoint_flags'' (32): the SHA256 of the serialization of the concatenation of one byte per input of the input's outpoint flags shifted right by 24 bits. (The byte for an pegin input would be 0x40. The byte for an issuance input would be 0x80. The byte for both a pegin and issuance would be 0xc0.)
+*** ''sha_prevouts'' (32): the SHA256 of the serialization of all input outpoints. Each prevout output index is the masked version containing outpoint flags
+*** '''''NEW''''' ''sha_asset_amounts'' (32): the SHA256 of the serialization of ''nAsset||nValue'' for each output
+*** ''sha_scriptpubkeys'' (32): the SHA256 of the serialization of all spent output ''scriptPubKey''s.
+*** ''sha_sequences'' (32): the SHA256 of the serialization of all input ''nSequence''.
+*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of all input ''assetIssuance'' or 130 '0x00' bytes for inputs with no issuance
+** If ''hash_type & 3'' does not equal <code>SIGHASH_NONE</code> or <code>SIGHASH_SINGLE</code>:
+*** ''sha_outputs'' (32): the SHA256 of the serialization of all outputs in extended <code>CTxOut</code> format.
```
roconnor-blockstream, line 41: needs a (**NEW**) before the word extended.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,63 @@
...
+* Using <code>SIGHASH_SINGLE</code> without a "corresponding output" (an output with the same index as the input being verified).
+
+'''''NEW''''' If the input under consideration is a pegin input, the fields ''nAsset'', ''nValue'' and ''scriptPubKey'', where they appear, are taken from the pegin witness data. The asset used is the asset ID on the sidechain, not that on the parent chain (if any).
+'''''NEW''''' The epoch field prepended before signature hash is completely dropped. If there are new updates to the taproot signature hashes in elements, they will use new tagged hashes instead of incrementing epochs
+'''''NEW''''' The fields ''nNonce'', ''nAsset'' and ''nValue'' are serialized in a fixed-length format, which consists of their ordinary 1/9/33 byte encoding followed by sufficiently many 0 to pad the length out to 33. When we refer to ''extended <code>CTxOut</code> format'' below, we mean the fields ''nAsset'', ''nValue'', ''nNonce'' serialized in that order in fixed-length format, followed by the ordinary length-prefixed ''scriptPubKey''.
```
roconnor-blockstream, line 18: I might prefer a term like "padded CTxOut format" or something rather than "extended".  To me "extended" conveys that it has more fields.  However it isn't a big deal.

doc/taproot-sighash.mediawiki
apoelstra: Good catch, I'm sure that was my intent but I guess I didn't look at the rendered version too carefully.

apoelstra:
Added a commit to address all of @roconnor-blockstream comments

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,65 @@
...
+* Transaction data:
+** ''nVersion'' (4): the ''nVersion'' of the transaction.
+** ''nLockTime'' (4): the ''nLockTime'' of the transaction.
+** If the ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** '''''NEW''''' ''sha_outpoint_flags'' (32): the SHA256 of the serialization of the concatenation of one byte per input of the input's outpoint flags shifted right by 24 bits. (The byte for an pegin input would be 0x40. The byte for an issuance input would be 0x80. The byte for both a pegin and issuance would be 0xc0.)
+*** ''sha_prevouts'' (32): the SHA256 of the serialization of all input outpoints. Each prevout output index is the masked version excluding outpoint flags
+*** '''''NEW''''' ''sha_asset_amounts'' (32): the SHA256 of the serialization of ''nAsset||nValue'' for each output
+*** ''sha_scriptpubkeys'' (32): the SHA256 of the serialization of all spent output ''scriptPubKey''s.
+*** ''sha_sequences'' (32): the SHA256 of the serialization of all input ''nSequence''.
+*** '''''NEW''''' ''sha_issuances'' (32): the SHA256 of the serialization of the concatenation of all input ''assetIssuance'' or 130 '0x00' bytes for inputs with no issuance
```
roconnor-blockstream, line 41: s/(32):/(130):/

roconnor-blockstream: oops I am wrong.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,65 @@
...
+** If ''hash_type & 3'' does not equal <code>SIGHASH_NONE</code> or <code>SIGHASH_SINGLE</code>:
+*** ''sha_outputs'' (32): the SHA256 of the serialization of all outputs in '''NEW''' padded <code>CTxOut</code> format.
+*** '''''NEW''''' ''sha_output_witnesses'' (32): the SHA256 of the serialization of all output witnesses (rangeproof and surjection proof) in <code>CTxOutWitness</code> format.
+* Data about this input:
+** ''spend_type'' (1): equal to ''(ext_flag * 2) + annex_present'', where ''annex_present'' is 0 if no annex is present, or 1 otherwise (the original witness stack has two or more witness elements, and the first byte of the last element is ''0x50'')
+** If ''hash_type & 0x80'' equals <code>SIGHASH_ANYONECANPAY</code>:
+*** '''''NEW''''' ''outpoint_flag'' (1): the input's outpoint flags shifted right by 24 bits. (Compare ''sha_outpoint_flags'' above.)
+*** ''outpoint'' (36): the <code>COutPoint</code> of this input (32-byte hash + 4-byte little-endian) where the output index excludes the outpoint flags.
+*** '''''NEW''''' ''nAsset'' (33): (possibly confidential) assetID of the previous output spent by this input, in fixed-length format
```
roconnor-blockstream, line 50: slight preference for this to read ", in padded format".  And below too.

doc/taproot-sighash.mediawiki
```diff
@@ -0,0 +1,65 @@
...
+** If ''hash_type & 0x80'' does not equal <code>SIGHASH_ANYONECANPAY</code>:
+*** ''input_index'' (4): index of this input in the transaction input vector. Index of the first input is 0.
+** If an annex is present (the lowest bit of ''spend_type'' is set):
+*** ''sha_annex'' (32): the SHA256 of ''(compact_size(size of annex) || annex)'', where ''annex'' includes the mandatory ''0x50'' prefix.
+* Data about this output:
+** If ''hash_type & 3'' equals <code>SIGHASH_SINGLE</code>:
+*** ''sha_single_output'' (32): the SHA256 of the corresponding output in '''NEW''' padded <code>CTxOut</code> format.
+*** '''''NEW''''' ''sha_single_output_witness'' (32): the SHA256 of the serialization of the corresponding output witnesses (rangeproof and surjection proof) in <code>CTxOutWitness</code> format.
+
+The total length of ''SigMsg()'' is ''443'' bytes for <code>ANYONECANPAY</code> sighashes, ''366'' bytes for non-<code>ANYONECANPAY</code> sighashes, and both numbers are reduced by 64 bytes for <code>SIGHASH_NONE</code> sighashes. Note that this does not include the size of sub-hashes such as ''sha_prevouts'', which may be cached across signatures of the same transaction.
```
roconnor-blockstream, line 64: I counted 442 bytes.  Anyone want to count again?

sanket1729: I think this is not updated with the latest numbers after we removed `nNonce`. Recalculating the numbers

sanket1729: It's `410/334` for `ACP/non-ACP` according to my calculations/implementation. Are you taking into consideration the 32-byte annex too?

roconnor-blockstream: I was including the annex.

apoelstra:
As per offline conversation, added two commits to

* clarify the story with pegin inputs
* remove the padded fields, since we don't really need them in light of our planned transaction introspection opcodes

I haven't updated the total length arithmetic. Would probably be easiest to do this after it's implemented by poking at the Python implementation. @sanket1729 ?

apoelstra:
(fuzz test failure in PSBT module which will be shortly replaced; just kicked the CI job)

apoelstra:
Rebased. CI should pass now.

cc @roconnor-blockstream can I have an ack?

sanket1729:
>  haven't updated the total length arithmetic. Would probably be easiest to do this after it's implemented by poking at the Python implementation. @sanket1729 ?

@apoelstra 
https://github.com/sanket1729/elements/blob/ea3f5eda5caacc1e4a86908728119eac61a8187f/test/functional/test_framework/script.py#L865-L872

apoelstra:
Fixed the doc, but @sanket1729 I notice that your Python (and C++) impl completely ignore `SIGHASH_NONE`!

sanket1729:
> (out_type != SIGHASH_ALL and out_type != SIGHASH_SINGLE)

is the same of `out_type == SIGHHASH_NONE`

roconnor-blockstream:
LGTM

sanket1729 (approved):
ACK cca82245f83c3dd4af74137dd2d5d2faa2e35fc4

Issue #929, opened 2020-11-24
Assess current value of `g_con_elementsmode` and consider removal


apoelstra:
@instagibbs can you offer some insight into why we have this?

apoelstra:
Closing. In short, `g_con_elementsmode` is needed or else we would need to regenerate a huge proportion of the unit tests and functional tests, and continue to do this on an ongoing basis as Core adds news tests. I no longer consider removal of `g_con_elementsmode` to be a priority.

Issue #743, opened 2019-10-09
0.19 sync up checklist


instagibbs:
Running list before I forget:

[ ] - Core 0.19.0 release
[ ] - merge to the 0.19 branch off point of Core master
[ ] - fix tests: update bitcoin compatibility tests, fix functional/unit tests, make sure travis jobs match Core builds + extra ones, make sure all travis output looks correct
[ ] - make sure assets tutorial still works
[ ] - validate master tip: tests pass, manual re-sync of liquidv1 network
[ ] - branch off elements-0.19
[ ] - sync with Core 0.19 branch up to release commit
[ ] - Make RC
[ ] - elements-0.19 release!

RdeWilde:
How does this work? Is this manual comparison?

stevenroose:
No this is kind of the checklist we used for the 0.18 sync up. We do all the steps when we want to sync up with an upstream release. We're definitely overdue syncing with 0.19.

apoelstra:
Closing, as we have a different merge/branch strategy for releases now.

Issue #527, opened 2019-03-21
Fix bumpfee


instagibbs:
Currently not working. 

Probably will backport https://github.com/bitcoin/bitcoin/pull/15557 since the code is significantly easier to understand and directly uses `CWallet::CreateTransaction`

PR #935, opened 2020-12-03
Bring Elements up to date with Bitcoin Core 0.21

apoelstra:
Requires a bunch of followup work which I will open other issues for. See also #743 which was our checklist for 0.19.

Then in followup PRs:

* ~Fix the fuzz tests; I did not test these at any point throughout the rebase https://github.com/ElementsProject/elements/issues/936~
* Update and publish the `merge-prs` script to run the fuzzer, linter etc., as well as do a full clean `make check` https://github.com/ElementsProject/elements/pull/923/
* Audit translations; at a couple points I took Core's translations wholesale and we will need to re-edit them to change Bitcoin to Elements https://github.com/ElementsProject/elements/issues/937
* Audit the GUI. I have not opened the bitcoin GUI since 0.12 and not opened Elements ever, so probably I broke this. https://github.com/ElementsProject/elements/issues/938
* Assess wallet/node coupling; Core did some separation in the build system which I undid, to support the `signblock` RPC which involves both wallet and node logic. Core's solution to this for Signet may be to introduce a separate signblock tool which we may want to also do; see #bitcoin-core-dev IRC logs for the meeting on 2020-12-03. https://github.com/ElementsProject/elements/issues/939
* Assess functional tests: chase down a couple intermittent failures, assess the new broken `loadblocks` test, assess the "extended tests" which are broken on master https://github.com/ElementsProject/elements/issues/940
* Assess bitcoin functional tests and `g_con_elementsmode` #929 
* Strengthen the functional tests; add more thorough PAK checks (see #926), move `computefastmerkleroot` into Python rather than using the RPC; https://github.com/ElementsProject/elements/issues/941
* Fix the `bumpfee` RPC to support CT. There is an existing issue for this #527 
* Fix the PAK standardness bug #925 which was discovered through the rebase.
* Followup on conversion to `bilingual_str` from `std::string` for many Core RPCs; we should follow suit for our own RPCs https://github.com/ElementsProject/elements/issues/942
* The new `avoid_reuse` param to `sendtoaddress` and a couple other RPCs moves some Elements fields out of the way. For backward compatibility we should put `avoid_reuse` after these Elements fields. (I think this is the only time I did this; after talking to Glenn I tried to make future RPC extensions backward compatible)
* Rebase #900 on this and un-disable the large number of lines in the PSET functional tests which are disabled (in particular around the missing `fee` field of the RPC output)

Then a couple large followups. Tracking issues:
* Taproot: https://github.com/ElementsProject/elements/issues/943
* Descriptors: https://github.com/ElementsProject/elements/issues/944
* Simplicity: https://github.com/ElementsProject/elements/issues/945


stevenroose:
Some additional things that I did after the last catchups: 

- copy `./src/primitives` from core into our `src/primitives/bitcoin`
- copy `./test/functional` from core into out `./test/bitcoin_functional` and reinstate all `ELEMENTS:` modifications to that directory

stevenroose:
Also, should we have a `dev-catchup` branch for this until all above issues are resolves so we don't break master? (Functional and CI breaking instead of compilation breaking.)

apoelstra:
* `./src/primitives/bitcoin` is up to date. I made sure it was in sync throughout the rebase.
* As for `test/bitcoin_functional`, this will be very annoying. Core more than doubled the number of functional tests between 0.18 and 0.21 so there will be a ton of new `ELEMENTS:` modifications to add for very unclear value.

gwillen:
@stevenroose I don't think we need a new branch, it seems like we want to just get all this stuff working on andrew's branch before we merge.

gwillen:
Things I think we're consciously punting for now:
- translations (I think we've always punted those, we should probably be paying Transifex or whatever Core does, so we can have translations for Elements-QT, but we can do that later.)
- bitcoin_functional tests (which it seems like should just go away, along with g_con_elementsmode)
- fuzz tests (which apparently are slightly complex to get running, and don't run in CI anyway)
- at least one more thing I'm forgetting.

src/rpc/rawtransaction.cpp
```diff
@@ -2689,13 +2689,13 @@ static RPCHelpMan rawreissueasset()
             throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Invalid asset address provided: %s", asset_address_uni.get_str()));
         }
 
-        size_t input_index = -1;
+        int input_index = -1;
         const UniValue& input_index_o = issuance_o["input_index"];
         if (input_index_o.isNum()) {
             input_index = input_index_o.get_int();
```
dgpv, line 2695: no need to initialize `input index` above, can just `int input_index = input_index_o.get_int();`

dgpv: (why I'm commenting here at all? Just being curious what code ubsan complains about...)

apoelstra: Lol! Good catch, if I need to touch this code again, I'll use your suggestion. (Or, once I get CI working, if I go back to clean things up..)

gwillen:
In reviewing the merges using my script from #954, I found something that could be an error in the merge 19568c38916bc22f898350b85624345e5ad61c3d . (This is the first one I've found so far! Very exciting.)

The bitcoin side, c6e42f1ca9c8a2eeb315c68e66dc7e9e8153882a (https://github.com/bitcoin/bitcoin/pull/14193) adds the `EXCLUSIVE_LOCKS_REQUIRED(cs)` annotation to 4 mempool methods. One of those, `removeForReorg`, appears to have been dropped in the merge (probably because that method already had a different lock annotation on it, so the change was only 3 characters.)

merge base, "our side", and final merged version:

`void removeForReorg(const CCoinsViewCache *pcoins, unsigned int nMemPoolHeight, int flags) EXCLUSIVE_LOCKS_REQUIRED(cs_main);`

"their side":

`void removeForReorg(const CCoinsViewCache* pcoins, unsigned int nMemPoolHeight, int flags) EXCLUSIVE_LOCKS_REQUIRED(cs, cs_main);`

gwillen:
Found what seems to be another merge conflict resolution issue in 0e2c963dbfedbe5c22f08e9f1aecc62882bfdbea, which is the very next conflicted merge. An `|| ShutdownRequested()` got dropped from a conditional, which I think needs to still be there. ("Our side's" change here is just to push the whole block-generation PoW stuff inside an `if (!g_signed_blocks) {` conditional, so I expect that the contents of "their side's" changes should be otherwise used unaltered.)

apoelstra:
@gwillen regarding the locking error, we **do** actually have CI lints for this now! So this is fixed somewhere in the final 20 commits of the PR where I was fixing things for CI. But great find, agreed this gives us some confidence in the review process. I doubt we would've been able to find this with our previous rebasing workflows.

apoelstra:
The `ShutdownRequested` thing looks real, good catch! Will fix it in a followup PR or by adding a commit to this one. For what it's worth, it looks like this codepath is only accessed if you start the daemon in bitcoin mode (lol) and then run the CPU miner (lol) and then try to Ctrl+C it.

jhfrontz:
Under guidance from @gwillen, something looks odd in 3a0de44d90bf8a01d3f7431fd849e8b63b1e951f (maybe some dead code?).

gwillen:
To elaborate -- there is at least some dead/duplicate code here, where `txout` gets set above the loop, pointlessly, then set again inside the loop. It seems harmless, but it suggests that something weird happened in the merge, and some of the other stuff in the diff is kind of hard to read. So probably it merits (1) fixing the mismerge in a followup, and (2) double-checking that the rest of the merge was correct.

gwillen:
In `9c3480fea96abe7cc1adbdb2c754cce0224b23fa`:

possibly non-nit, not sure what the deal is here:
* in `chainparamsbase.cpp`, the third argument to `AddArg` is now almost always an enum, but for `con_genesis_style` it's still a bool (I'm surprised it compiles.)

nits: some elements fixes to help strings got accidentally reverted:
* in `bitcoin-cli.cpp`, in "needs to exactly match corresponding -wallet option passed to bitcoind" -> "bitcoind" change to say "the daemon" got reverted
* in `bitcoin-tx.cpp`, "outaddr=VALUE:ADDRESS" should be "outaddr=VALUE:ADDRESS(:ASSET)"

gwillen:
9cc26a716abffda6ce20dd057cfb59fd8c36a46a has a FIXME, which you already mentioned in the list above. However, I'd also like it if a few more eyes looked at it, since it's a huge change (all of dynafed). In particular, the resolution of the refactor involving `SignTransaction` and `ConstructTransaction`, which looks like it involved some manual work; and the removal of what _seems_ to just be a debug `print` in `feature_blocksign.py` (search for `generated key`), the removal of which is presumably fine but was not actually part of either input to the merge.

gwillen:
f89b990b780aa7063a46c244aed22b119113e338 brings in some serialization changes, and contains a bunch of deep, apparently-new modifications for elements, that really need to be reviewed as new code, which I don't have the brainpower for right now so I'm skipping over it.

gwillen:
In c3b6bbba4aa039d6fbec6bb89f90aaa27882c2ae you accidentally double-scoped the scoped enum in one of the uses that had to be fixed up:

`QString amount = GUIUtil::formatAssetAmount(rcp.asset, rcp.asset_amount, bitcoin_unit, BitcoinUnits::BitcoinUnits::SeparatorStyle::STANDARD, true);`

`BitcoinUnits::BitcoinUnits::SeparatorStyle::STANDARD` should be `BitcoinUnits::SeparatorStyle::STANDARD`
I'm kind of surprised that compiles, actually. Does the name of a scoped enum exist within its own scope?

gwillen:
In 4839db857d5fbe2a27ebabb7bf5bf9d59ef244e9, the PSBT-related menu items had been changed to "PSET" in Elements, but accidentally got changed back to saying "PSBT".

gwillen:
There's a FIXME in 8e770d3f211b5718a08457d25f9b82aa261efbfc which I don't see mentioned in the PR notes:

`ChainstateManager& chainman = g_chainman; // FIXME avoid using this global, see #19413`

gwillen:
Nit: in 5e62edcf8e8beb2a1a7760d6cf015effa6da0a78, the error message for when the change address field is neither a string nor a map became less helpful, and changed error code, and it's hard to tell whether it was on purpose.

`throw JSONRPCError(RPC_INVALID_PARAMETER, "changeAddress must be either a map or a string");`

vs

`throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Change address must be a valid address");`

(The upstream change was that the parameter can now be either `changeAddress` or `change_address`, which makes the wording of the error here a little awkward; upstream doesn't have this case because it's always a string there.)

gwillen:
The size math in 6682a291c24db71c140f7490063a6c562aaf8662 is pretty messy, and contains:

```
// Elements: Hack to get QR address to print right
const size_t MORE_WIDTH = 80;
```

I didn't see any special comment from andrew in the merge, and I'm suspicious that, if this was not manually eyeball-tested, there is a good chance the resulting QR code comes out wrong, or at least weird.

This probably isn't critical, but I have no idea what these QR codes are used for.

gwillen:
9e4db637d817f58e9049b0e275c5a19005ab2944 is a dramatic change to the way version numbers are used in Core (from 0.22.0 to 22.0.0). We should check how this interacts with packaging and such for us (e.g. we have various version number checks in scripts that assume very large version numbers are the old liquid-daemon package, and small reasonable version numbers are the new elements package, which may no longer hold depending on the effects of this bump.)

gwillen:
Exciting: I just reviewed up through 7a8ccee15512fc7c4c40f6d8a107d3c13271f901 ! Aka `merged-master`, the last commit merged into this branch, before the non-merge fixups.

@apoelstra I imagine the next thing is to review the non-merge fixup commits? Do you have any other thoughts on how to proceed further? (Have you looked over my various intermediate comments?)

test/sanitizer_suppressions/ubsan
```diff
@@ -80,5 +80,6 @@ implicit-unsigned-integer-truncation:leveldb/*
 implicit-integer-sign-change:crc32c/*
 
 implicit-signed-integer-truncation:script/interpreter.cpp
+implicit-integer-sign-change:primitives/block.h
```
stevenroose, line 83: what was in here that was causing issues?

ci/test/00_setup_env_mac_host.sh
```diff
@@ -9,7 +9,8 @@ export LC_ALL=C.UTF-8
 export HOST=x86_64-apple-darwin18
 export PIP_PACKAGES="zmq"
 export GOAL="install"
-export BITCOIN_CONFIG="--with-gui --enable-reduce-exports --enable-werror --with-boost-process"
+# ELEMENTS: add -fno-stack-check to work around clang bug on macos
+export BITCOIN_CONFIG="--with-gui --enable-reduce-exports --enable-werror --with-boost-process CXXFLAGS=-fno-stack-check"
```
stevenroose, line 13: hmm, this should be temporary, right?

stevenroose (commented):
So what is left for getting this merged? Would you have to continue use the script to get current progress on master into this branch to get rid of merge conflicts?

test/lint/commit-script-check.sh
```diff
@@ -35,7 +35,7 @@ for commit in $(git rev-list --reverse $1); do
             git --no-pager diff --exit-code $commit && echo "OK" || (echo "Failed"; false) || RET=1
         fi
         git reset --quiet --hard HEAD
-     else
+     elif [ "$commit" != "f471a3be00c2b6433b8c258b716982c0539da13f" ]; then
```
gwillen, line 38: Looks good, but I'm curious why this doesn't hit upstream. (It's an upstream commit that has the issue, yeah?)

src/rpc/rawtransaction.cpp
```diff
@@ -2208,7 +2208,7 @@ static RPCHelpMan rawblindrawtransaction()
                     },
                     {"inputamounts", RPCArg::Type::ARR, RPCArg::Optional::NO, "An array with one entry per transaction input.",
                         {
-                            {"inputamount", RPCArg::Type::AMOUNT, RPCArg::Optional::NO, "An amount for each input."},
+                            {"ixxnputamount", RPCArg::Type::AMOUNT, RPCArg::Optional::NO, "An amount for each input."},
```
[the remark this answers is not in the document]
gwillen: This must be a typo? Looks like what happens when I accidentally have the cursor active my editor and don't notice where my stray keypresses went.

src/test/fuzz/checkqueue.cpp
```diff
@@ -39,13 +39,13 @@ void test_one_input(const std::vector<uint8_t>& buffer)
...
-    std::vector<DumbCheck> checks_1;
-    std::vector<DumbCheck> checks_2;
+    std::vector<DumbCheck*> checks_1;
+    std::vector<DumbCheck*> checks_2;
     const int size = fuzzed_data_provider.ConsumeIntegralInRange<int>(0, 1024);
     for (int i = 0; i < size; ++i) {
         const bool result = fuzzed_data_provider.ConsumeBool();
-        checks_1.emplace_back(result);
-        checks_2.emplace_back(result);
+        checks_1.emplace_back(new DumbCheck(result));
```
gwillen, line 47: Is this leak unproblematic? How many times does this run? (Obviously if it runs once per execution of the test binary it's fine, but it looks like it may run repeatedly.)

If nothing else, at least comment explaining why the leak is okay. (I'm curious why the original code had to be changed in the first place.)

test/lint/lint-python-mutable-default-parameters.sh
```diff
@@ -8,7 +8,7 @@
 
 export LC_ALL=C
 EXIT_CODE=0
-OUTPUT=$(git grep -E '^\s*def [a-zA-Z0-9_]+\(.*=\s*(\[|\{)' -- "*.py")
+OUTPUT=$(git grep -E '^\s*def [a-zA-Z0-9_]+\(.*=\s*(\[|\{)' -- "*.py" ':!*/bitcoin_functional/*')
```
gwillen, line 11: Are we planning to just nuke the entire `bitcoin_functional` directory? Can we just do that here instead?

test/sanitizer_suppressions/ubsan
```diff
@@ -78,3 +78,7 @@ implicit-signed-integer-truncation:torcontrol.cpp
 implicit-unsigned-integer-truncation:crypto/*
 implicit-unsigned-integer-truncation:leveldb/*
 implicit-integer-sign-change:crc32c/*
+
+implicit-signed-integer-truncation:script/interpreter.cpp
```
gwillen, line 82: Man, is ubsan really this overactive? I see that a bunch of other exceptions in this file already exist, suggesting that it is, but it scares me to suppress things without very careful review of exactly what we're suppressing, line-by-line. I'm surprised we do exceptions for entire files at a time. (Are these specific checks just broken in ubsan?)

src/Makefile.am
```diff
@@ -26,7 +26,7 @@ BITCOIN_INCLUDES += $(UNIVALUE_CFLAGS)
 
 LIBBITCOIN_SERVER=libbitcoin_server.a
 LIBBITCOIN_COMMON=libbitcoin_common.a
-LIBBITCOIN_CONSENSUS=libbitcoin_consensus.a
+LIBBITCOIN_CONSENSUS=libelements_consensus.a
```
gwillen, line 29: I mentioned elsewhere that I think we may actually want to _reduce_ the number of non-user-visible places that we rename things from bitcoin to elements in the build stuff. I guess the consensus library is in theory a build product that is user-visible, which is why to rename that, but not the stuff above and below it?

Making it easier to understand what's going on is not a bad goal, but it might be in tension here with reducing the total amount of divergence with upstream.

src/test/fuzz/checkqueue.cpp
gwillen: Ah, it looks like this got fixed in a later commit.

src/rpc/rawtransaction.cpp
```diff
@@ -2689,13 +2689,13 @@ static RPCHelpMan rawreissueasset()
             throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Invalid asset address provided: %s", asset_address_uni.get_str()));
         }
 
-        size_t input_index = -1;
+        int input_index = -1;
```
gwillen, line 2692: May as well use the fix from dgpv's old comment here, and push the declaration of input_index inside the conditional so we don't need to separately initialize it at all.

test/lint/lint-python-mutable-default-parameters.sh
stevenroose: Hmm, I don't really know. I understand the original motivation. But it's also annoying to maintain and maybe it doesn't really make sense to try and keep supporting running in Bitcoin mode.

gwillen:
ok, I've reviewed all the fixup commits -- anything I didn't comment on looks good to me.

test/sanitizer_suppressions/ubsan
apoelstra: Yes they are just broken. chars promoted to ints because C++ is bullshit, then "truncated" back within the same expression, etc.

test/sanitizer_suppressions/ubsan
apoelstra: I don't remember. Probably
```
    static const uint32_t DYNAFED_HF_MASK = 1 << 31; 
```
though maybe other things.

src/rpc/rawtransaction.cpp
apoelstra: I think this fix matches what Core does.

gwillen: (If you want to leave it as-is, that's also fine.)

src/rpc/rawtransaction.cpp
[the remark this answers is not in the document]
apoelstra: yeah looks like a typo. i'll fix it.

ci/test/00_setup_env_mac_host.sh
apoelstra: Yes, just until Cirrus updates their MacOS instances.

test/lint/commit-script-check.sh
apoelstra: Because it was merged into upstream before this check script was.

gwillen: (This is fine to ignore if you don't remember.)

test/lint/lint-python-mutable-default-parameters.sh
gwillen: For the sake of getting this in, let's leave this the way it is, and nuke bitcoin_functional as a followup.

src/Makefile.am
gwillen: (For the sake of getting this in, fine to leave it if you prefer.)

apoelstra:
Rebased. Removed the bitcoin PRs that were from past the 0.21 fork point, and added the most recent 6 PRs from Elements. Also removed the extraneous `xx` thing from the fixups.

Everything from 22cf380984d (Elements #960 is new).

gwillen:
It looks like the last commit that merges a PR into the new branch is 0699c4d5aa2bd30ffe1e5c535e8831a5af60e24b, and the fixups are after that. I'm going to create a `merged-master` branch on this repo, pointing there. (I was using the one on the `apoelstra` repo before, but it looks like you didn't update it when you did more merging, and anyway we should have it in the main repo.)

gwillen:
Looking at the newly-merged commits: copy-paste error in d89112072cd6a4cbbd40a6ca3b44018c5992b66b:

```
             if (!proposed.m_fedpeg_program.IsWitnessProgram(fedpeg_version, fedpeg_program)) {
-                return state.Invalid(false, REJECT_INVALID, "invalid-dyna-fed", "proposed fedpegs program must be native segwit scriptPubkey");
+                return state.Invalid(BlockValidationResult::BLOCK_CONSENSUS, "invalid-dyna-fed", "proposed signblockscript must be native segwit scriptPubkey");
             }
```

(New error message refers to `signblockscript` instead of fedpeg script. (I assume "fedpegs program" was a typo.))

I think it's your call whether you want to make this a fixup commit, or try to fix it in-place, given that it's fairly shallow at this point. (This would require rebasing the fixups, obviously, but the rebase should be trivial.)

I think it ought to be possible to fix this in-place with an interactive rebase, without having to redo any of the subsequent merges, as long as the fix doesn't overlap with any of them. This might be worth trying as an experiment, just to figure out whether we can, but I'm also fine with leaving that for future work.

gwillen:
For the benefit of the peanut gallery: andrew is seeing a test error that doesn't reproduce locally, so there will be some noise trying to pin it down in CI.

stevenroose:
I looked at all the recent changes up until 291da8bedb533e511090fbd7c0943175d99c9fc2 and can utACK that. Good to see CI entirely working!

It's still all a bit magical, I'll go over the list of my "just check on that to make sure nothing is accidentally broken" items after this is merged.

If you can ack, I can merge it, @gwillen .

gwillen:
utACK from me!

With the understanding that there's going to be a bunch of followup work, of course.

PR #949, opened 2020-12-10
[Doc] New Opcodes for tapscript

sanket1729:
This PR adds a doc describing suggested new opcodes for better script programming with elements. The purpose of these opcodes to aid easier writing of complex scripts like covenants in elements.

This introduces 1) Opcodes for Streaming hashes, 2) Transaction introspection opcodes 3) 64 byte athematic opcodes 4) Some utility conversion opcodes 5) Crypto opcodes for low-level operations and tweaking 6) A cleaner way to write For loops and 7) Changes to existing opcodes. 

The doc also describes a brief rationale for choice. Feedback welcome :)

### DIFF
--- a/doc/tapscript_opcodes.md
+++ b/doc/tapscript_opcodes.md
@@ -1,0 +1,96 @@
+This document proposes new opcodes to be added to the elements network along with the taproot upgrade. The new tapscript `OP_SUCCESS` opcodes allow introducing new opcodes more cleanly than through `OP_NOP`. In this document, we propose modifying the following `OP_SUCCESS`
+to have the additional semantics. We use opcodes serially `OP_SUCCESS196`, `197`... in order
+to avoid conflict with bitcoin potentially using `OP_SUCESSSx`(assuming bitcoin uses those serially based on availability). The initial version of this document had additional opcodes(`OP_FOR`, multi-byte opcodes) has since been updated to this current version in favor of application complexity.
+
+# Resource Limits
+
+## Changes in Taproot(Including Standardness Policy Changes)
+Taproot already increases a lot of resource limitations from segwitv0, so there is no additional need to alter any of those. In particular, from BIP 342
+
+- Script size limit: the maximum script size of 10000 bytes does not apply. Their size is only implicitly bounded by the block weight limit.
+- Non-push opcodes limit: The maximum non-push opcodes limit of 201 per script does not apply.
+- Sigops limit: The sigops in tapscripts do not count towards the block-wide limit of 80000 (weighted). Instead, there is a per-script sigops budget. The budget equals 50 + the total serialized size in bytes of the transaction input's witness (including the CompactSize prefix). Executing a signature opcode (`OP_CHECKSIG`, `OP_CHECKSIGVERIFY`, or `OP_CHECKSIGADD`) with a non-empty signature decrements the budget by 50. If that brings the budget below zero, the script fails immediately.
+- Stack + altstack element count limit: The existing limit of 1000 elements in the stack and altstack together after every executed opcode remains. It is extended to also apply to the size of the initial stack.
+- Stack element size limit: The existing limit of maximum 520 bytes per stack element remains, during the stack machine operations. There is an additional policy rule limiting the initial push size to `80` bytes.
+
+## Additional resource limits changes in Elements
+
+- New added opcodes `OP_MULTISCALAREXPVERIFY` for `k` base multi scalar exponentiation is counted as `50*k` units towards the SIGOPS budget. If the operation requires extra script_budget, the user must add additional witness elements to make sure that the script executes within the desired budget.
+
+# New Opcodes for additional functionality:
+
+1. **Streaming Opcodes for streaming hashes**: There is an existing limitation of `MAX_SCRIPT_ELEMENT_SIZE`(520 bytes) because of which we cannot operate hash functions like `OP_SHA256` on messages more than 520 bytes. This allows hashing on more than 520 bytes while still preserving the existing security against resource exhaustion attacks. The proposal for this is still under discussion in https://github.com/ElementsProject/elements/pull/817. We suggest the latest scheme suggested by Russell O'Connor
+   -  Define `OP_SUCCESS196` as `OP_SHA256INITIALIZE` which pops a bytestring and push SHA256 context creating by adding the bytestring to the initial SHA256 context.
+   - Define `OP_SUCCESS197` as `OP_SHA256UPDATE` which pops 1) a SHA256 context and 2) bytestring and push an updated context by adding the bytestring to the data stream being hashed.
+   - Define `OP_SUCCESS198` as `OP_SHA256FINALIZE` to pop a SHA256 context and bytestring and push a SHA256 hash value after adding the bytestring and completing the padding.
+
+
+2. **Transaction Introspection codes**: Transaction introspection is already possible in elements script by use of `OP_CHECKSIGFROMSTACKVERIFY`, however the current solutions are really expensive in applications like [covenants](https://github.com/sanket1729/covenants-demo). Therefore, we are not adding any new functionality by supporting introspection, only making it easier to use. The warning still remains the same as with covenants, if the user is inspecting data from parts of the transaction that are not signed, the script can cause unexpected behavior.
+For opcodes that inspect data that is not committed in sighash, introspection is safe because any changes to the witness data would cause wtxid to change and it would revalidate the tx again.
+   - Define `OP_SUCCESS199` as `OP_INSPECTINPUT` with the following semantics: Pop two minimal `CScriptNum`s from stack. The first number `n` denotes the type of introspection, and the second number `idx `denotes the position of input to inspect. Immediately abort if `idx` is out of bounds.
+      1. If `n=0`, `OP_INSPECTINPUTSPENDTYPE` Push 1 byte "spend_type" onto the stack. spend_type (1) is equal to `(ext_flag * 2) + annex_present` as defined in [Modified BIP-341 SigMsg for Elements](https://gist.github.com/roconnor-blockstream/9f0753711153de2e254f7e54314f7169)
+      2. If `n=1`, `OP_INSPECTINPUTOUTPOINTFLAG` Push the outpoint_flag(1) as defined in [Modified BIP-341 SigMsg for Elements](https://gist.github.com/roconnor-blockstream/9f0753711153de2e254f7e54314f7169)
+      3. if `n=2`, `OP_INSPECTINPUTOUTPOINT` Push the outpoint as a tuple. First push the `txid`(32) of the `prev_out`, followed by a 4 byte push of `vout`
+      4. If `n=3`, `OP_INSPECTINPUTASSET` Push the `nAsset` onto the stack as two elements. The first push the assetID(32), followed by the prefix(1)
+      5. If `n=4`, `OP_INSPECTINPUTVALUE` Push the `nValue` as a tuple, value(8, 32) followed by prefix,
+      6. If `n=5`, `OP_INSPECTINPUTSSCRIPTPUBKEY` Push the scriptPubkey(35) onto the stack.
+      7. If `n=6`, `OP_INSPECTINPUTSEQUENCE` Push the `nSequence`(4) as little-endian number.
+      8. If `n=7`, `OP_INSPECTINPUTISSUANCE` Push the assetIssuance information(74-130) if the asset has issuance, otherwise push an empty vector
+      9. If `n=8`, `OP_INSPECTINPUTANNEX` Push the annex onto the stack where the annex includes the prefix(0x50). If the annex does not exist, push an empty vector
+      10. Otherwise fail.
+   - Define `OP_SUCCESS200` as `OP_INSPECTCURRENTINPUT` that pushes the current input index(4) as little-endian onto the stack
+   - Define `OP_SUCCESS201` as `OP_INSPECTOUTPUT` with the following semantics:
+      - Pop the stack pop as minimal `CScriptNum` as `n`. Next, pop another element as minimal `CScriptNum` input index `idx`.
+      1. If `n=0`, `OP_INSPECTOUTPUTASSET` Push the `nAsset` as a tuple, first push the assetID(32), followed by the prefix(1)
+      2. If `n=1`, `OP_INSPECTOUTPUTVALUE` Push the `nValue` as a tuple, value(8, 32) followed by prefix
+      3. If `n=2`, `OP_INSPECTOUTPUTNONCE` Push the `nNonce` as a tuple, nonce(32, 0) followed by prefix. Push empty vector for `None` nonce
+      4. If `n=3`, `OP_INSPECTOUTPUTSCRIPTPUBKEY` Push the scriptPubkey(35).
+      5. Otherwise, fail
+   - Define `OP_SUCCESS202` as `OP_INSPECTTX` with the following semantics:
+      - Pop the stack pop as minimal `CScriptNum` as `n`.
+      1. If `n=0`, `OP_INSPECTVERSION` Push the nVersion(4) as little-endian.
+      2. If `n=1`, `OP_INSPECTLOCKTIME` Push the nLockTime(4) as little-endian.
+      3. If `n=2`, `OP_INSPECTNUMINPUTS` Push the number of inputs(4) as little-endian
+      4. If `n=3`, `OP_INSPECTNUMOUTPUTS` Push the number of outputs(4) as little-endian
+      5. If `n=4`, `OP_TXSIZE` Push the transaction size in vbytes (4) as little-endian
+      6. Otherwise, abort
+
+5. **Crypto**: In order to allow more complex operations on elements, we introduce the following new crypto-operators.
+   - Define `OP_SUCCESS203` as `OP_ECMULSCALAREXPVERIFY`, pop the top element as `k`. Then pop next elements as points `G1`(first), `G2`(second)
+      ..`Gk`. Next pop `k` scalars `x1`, `x2`... `xk`. Finally pop result element as as point `Q`. Assert `x1G1+x2G2+x3G3.. xkGk == Q` This counts as `k*50` towards budget sigops. If any of `G_i`,`Q` is point at infinity fail.
+   - Define `OP_SUCCESS204` as `OP_TAPTWEAK` with the following semantics. Pop the first element as point `P`, second element as script blob `S`. Push the Taptweak on the top of stack `Q = P + H(P||S)*G`. If `|S| > MAX_ELEMENT_SIZE`, the user should use the streaming opcodes to compute the Hash function.
+
+3. **Signed 64-bit arithmetic opcodes:** Current operations on `CScriptNum` as limited to 4 bytes and are difficult to compose because of minimality rules. having a fixed width little operations with 8 byte signed operations helps doing calculations on amounts which are encoded as 8 byte little endian.
+   - When dealing with overflows, we explicitly return the success bit as a `CScriptNum` at the top of the stack and the result being the second element from the top. If the operation overflows, first the operands are pushed onto the stack followed by success bit. \[`a_second` `a_top`\] overflows, the stack state after the operation is \[`a_second` `a_top 0`\] and if the operation does not overflow, the stack state is \[`res 1`\].
+   - This gives the user flexibility to deal if they script to have overflows using `OP_IF\OP_ELSE` or `OP_VERIFY` the success bit if they expect that operation would never fail.
+When defining the opcodes which can fail, we only define the success path, and assume the overflow behavior as stated above.
+   - Define `OP_SUCCESS205` as `OP_ADD64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push a + b onto the stack. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS206` as `OP_SUB64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push a - b onto the stack. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS207` as `OP_MUL64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push `a*b` onto the stack. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS208` as `OP_DIV64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). First push remainder `a%b`(must be non-negative and less than |b|) onto the stack followed by quotient(`a//b`) onto the stack. Abort if `b=0`. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS209` as `OP_LESSTHAN64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a < b`.
+   - Define `OP_SUCCESS210` as `OP_LESSTHANOREQUAL64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a <= b`.
+   - Define `OP_SUCCESS211` as `OP_GREATERTHAN64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a > b`.
+   - Define `OP_SUCCESS212` as `OP_GREATERTHANOREQUAL64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a >= b`.
+   - Define `OP_SUCCESS213` as `OP_EQUAL64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a == b`.
+   - Define `OP_SUCCESS214` as `OP_AND64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a & b`.
+   - Define `OP_SUCCESS215` as `OP_OR64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a | b`.
+   - Define `OP_SUCCESS216` as `OP_XOR64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a ^ b`.
+   - Define `OP_SUCCESS217` as `OP_NOT64`(cannot fail!): pop the first number(8 byte LE) as `a`. Push `~a`.
+   - Define `OP_SUCCESS218` as `OP_LSHIFT`: pop the first number as `CScriptNum` `l`(abort if l < 0 or l > 63) followed another pop for `a` (8 byte LE). Push `a << l` preserving the sign bit. `(-1 << 3) = - 8` returns fixed 64 bits, extra-bits are discarded and sign is preserved.
+   - Define `OP_SUCCESS219` as `OP_RSHIFT`: pop the first number as `CScriptNum` `r`(abort if r < 0 or r > 63) followed another pop for `a` (8 byte LE). Push `a >> r`.(Sign bit is preserved).
+
+4. **Conversion opcodes:** Methods for conversion from `CScriptNum` to `8-byte LE`, `4-byte LE`.
+   - Define `OP_SUCCESS220` as `OP_SCIPTNUMTOLE64`: pop the stack as minimal `CSciptNum`, push 8 byte signed LE corresponding to that number.
+   - Define `OP_SUCCESS221` as `OP_LE64TOSCIPTNUM`: pop the stack as a 8 byte signed LE. Convert to `CScriptNum` and push it, abort on fail.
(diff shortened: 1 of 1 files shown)


doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,134 @@
...
+
+1. **Streaming Opcodes for streaming hashes**: There is an existing MAX_SCRIPT_ELEMENT_SIZE is 520 bytes, that is the maximum message size that can OP_SHA256 can operate on. This allows hashing on more than 520 bytes while still preserving the existing security against resource exhaustion attacks. The proposal for this is still under discussion in https://github.com/ElementsProject/elements/pull/817. We suggest the latest scheme suggested by Russel O Connor
+   -  Define OP_SUCCESS200 as a multi-byte opcode OP_STREAMINGHASH with the following semantics. Execute the push opcode following OP_STREAMINGHASH, if the opcode is not a push opcode, fail.
+   -  Pop the stack pop as minimal CScriptNum as n.
```
dgpv, line 28: "Pop" used twice

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,134 @@
...
+      7. If n=6, interpret `OP_HASH256INITIALIZE` to pop a bytestring and push HASH256 context creating by adding the bytestring to the initial HASH256 context.
+      8. If n=7, interpret as `OP_HASH256UPDATE` to pop a HASH256 context and bytestring and push an updated context by adding the bytestring to the data stream being hashed.
+      9. If n=8, interpret as `OP_HASH256FINALIZE` to pop a HASH256 context and bytestring and push a HASH256 hash value after adding the bytestring and completing the padding.
+      10. If n=9, interpret `OP_HASH160INITIALIZE` to pop a bytestring and push HASH160 context creating by adding the bytestring to the initial HASH160 context.
+      11. If n=10, interpret as `OP_HASH160UPDATE` to pop a HASH160 context and bytestring and push an updated context by adding the bytestring to the data stream being hashed.
+      12. If n=11, interpret as `OP_HASH160FINALIZE` to pop a HASH160 context and bytestring and push a HASH160 hash value after adding the bytestring and completing the padding.
```
dgpv, line 40: What is the context when hash algorithms besides SHA256 would be needed ?

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,134 @@
...
+      6. If `n=5`, `OP_INSPECTINPUTSSCRIPTPUBKEY` Push the scriptPubkey(35) onto the stack.
+      7. If `n=6`, `OP_INSPECTINPUTSEQUENCE` Push the `nSequence`(4) as little-endian number. 
+      8. If `n=7`, `OP_INSPECTINPUTISSUANCE` Push the assetIssuance information(74-130) if the asset has issuance, otherwise push an empty vector
+      9. If `n=8`, `OP_INSPECTINPUTINDEX` Push the current input index(4) as little-endian onto the stack
+      10. If `n=9`, `OP_INSPECTINPUTANNEX` Push the annex onto the stack where the annex includes the prefix(0x50). If the annex does not exist, push an empty vector
+      11. Otherwise fail.
+   - Define `OP_SUCCESS202` as a multi-byte opcode `OP_INSPECTCURRENTINPUT` with the following semantics. Execute the push opcode following `OP_INSPECTCURRENTINPUT`, if the opcode is not a push opcode, fail. 
+      - Pop the stack pop as minimal `CScriptNum` as `n`. All the implementation is exactly the same as `OP_INSPECTINPUT`, but uses the current input instead of reading the input from stack. The names of the `OP_INSPECTINPUTx` would be replaced by `OP_INSPECTCURRENTINPUTx`
```
dgpv, line 59: Can be substituted with `OP_INSPECTINPUTINDEX OP_INSPECTINPUTx`, is it worth it to add 9 opcodes to save 2 bytes ?

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,134 @@
...
+      4. Otherwise, fail.
+
+5. **Crypto**: In order to allow more complex operations on elements, we introduce the following new crypto-operators.
+   - Define `OP_SUCCESS207` as a multi-byte opcode `OP_CRYPTO` with the following semantics. Execute the push opcode following `OP_CRYPTO`, if the opcode is not a push opcode, fail. 
+   - Pop the stack pop as minimal `CScriptNum` as `n`.
+      1. If `n=0`, `OP_ECMULSCALAREXPVERIFY`, pop the top element as `k`. Then pop next elements as points `G1`(first), `G2`(second)
+      ..`Gk`. Next pop `k` scalars `x1`, `x2`... `xk`. Finally pop result element as as point `Q`. Assert `x1G1+x2G2+x3G3.. xkGk == Q` This counts as `k*50` towards budget sigops. If any of `G_i`,`Q` is point at infinity fail.
+      2. If `n=1`, `OP_ECNEGATE` pop a point `G`(33 bytes) from stack, push `-G` onto the stack.
+      3. If `n=2`, `OP_PUSH256NUM_0`, Push 256 bit scalar value 0.
+      4. If `n=3`, `OP_PUSH256NUM_`, Push 256 bit scalar value 1.
```
dgpv, line 115: should be `OP_PUSH256NUM_1`

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,134 @@
...
+
+5. **Crypto**: In order to allow more complex operations on elements, we introduce the following new crypto-operators.
+   - Define `OP_SUCCESS207` as a multi-byte opcode `OP_CRYPTO` with the following semantics. Execute the push opcode following `OP_CRYPTO`, if the opcode is not a push opcode, fail. 
+   - Pop the stack pop as minimal `CScriptNum` as `n`.
+      1. If `n=0`, `OP_ECMULSCALAREXPVERIFY`, pop the top element as `k`. Then pop next elements as points `G1`(first), `G2`(second)
+      ..`Gk`. Next pop `k` scalars `x1`, `x2`... `xk`. Finally pop result element as as point `Q`. Assert `x1G1+x2G2+x3G3.. xkGk == Q` This counts as `k*50` towards budget sigops. If any of `G_i`,`Q` is point at infinity fail.
+      2. If `n=1`, `OP_ECNEGATE` pop a point `G`(33 bytes) from stack, push `-G` onto the stack.
+      3. If `n=2`, `OP_PUSH256NUM_0`, Push 256 bit scalar value 0.
+      4. If `n=3`, `OP_PUSH256NUM_`, Push 256 bit scalar value 1. 
+      5. If `n=4`, `OP_SCALARADD` pop the top two stack elements(32 bytes)(`a` and `b`) as scalars, return (`(a+b)%n`) where `n` is order of secp256k1. Note that we do not error is `a>=n` or `b>=n`.
```
dgpv, line 116: should be "do not error if" rather than "is" ?

doc/tapscript_opcodes.md
sanket1729: Fair point. Maybe we don't want to do this.

doc/tapscript_opcodes.md
sanket1729: The goal is to be as expressive as possible without much additional cost. Since elements already has these implemented, adding support would not be hard. I don't know where this might be needed, but here are some guesses that I can make.
- sha256d for verifying bitcoin merkle-root(I need to double-check if bitcoin actually uses sha256d for this). 
- Usecases involving verification of some p2sh script hash.
- Users may want a 20 byte digest for saving fees instead of a 32 byte one.

dgpv:
Seems like a big effort that would increase the ergonomics of using Script, but will also make it more complex due to increased number of opcodes. I thought that Simplicity was a path to more expressive contracts. This seems to be a parallel effort to have another system that is more expressive than current Script (effectively, due to added ergonomics)

At the same time, propping up script with more power means that people's focus will be diluted - they will need to choose between studying Simplicity and this extended script.

The amount of new opcodes is huge, and implementation of each opcode is a huge undertaking. Even if it may seem to be simple to implement the opcodes, the amount of what can go wrong is usually very big. And there's no way to formally verify these implementations, as is the case with Simplicity. It seems to me that such effort, if applied towards Simplicity, would made it real-world-usable much faster

Is Simplicity not expected to be usage-ready for extended amount of time, so a decision to prop up Script is made instead ?

doc/tapscript_opcodes.md
dgpv: fair enough

apoelstra:
@dgpv The timeline for Simplicity is (hopefully) not that much longer than the timeline for this. We certainly don't mean this as a replacement for Simplicity or as some sort of crutch because we foresee long delays. The goal is rather to improve the ergonomics of the existing script and give people a sandbox to play with where they can continue to work in the "I can hex-edit a script for an imperative stack machine" model which is easy to experiment with.

So there are sorta two goals here, which I think explain how large this an effort this appears to be:
1. Let us eliminate some limits and simplify the existing covenant code that you and I have been working on. For this we basically need the streaming hashes **or** a couple of the tx introspection opcodes and the `FOR` loop. The streaming hashes are pretty-much trivial to think about and would get rid of the 520-byte limits. Replacing `CHECKSIGFROMSTACK` with introspection and `FOR` is harder, but I suspect you'll have a much easier time getting other people to review your stuff if it's structured this way.
2. Add a giant ad-hoc pile of "experimental" opcodes which are designed to be easy to implement, grant a lot of power "in principle", but which we otherwise don't plan to spend a lot of time thinking about. This is just continuing the spirit of Elements. Last time we did this was back in Elements Alpha, and some opcodes (e.g. CSV) wound up in Bitcoin, others proved powerful but too scary for Bitcoin (`CHECKSIGFROMSTACK`) and others AFAIK nobody ever found a use for (`DETERMINISTICRANDOM`).

So I dispute your charactization of "huge" for both the number of new opcodes and the size of the undertaking :). I think in practice, for you and I (and everyone else who isn't at the "basic experimentation" phase of development), the bulk of the complexity will be in understanding and reasoning about `FOR`. Then most of the other opcodes have simple/redundant descriptions "init/write/finalize hash X" or "fetch transaction data Y", and the rest we're basically just throwing over the fence to see what people do with them.

Then as for the relationship with Simplicity -- our first goals with Simplicity are to implement Miniscript fragments (you can see some [early work in this repo](https://github.com/apoelstra/rust-simplicity/)), with the idea that much of the "standard script operations" you can write Simplicty for by just taking a Miniscript and translating it directly into Simplicity, and then the difficult parts of actually writing Simplicity will be limited to the cool stuff that actually makes use of Simplicity's extra power. (And the way to do this would be either fork elements-miniscript to add some new crazy script fragment that could only be compiled to Simplicity, or to use some sort of `rawsimplicity` fragment that we'd add directly to elements-miniscript.)

We'd like to extend Miniscript on Elements to include a bit more than just sig/hash/time checks, e.g. to do some basic forms of transaction introspection and oracle signature checks, to make this progarmming model even easier to get started with. But to do that, we need `FOR` and transaction introspection.

I hope this clarifies our goals and the scope of this project.

dgpv:
> So I dispute your charactization of "huge"

This is subjective, of course. It is just that implementing/reviewing such critical code parts is always a bit anxiety-inducing (at least for me), and thus the effort expended seem *extra huge* relative to the size of the code. Each implemented opcode is in consensus-critical path, and any bug might be consensus bug. It is not on Bitcoin, so maybe the potential consequences of a bug are not that bad, but at the same time - less eyes for review before merge, less people trying to break it after merge, the bug might lurk for a long time.

It seemed to me that goal of the new opcodes is to make it easier to build more expressive miniscript, and this is what most of the people would use, not raw script. And the miniscript -> simplicity link seemed more logical. But if the goal of these is first give the tools to raw script writers, that's fine by me. Given these more specialized opcodes, maybe raw scripts will be smaller an simpler, that's a plus.

apoelstra:
It is definitely anxiety-inducing! But hopefully by only working with fixed-length hashes and EC points, and fixed-width numbers, we can eliminate the majority of the consensus risk, and reduce the review burden of these opcodes.

It may be that when we actually start writing code, we find that some of them are excessively branch-heavy or do things where we fear we may touch resource limits or allow division by zero or somehow depend on endianness, and then we'll just drop them. But we did consider this when choosing which opcodes to include. Personally for me, the multi-byte opcode stuff is more worrisome than any individual proposed opcode.

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,134 @@
...
+
+# New Opcodes for additional functionality:
+
+## Multi-byte op-codes
+
+We suggest the use of multi-byte opcodes for proper classification of additional opcodes and for limiting the number of OP_SUCCESSx used. In particular, a multi-byte opcode is a variable-length opcode containing an OP_SUCCESSx followed by a CscriptNum push with minimal representation. If the next opcode after OP_SUCCESS is not a push code or if the number is not a CScriptNum with minimal form, the script fails. The number pushed by the push operation serves as a selector for the operation to perform on the OP_SUCCESSx.
+
+1. **Streaming Opcodes for streaming hashes**: There is an existing MAX_SCRIPT_ELEMENT_SIZE is 520 bytes, that is the maximum message size that can OP_SHA256 can operate on. This allows hashing on more than 520 bytes while still preserving the existing security against resource exhaustion attacks. The proposal for this is still under discussion in https://github.com/ElementsProject/elements/pull/817. We suggest the latest scheme suggested by Russel O Connor
```
kallewoof, line 26:
```suggestion
1. **Streaming Opcodes for streaming hashes**: There is an existing MAX_SCRIPT_ELEMENT_SIZE is 520 bytes, that is the maximum message size that OP_SHA256 can operate on. This allows hashing on more than 520 bytes while still preserving the existing security against resource exhaustion attacks. The proposal for this is still under discussion in https://github.com/ElementsProject/elements/pull/817. We suggest the latest scheme suggested by Russel O'Connor
```

doc/tapscript_opcodes.md
kallewoof, line 59:
```suggestion
      - Pop the stack as minimal `CScriptNum` as `n`. The implementation is exactly the same as `OP_INSPECTINPUT`, but uses the current input instead of reading the input from stack. The names of the `OP_INSPECTINPUTx` would be replaced by `OP_INSPECTCURRENTINPUTx`
```

JeremyRubin:
It might be helpful to either add CTV explicitly, or a modified version with whatever elements extensions are needed, https://github.com/bitcoin/bitcoin/pull/21702, or to ensure that the transaction data inspection opcodes are sufficient to construct the template hash on the stack. Currently I think some fields (like scriptsigs) are missing, for legacy address support.

apoelstra:
As a MVP I think we should reduce scope to:

* streaming hash opcodes
* introspection opcodes
* 64-bit arithmetic opcodes (consistent with endianness/formatting of introspection ops)
* op_taptweak

brqgoo:
Would it be possible to add OP_INSPECTTXSIZE as an extension to OP_INSPECTTX? Advanced covenants in some cases need to enforce fee policies and it’s required to know transaction size in vBytes. Also, OP_INSPECTTXFEE should be very useful since a tx can have multiple fee outputs and it’s very tricky to inspect total fees paid.

apoelstra:
For now, to keep implementation scope down, I'm gonna say no ... you can use `INSPECTX SIZE NIP` to simulate this. (Unlike pre-taproot, there is no 201-opcode limit (see BIP 342) so the cost of this is "only" the two extra bytes for opcodes.

If this turns out to be a common pattern we'd be happy to add it in a future tapscript version (alongside the other "experimental" opcodes that are listed in this doc but which we are cutting from the initial implementation).

sanket1729:
@apoelstra, there is no `OP_INSPECTTX`. So far, we only have things that inspect stuff covered in sighash. The transaction is not covered in sighash, so it is not immediately clear how to evaluate `OP_INSPECTTXSIZE` as it depends on witness size(not covered in sighash)

Kukks:
> It might be helpful to either add CTV explicitly, or a modified version with whatever elements extensions are needed, [bitcoin/bitcoin#21702](https://github.com/bitcoin/bitcoin/pull/21702), or to ensure that the transaction data inspection opcodes are sufficient to construct the template hash on the stack. Currently I think some fields (like scriptsigs) are missing, for legacy address support.

I'd like to jump in to also suggest including BIP-119 CheckTemplateVerify, perhaps even instead of the dozen introspection opcodes (I think CTV can fulfill most use-cases these target?), for the sake of providing smart contracts that can be potentially replicated both on main chain and elements in the future.

apoelstra:
Lol, we are definitley not replacing introspection opcodes with CTV. We already have CHECKSIGFROMSTACK+CAT if you want to do covenants in indirect/inefficient ways.

@sanket1729 oh! Lol, I misread @brqgoo as writing "inspect X" i.e. "inspect arbitrary parts of the transaction" rather than "inspect TX".

I definitely agree that `INSPECTTXSIZE` would be useful. And probably more useful than `INSPECTTX` since it would not limit the total transaction size to fit into stack elements.

apoelstra:
Another thing we might want to bring into scope is widening the `CSV` timeout. See https://github.com/ElementsProject/elements/pull/961. It'd be worth doing this now because it'll be hard to do later.

Kukks:
> Lol, we are definitley not replacing introspection opcodes with CTV. We already have CHECKSIGFROMSTACK+CAT if you want to do covenants in indirect/general ways. We could plausibly introduce CTV in the future as a way to onboard developers who are excited about it, but it's not the direction that we're focusing our script development on.

Fair enough, I can live with either solution for my Liquid projects but would have preferred something that's more likely to find its way to BTC.

doc/tapscript_opcodes.md
sanket1729: In favor for simplicity, I think we should go with your suggestion of only requiring sha256

apoelstra:
Certainly if it looked likely that CTV would get into BTC then we'd pull it into Elements.

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,96 @@
...
+For opcodes that inspect data that is not committed in sighash, introspection is safe because any changes to the witness data would cause wtxid to change and it would revalidate the tx again.
+   - Define `OP_SUCCESS199` as `OP_INSPECTINPUT` with the following semantics: Pop two minimal `CScriptNum`s from stack. The first number `n` denotes the type of introspection, and the second number `idx `denotes the position of input to inspect. Immediately abort if `idx` is out of bounds.
+      1. If `n=0`, `OP_INSPECTINPUTSPENDTYPE` Push 1 byte "spend_type" onto the stack. spend_type (1) is equal to `(ext_flag * 2) + annex_present` as defined in [Modified BIP-341 SigMsg for Elements](https://gist.github.com/roconnor-blockstream/9f0753711153de2e254f7e54314f7169)
```
apoelstra, line 31: I think it might be simpler to not name these since they aren't really "opcodes" now.

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,96 @@
...
+      2. If `n=1`, `OP_INSPECTINPUTOUTPOINTFLAG` Push the outpoint_flag(1) as defined in [Modified BIP-341 SigMsg for Elements](https://gist.github.com/roconnor-blockstream/9f0753711153de2e254f7e54314f7169)
+      3. if `n=2`, `OP_INSPECTINPUTOUTPOINT` Push the outpoint as a tuple. First push the `txid`(32) of the `prev_out`, followed by a 4 byte push of `vout`
+      4. If `n=3`, `OP_INSPECTINPUTASSET` Push the `nAsset` onto the stack as two elements. The first push the assetID(32), followed by the prefix(1)
+      5. If `n=4`, `OP_INSPECTINPUTVALUE` Push the `nValue` as a tuple, value(8, 32) followed by prefix,
+      6. If `n=5`, `OP_INSPECTINPUTSSCRIPTPUBKEY` Push the scriptPubkey(35) onto the stack.
+      7. If `n=6`, `OP_INSPECTINPUTSEQUENCE` Push the `nSequence`(4) as little-endian number.
+      8. If `n=7`, `OP_INSPECTINPUTISSUANCE` Push the assetIssuance information(74-130) if the asset has issuance, otherwise push an empty vector
+      9. If `n=8`, `OP_INSPECTINPUTANNEX` Push the annex onto the stack where the annex includes the prefix(0x50). If the annex does not exist, push an empty vector
+      10. Otherwise fail.
```
apoelstra, line 40: Maybe we shoud "otherwise end the script with success" so that we can softfork in new data?

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,96 @@
...
+   - Define `OP_SUCCESS204` as `OP_TAPTWEAK` with the following semantics. Pop the first element as point `P`, second element as script blob `S`. Push the Taptweak on the top of stack `Q = P + H(P||S)*G`. If `|S| > MAX_ELEMENT_SIZE`, the user should use the streaming opcodes to compute the Hash function.
+
+3. **Signed 64-bit arithmetic opcodes:** Current operations on `CScriptNum` as limited to 4 bytes and are difficult to compose because of minimality rules. having a fixed width little operations with 8 byte signed operations helps doing calculations on amounts which are encoded as 8 byte little endian.
+   - When dealing with overflows, we explicitly return the success bit as a `CScriptNum` at the top of the stack and the result being the second element from the top. If the operation overflows, first the operands are pushed onto the stack followed by success bit. \[`a_second` `a_top`\] overflows, the stack state after the operation is \[`a_second` `a_top 0`\] and if the operation does not overflow, the stack state is \[`res 1`\].
```
apoelstra, line 64: nit: missing a couple `s between `a_top` and `0`

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,96 @@
...
+   - Define `OP_SUCCESS206` as `OP_SUB64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push a - b onto the stack. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS207` as `OP_MUL64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push `a*b` onto the stack. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS208` as `OP_DIV64`: pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). First push remainder `a%b`(must be non-negative and less than |b|) onto the stack followed by quotient(`a//b`) onto the stack. Abort if `b=0`. Push 1 `CScriptNum` if there is no overflow. Overflow behavior defined above.
+   - Define `OP_SUCCESS209` as `OP_LESSTHAN64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a < b`.
+   - Define `OP_SUCCESS210` as `OP_LESSTHANOREQUAL64`(cannot fail!): pop the first number(8 byte LE) as `b` followed another pop for `a`(8 byte LE). Push ` a <= b`.
```
apoelstra, line 72: I'm torn between having the "less than or equal" opcodes and requiring the user do `GREATERTHAN` `NOT`. What do others think?

doc/tapscript_opcodes.md
```diff
@@ -0,0 +1,96 @@
...
+   - Define `OP_SUCCESS219` as `OP_RSHIFT`: pop the first number as `CScriptNum` `r`(abort if r < 0 or r > 63) followed another pop for `a` (8 byte LE). Push `a >> r`.(Sign bit is preserved).
+
+4. **Conversion opcodes:** Methods for conversion from `CScriptNum` to `8-byte LE`, `4-byte LE`.
+   - Define `OP_SUCCESS220` as `OP_SCIPTNUMTOLE64`: pop the stack as minimal `CSciptNum`, push 8 byte signed LE corresponding to that number.
+   - Define `OP_SUCCESS221` as `OP_LE64TOSCIPTNUM`: pop the stack as a 8 byte signed LE. Convert to `CScriptNum` and push it, abort on fail.
+   - Define `OP_SUCCESS222` as `OP_LE32TOLE64`: pop the stack as a 4 byte signed LE. Push the corresponding 8 byte LE number. Cannot fail, useful for conversion of version/sequence.
+
+6. **Changes to existing Opcodes**:
+   - Add `OP_CHECKSIGFROMSTACK` and `OP_CHECKSIGFROMSTACKVERIFY` to follow the semantics from bip340-342 when witness program is v1.
```
apoelstra, line 89: How about we remove `CHECKSIGFROMSTACK` and leave only the VERIFY variant?

apoelstra:
FYI to readers, Sanket's recent force-push actually rewrites the document to reduce its scope. The new document lists everything we actually plan to deploy on Liquid alongside Taproot. The other opcodes will have to wait for another softfork.

doc/tapscript_opcodes.md
sanket1729: It is somewhat risky. If the user does some manipulation on the stack to figure out which value `n` to use and some makes an error in it, it could result in loss of funds. 

I guess the same could be said for `k` in `k on n` multisig. So, I feel safe in changing it to success.

doc/tapscript_opcodes.md
sanket1729: I don't feel strongly about it. It's not additional code complexity so might as well have it?

sanket1729 (commented):
Pushed another commit addressing @apoelstra's suggestions

JeremyRubin:
somewhat of a concept NACK towards inspect annex, given that there are no defined uses for annex yet -- when one is soft forked in, it should be enabled then?

i think there are some circumstances where being able to put the annex on the stack might be counter to the purpose of the annex itself.

JeremyRubin:
> Certainly if/when it looked likely that CTV would get into BTC then we'd pull it into Elements.

Is that normative for Elements? I.e., do the other upgrades planned in this release pass such a test?

doc/tapscript_opcodes.md
JeremyRubin: I agree that both seems reasonable -- although I thought CHECKSIGFROMSTACK was already part of elements?

apoelstra:
Yes, if it looked likely that anything in this planned release were to get into Bitcoin, we would still include it in the planned release.

apoelstra:
> I don't feel strongly about it. It's not additional code complexity so might as well have it?

@sanket1729 the non-`VERIFY` version has no benefit over the `VERIFY` version but cannot be batch-verified.

A similar argument applies to `TAPTWEAK`, although in this case (theoretically) the non-`VERIFY` version is more efficient because it does not require the user to include the result on the witness stack. But in practice I think the same argument applies, and we should have only `TAPTWEAKVERIFY`.

doc/tapscript_opcodes.md
JeremyRubin: I think there's a decent amount of value in just having GT in terms of minimalism, but in practice having both GT/LT should be able to yield more efficient scripts...

JeremyRubin:
> Yes, if it looked likely that anything in this planned release were to get into Bitcoin, we would still include it in the planned release.

I mean that is the inclusion of features in this upgrade to Elements an indication that the house view is these are going to be included in Bitcoin?

Or would it being clear a feature would not go into Bitcoin be a reason to disinclude from elements?

sanket1729:
Hi @JeremyRubin, I spent some time trying to read BIP119 and figuring out how it might adapt to elements. Some comments in no particular order of importance:

- Directly migrating `OP_CTV` is tricky because fees in elements are output UTXO. The value cannot be known statically while receiving money in `CTV` controlled output. We would have to modify the template w.r.t to not include fee outputs.
- It might require more thinking about how the templating would interact with confidential assets. Just directly using the same template from bitcoin BIP 119 might not be the best way.
- All the functionality which `OP_CTV` allows is permitted by the current design. You can directly fix the transaction structure by looking at the outputs themselves. The solution is more generic and flexible than a fixed template. You can emulate any specific template that you want here. 
- The best selling point about BIP119 is IMO the tooling around it(Sapio). But I don't think we have the bandwidth to extend/maintain Sapio for elements.

These things might require more discussion, so it's to move this to a new issue?

sanket1729:
Opened an issue https://github.com/ElementsProject/elements/issues/1015 for more discussion.

doc/tapscript_opcodes.md
sanket1729: They are part of elements, but we need to modify them to use BIP340-342 type schnorr signatures. Currently, this uses ECDSA sigs

doc/tapscript_opcodes.md
sanket1729: The concern here is that we are consuming opcodes(already up to 222/254) and we don't want to run out of them in future upgrades. 

The previous version of this doc had multi-byte opcodes(opcodes for ec operations, scalar exps etc) which were removed in favor of implementation simplicity.

sanket1729:
> somewhat of a concept NACK towards inspect annex, given that there are no defined uses for annex yet -- when one is soft forked in, it should be enabled then?

It is already possible to inspect `annex` using the CSFS trick. So, having an additional opcode to inspect it does not any new functionality that elements codebase currently does not have.

JeremyRubin:
> > somewhat of a concept NACK towards inspect annex, given that there are no defined uses for annex yet -- when one is soft forked in, it should be enabled then?
> 
> It is already possible to inspect `annex` using the CSFS trick. So, having an additional opcode to inspect it does not any new functionality that elements codebase currently does not have.

right... that's an interesting observation!

apoelstra:
> @sanket1729 the non-VERIFY version of CSFS has no benefit over the VERIFY version but cannot be batch-verified.

Sanket pointed out to me that we already have a consensus rule for CHECKSIG that any invalid signature needs to be the null signature. If we extend this rule to CSFS then there is no issue wit h batch verification.

apoelstra:
Hmm, crap, I meant to merge the other doc PR #933.

Since this is documentation only I think it's OK to just leave it be, and to make any additional changes in a future PR.

Issue #875, opened 2020-05-22
createpsbt output_assets parameter not working


asoltys:
According to the docs for createpsbt, I should be able to provide a JSON object, "output_assets", as the fifth argument to the command to map outputs to assets as with createrawtransaction. But when I provide it, I get error code: -1 and the full createpsbt help text back as an error message.

This works with 4 arguments:

```
elements-cli createpsbt [] "[{\"AzpkXpbFsWC9aWJvah3JKg99zenqsGyYVhUuYWePmEK3LVozbf6by5wKRMXVKeZP34WzSJ5XwuKSukzM\": 0.05}]" 0 false
```

But as soon as I add the fifth argument, it doesn't:

```
elements-cli createpsbt [] "[{\"AzpkXpbFsWC9aWJvah3JKg99zenqsGyYVhUuYWePmEK3LVozbf6by5wKRMXVKeZP34WzSJ5XwuKSukzM\": 0.05}]" 0 false "{\"AzpkXpbFsWC9aWJvah3JKg99zenqsGyYVhUuYWePmEK3LVozbf6by5wKRMXVKeZP34WzSJ5XwuKSukzM\":\"b2e15d0d7a0c94e4e2ce0fe6e8691b9e451377f6e46e8045a86f7c4b5d4f0f23\"}"
```

I can try to investigate more, just raising the issue in the meantime.

tiero:
@asoltys A possible workaround would be to call the `createrawatransaction` first and then pass to the the `converttopsbt` command the resulting hex string.

This is the only way AFAIK to create a PSET that spend an assets beside bitcoin.

asoltys:
I think that's what I was attempting to do when I ran into #877

apoelstra:
In #900 we overhauled the PSBT code and removed this parameter from `createpsbt`.

PR #950, opened 2020-12-15
Add output_assets arg to createpsbt rpc

Sosthene00:
- createpsbt used to fail when called with an output_assets.
- update the rpc code to make it work with other assets than bitcoin

Fix issue #875 

I noticed that the cause for `createpsbt` not working with `output_assets` argument was that the code had obviously not been updated to account for the new `output_assets` argument compared to the Bitcoin Core version of the rpc, so I made a couple of changes and it works now.


apoelstra:
Thanks! These changes will be clobbered by #935 but I'd like to keep this open to see if the same issue persists once that PR gets in.

PR #951, opened 2020-12-16
Spec for PSET

achow101:
Adds the document specifying the new PSET format that #900 implements.

Pulled out of #900

### DIFF
--- a/doc/pset.mediawiki
+++ b/doc/pset.mediawiki
@@ -1,0 +1,443 @@
+Note that this format is not yet implemented by Elements.
+
+<pre>
+  BIP: PSET
+  Layer: Applications
+  Title: Partially Signed Elements Transaction Format
+  Author: Andrew Chow <achow101@gmail.com>
+  Comments-Summary: No comments yet.
+  Status: Proposed
+  Type: Standards Track
+  Created: 2020-07-09
+  License: BSD-2-Clause
+</pre>
+
+==Introduction==
+
+===Abstract===
+
+This document describes proprietary extensions to the BIP 174 Partially Signed Bitcoin Transaction format.
+These extensions are for use in Elements Sidechains to hold the information necessary for
+Confidential Transactions, Confidential Assets, and Peg-in transactions.
+
+===Copyright===
+
+This BIP is licensed under the 2-clause BSD license.
+
+==Specification==
+
+The Partially Signed ELements Transaction (PSET) format is identical to the BIP 370 PSBT format.
+The changes are new proprietary type fields, a new magic sequence, and new roles.
+The fields added for PSET are only allowed when the PSBT version is 2.
+
+For elements, the identifier prefix string is <tt>pset</tt>.
+Note that previous implementations of PSET used a different prefix string and had different fields.
+The use of a new identifier prefix is to avoid conflicts with the previous implementations.
+
+The currently defined elements global proprietary types are as follows:
+
+{|
+! Name
+! <tt><keytype></tt>
+! <tt><keydata></tt>
+! <tt><keydata></tt> Description
+! <tt><valuedata></tt>
+! <tt><valuedata></tt> Description
+! Versions Requiring Inclusion
+! Versions Requiring Exclusion
+! Versions Allowing Inclusion
+|-
+| Scalar Offset
+| <tt>PSBT_ELEMENTS_GLOBAL_SCALAR = 0x00</tt>
+| <tt><32 byte scalar></tt>
+| A 32 byte Scalar Offset computed by a blinder.
+| None
+| No value. The value must have a length of zero.
+|
+| 0
+| 2
+|-
+| Elements Transaction Modifiable Flag
+| <tt>PSBT_ELEMENTS_GLOBAL_TX_MODIFIABLE = 0x01</tt>
+| None
+| No key data
+| <tt><8-bit uint></tt>
+| An 8 bit little endian unsigned integer as a bitfield for various elements specific transaction modification flags. Bit 0 is the PSBT Blinded flag and it is set to 1 to indicate that the PSET has not been blinded yet. Once all confidential values, rangeproofs, and asset surjection proofs have been attached to the PSET, it must be set to 0.
+|
+| 0
+| 2
+|}
+
+The currently defined elements per-input proprietary types are as folows:
+
+{|
+! Name
+! <tt><keytype></tt>
+! <tt><keydata></tt>
+! <tt><keydata></tt> Description
+! <tt><valuedata></tt>
+! <tt><valuedata></tt> Description
+! Versions Requiring Inclusion
+! Versions Requiring Exclusion
+! Versions Allowing Inclusion
+|-
+| Issuance Value
+| <tt>PSBT_ELEMENTS_IN_ISSUANCE_VALUE = 0x00</tt>
+| None
+| No key data
+| <tt><64-bit int></tt>
+| The explicit little endian 64-bit integer for the value of this issuance. This is mutually exclusive with <tt>PSBT_ELEMENTS_IN_ISSUANCE_VALUE_COMMITMENT</tt>
+|
+| 0
+| 2
+|-
+| Issuance Value Commitment
+| <tt>PSBT_ELEMENTS_IN_ISSUANCE_VALUE_COMMITMENT = 0x01</tt>
+| None
+| No key data
+| <tt><33 byte commitment></tt>
+| The 33 byte Value Commitment. This is mutually exclusive with <tt>PSBT_IN_ISSUANCE_VALUE</tt>.
+|
+| 0
+| 2
+|-
+| Issuance Value Rangeproof
+| <tt>PSBT_ELEMENTS_IN_ISSUANCE_VALUE_RANGEPROOF = 0x02</tt>
+| None
+| No key data
+| <tt><rangeproof></tt>
+| The rangeproof
+|
+| 0
+| 2
+|-
+| Issuance Inflation Keys Rangeproof
+| <tt>PSBT_ELEMENTS_IN_ISSUANCE_KEYS_RANGEPROOF = 0x03</tt>
+| None
+| No key data
+| <tt><rangeproof></tt>
+| The rangeproof
+|
+| 0
+| 2
+|-
+| Peg-in Transaction
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_TX = 0x04</tt>
+| None
+| No key data
+| <tt><tx></tt>
+| The Peg-in Transaction serialized without witnesses.
+|
+| 0
+| 2
+|-
+| Peg-in Transaction Output Proof
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_TXOUT_PROOF = 0x05</tt>
+| None
+| No key data
+| <tt><txoutproof></tt>
+| The transaction output proof for the Peg-in Transaction.
+|
+| 0
+| 2
+|-
+| Peg-in Genesis Hash
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_GENESIS = 0x06</tt>
+| None
+| No key data
+| <tt><hash></tt>
+| The 32 byte genesis hash for the Peg-in Transaction.
+|
+| 0
+| 2
+|-
+| Peg-in Claim Script
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_CLAIM_SCRIPT = 0x07</tt>
+| None
+| No key data
+| <tt><script></tt>
+| The claim script for the Peg-in Transaction.
+|
+| 0
+| 2
+|-
+| Peg-in Value
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_VALUE = 0x08</tt>
+| None
+| No key data
+| <tt><64-bit int></tt>
+| The little endian 64-bit value of the peg-in for the Peg-in Transaction.
+|
+| 0
+| 2
+|-
+| Peg-in Witness
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_WITNESS = 0x09</tt>
+| None
+| No key data
+| <tt><script></tt>
+| The Peg-in witness for the Peg-in Transaction.
+|
+| 0
+| 2
+|-
+| Issuance Inflation Keys Amount
+| <tt>PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS = 0x0a</tt>
+| None
+| No key data
+| <tt><64-bit int></tt>
+| The value for the inflation keys output to set in this issuance. This is mutually exclusive with <tt>PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS</tt>.
+|
+| 0
+| 2
+|-
+| Issuance Inflation Keys Amount Commitment
+| <tt>PSBT_ELEMENTS_IN_ISSUANCE_INFLATION_KEYS_COMMITMENT = 0x0b</tt>
+| None
+| No key data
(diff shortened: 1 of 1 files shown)


apoelstra:
Thanks!

I think we are going to change this to make it based off of PSBT2, which should simplify a lot of the transaction-hacking (and perhaps PSBT2 will have a "psbt ID" that we can make stay constant even after reblinding...maybe).

Useful to have this PR so that we can iterate on that independently of the code review.

doc/pset.mediawiki
```diff
@@ -0,0 +1,290 @@
...
+The currently defined elements global proprietary types are as follows:
+
+* Type: Scalar Offset <tt>PSET_ELEMENTS_GLOBAL_SCALAR = 0x00</tt>
+** Key: A 32 byte Scalar Offset computed by a blinder.
+*** <tt>{0x00}|{offset}</tt>
+** Value: None. The value must have a length of zero.
+
+The currently defined elements per-input proprietary types are as folows:
+
+* Type: Issuance Value <tt>PSBT_ELEMENTS_IN_ISSUANCE_VALUE = 0x00</tt>
```
tiero, line 53: Am I wrong, but this maybe should be called `PSET_` and not `PSBT_` ?

dgpv: I wrote in a comment to the previous PR (https://github.com/ElementsProject/elements/pull/900#issuecomment-693228600) why I think that `PSBT_ELEMENTS_...` is better than `PSET_...` and it seems that my arguments were accepted

doc/pset.mediawiki
```diff
@@ -0,0 +1,290 @@
...
+
+The blinder will then compute a scalar offset that will be added as a <tt>PSBT_ELEMENTS_GLOBAL_SCALAR</tt>.
+For each input and output owned/blinded by this blinder, the following formula is computed:
+<tt> asset_blinding_factor * amount + amount_blinding_factor (mod n)<tt>.
+The scalars for the inputs are summed, and then that sum is subtracted from the sum of the scalars for the outputs.
+The result is the scalar offset added as a <tt>PSBT_ELEMENTS_GLOBAL_SCALAR</tt>.
+
+The Blinder can identify which outputs it should blind by checking the <tt>PSBT_ELEMENTS_OUT_BLINDER_INDEX</tt>.
+Using that index, the Blinder should inspect the input at that same index.
+If that input belongs to the Blinder, then the output is to be blinded by it.
```
tiero, line 248: The language here is not entirely clear here IMHO. Unless I don't understand what it mean: `the output is to be blinded by it` 

Also opens up some question: what about swap-like or "complex" coin-joining scheme? How we can link input to output by same index?

tiero:
Let's say we want to support use-cases where multiple parties adds inputs and interactively sign (coinjoin, swaps) and is acceptable that the last party adding inputs/outputs (ie. SIGHASH_ALL) will cover the *Blinder role* knowing therefore the blinding keys of others (or asset/amount blinders):

We want to have a dedicated field for sharing these secrets in the PSET or this is out of scope and we will leave to be a construction swap/coinjoin protocols will build on top?

apoelstra:
The recipients can learn the blinding factors of their outputs in the usual way (using the rangeproof sidechannel), so PSET doesn't need to worry about this.

tiero:
> The recipients can learn the blinding factors of their outputs in the usual way (using the rangeproof sidechannel), so PSET doesn't need to worry about this.

Pardon, what's a rangeproof "sidechannel" ?

Also here it's the blinder (which may or not may be one of the recipients) that needs to get blinders somehow from other parties. 
I agree that this should be out of scope of PSET and will be done eventually by specific procotols (eg. [for atomic swaps in tdex each party incudes his own blinding keys,](https://github.com/TDex-network/tdex-specs/blob/master/03-swap-protocol.md#data-structures) so the the `Responder` could blind all the outputs one-shot)

doc/pset.mediawiki
achow101: An output `O` has an int `I_i` which is the index of an input. If a blinder owns the input at `I_i`, it must also blind the output `O`.

louisinger: I understand that this makes it possible to define a specific blinder for each output. But why is this necessary? Is it a problem if the blinder has no input in the transaction?

tiero: Sorry, reading it I thought "it" was about the input, but `it` means the Blinder. So all good

doc/pset.mediawiki
```diff
@@ -0,0 +1,290 @@
...
+For convenience, this BIP will specify types using their full serialization, so a multi-byte type will have it's full prefix and zero padding as necessary.
+There are global types, per-input types, and per-output types.
+
+For elements, the identifier prefix string is <tt>pset</tt>.
+Note that previous implementations of PSET used a different prefix string and had different fields.
+The use of a new identifier prefix is to avoid conflicts with the previous implementations.
+
+The currently defined elements global proprietary types are as follows:
+
+* Type: Scalar Offset <tt>PSET_ELEMENTS_GLOBAL_SCALAR = 0x00</tt>
```
louisinger, line 46: should be `PSBT_ELEMENTS...` instead of `PSET_ELEMENTS...`?

apoelstra:
@tiero the "rangeproof sidechannel" is a mechanism in Elements that lets you encrypt data into a rangeproof with no additional space, such that the recepient can decrypt. This is currently used in Liquid so that senders can create blinded outputs and then communicate the blinding factors to the recipients.

apoelstra:
Doc needs update to be based on PSBT2 BIP-0370 https://github.com/achow101/bips/blob/psbt2/bip-0370.mediawiki ... should be fairly minor stuff but will let us clean up the "replace explicit outputs with blinded ones" workflow.g

apoelstra:
Opened a PR to this branch over here https://github.com/achow101/elements/pull/2

Will start working on implementing this on rust-elements and elements-miniscript, and in parallel @achow101 will work on building something on top of #935, with the goal of getting some test vectors so that everyone else can start moving forward again (and we can do a libwally implementation). Sorry for the delays on this.

doc/pset.mediawiki
achow101: Fixed

achow101:
I have updated this to use the formatting used in BIPs 174 and 370. Additionally this has been updated to be dependent on BIP 370 and PSET will only work with PSBTv2.

doc/pset.mediawiki
```diff
@@ -0,0 +1,429 @@
...
+| <tt><32 byte scalar></tt>
+| A 32 byte Scalar Offset computed by a blinder.
+| None
+| No value. The value must have a length of zero.
+|
+| 0
+| 2
+|-
+| Elements Transaction Modifiable Flag
+| <tt>PSBT_ELEMENTS_GLOBAL_TX_MODIFIABLE = 0x01</tt>
```
sanket1729, line 61: Is there any reason why we don't support the `GLOBAL_XPUB` in pset?

sanket1729 (commented):
In the elements Taproot sighash, we also a global `blockchain_genesis_hash` in order to avoid reuse of signatures across blockchains. I think we should add a pset global proprietary field for it as it would needed to construct the Sighash Message

doc/pset.mediawiki
```diff
@@ -0,0 +1,429 @@
...
+| None
+| No key data
+| <tt><64-bit int></tt>
+| The little endian 64-bit value of the peg-in for the Peg-in Transaction.
+|
+| 0
+| 2
+|-
+| Peg-in Witness
+| <tt>PSBT_ELEMENTS_IN_PEG_IN_WITNESS = 0x09</tt>
```
sanket1729, line 175: The name makes it seem that the type of the value should Witness, but it is shown as a script. How does this work for segwit transactions?

sanket1729: Maybe this question does not make sense, and pegin transactions input only have witness in ScriptSig.

achow101: It's a `CScriptWitness`: https://github.com/ElementsProject/elements/blob/master/src/pegins.cpp#L521

doc/pset.mediawiki
```diff
@@ -0,0 +1,429 @@
...
+! Versions Requiring Inclusion
+! Versions Requiring Exclusion
+! Versions Allowing Inclusion
+|-
+| Value Commitment
+| <tt>PSBT_ELEMENTS_OUT_VALUE_COMMITMENT = 0x01</tt>
+| None
+| No key data
+| <tt><33 byte commitment></tt>
+| The 33 byte Value Commitment for this output. This is mutually exclusive with <tt>PSBT_OUT_VALUE</tt>.
```
sanket1729, line 243: We should also add a note mentioning that `PSBT_OUT_VALUE` is not mandatory in pset whereas it is in psbt2.

doc/pset.mediawiki
```diff
@@ -0,0 +1,429 @@
...
+! <tt><keydata></tt>
+! <tt><keydata></tt> Description
+! <tt><valuedata></tt>
+! <tt><valuedata></tt> Description
+! Versions Requiring Inclusion
+! Versions Requiring Exclusion
+! Versions Allowing Inclusion
+|-
+| Value Commitment
+| <tt>PSBT_ELEMENTS_OUT_VALUE_COMMITMENT = 0x01</tt>
```
sanket1729, line 239: Proprietary fields for Input, Global start from `0x00`. Is there any special reason why this starts from `0x01`?

doc/pset.mediawiki
achow101: Done

doc/pset.mediawiki
achow101: Originally, the output's value was in field `0x00`. However with the move to PSBTv2, this was no longer necessary and I removed that field. I just didn't change the rest of the field numbers. I don't feel strongly about keeping these, but it does make implementation easier to not change them.

sanket1729:
How is the serialization for range proof and surjection proofs `witness_utxo` supposed to be done? In the current elements, there are not serialized as a part of TxOut, but as additional witness data at the end of the transaction. 

Should we add separate fields for those? I only see proprietary fields for issuances/pegins in inputs. In a common workflow, the user would need to rewind the rangeproof in order to get the value/asset.

sanket1729:
Or is the type of `witness_utxo` a Psbt/Pset output and not an elements TxOut?

achow101:
> @achow101 How is the serialization for range proof and surjection proofs `witness_utxo` supposed to be done? In the current elements, there are not serialized as a part of TxOut, but as additional witness data at the end of the transaction.
> 
> Should we add separate fields for those? I only see proprietary fields for issuances/pegins in inputs. In a common workflow, the user would need to rewind the rangeproof in order to get the value/asset.

Currently neither the range proof nor the surjection proof is provided in the input UTXOs. I don't see why the surjection proof would be required, but for the rangeproof, the only reason to have it is to be able to get the encrypted blinding factor from it. However since the range proof is really large, I am hesitant to put it there in its entirety. Instead, we could have a field for the encrypted blinding factor. The ecdh pubkey is already stored in `CTxOut.nNonce`.

tiero:
> However since the range proof is really large, I am hesitant to put it there in its entirety. 

Owning the blinding private key and a a call to any explorer/node, anyone can get the proofs and unblind it. Especially if wallets implements slip77, just the script is enough to derive the blinding from the seed/master blind key

tiero:
> > a call to any explorer/node
> 
> I think the scenario is for offline blinders which don't have access to any nodes. After all, the point of a PSET is to have everything for signing and blinding, no need to call a node.

That's true, I am more in favor to put asset/value blinders then. This could unlock indeed more advanced cooperative multi-party blinding/unblinding

achow101:
> That's true, I am more in favor to put asset/value blinders then. This could unlock indeed more advanced cooperative multi-party blinding/unblinding

The purpose of this rework of PSET is to avoid having to put asset and value blinders into the PSET. The blinding factors are supposed to be secret, while PSETs can be passed around to untrusted entities. Having the blinding factors in the PSET destroys the privacy preserving aspects of elements, so I don't think it makes a lot of sense to have them in. What workflows does having the blinding factors in the PSET enable?

tiero:
@achow101  Sorry, I meant encrypted of course, ref:  your comment before

> Instead, we could have a field for the encrypted blinding factor


The main advantage would be cooperative blinding for swaps/coin join, but yeah there could be [other means to do it, on top of PSET](https://github.com/ElementsProject/elements/pull/951#issuecomment-771681268) without including in it

sanket1729:
@achow101, This would be an optional field, and only the users who really care about applications that require rewinding the proofs would use it. 

Here is one simple scenario where this might be useful. The user has an SW watch-only software that wants to confirm a receiving confidential transaction. The SW provides the 1) Rangeproof, 2) the `TxOut.nNonce` from the receiving transaction and 3) the receiver address(or the ecdh_pubkey) to the HW wallet that does the unblinding tells the SW about the asset and value it is receiving. 
In my opinion, we should add two per input optional fields for 1) rangeproof and 2) the receiver pubkey in the CT address. Since `txout.nNonce` already contains the `ephermal_pk`, the unblinder can use these three together to unblind the txout. 

1) Since the field is completely optional, applications that don't care about this don't have to use it. It is already possible to inflate psbts/psets by filling in unnecessary optional things in global maps/key maps etc. So, adding this is not creating any new DoS that does not already exist
2) We already have range proofs for outputs, issuance/re-issuance inputs. 

w.r.t to surjection proofs, you are right that these are not needed.  

>  So updaters can extract the encrypted blinding factor from the rangeproof and put that in the PSET.

The current `libsecp_zkp` API does not provide any such support for extracting encrypted blinding factors from range proofs. It only offers a `rewind` method that needs the secret key after the ecdh computation.

apoelstra:
I think we should add the extra field. As Sanket observes,

1. Encoding the value/blinder directly would be a serious privacy leak, but we don't have any standard way to encrypt this data to the Blinder, other than the entire rangeproof.
2. We already encode entire rangeproofs for the outputs of the transaction; so encoding them for the inputs as well isn't that big a loss.

Both of these problems are "temporary" in that they'll be eliminated by Bulletproofs.

achow101:
I've added an additional field where the UTXO's rangeproof can be specified.

sanket1729:
Thanks, @achow101. For full utility of the use-case that I described above, we need three things.

1) Rangeproof: available by `PSBT_ELEMENTS_IN_UTXO_RANGEPROOF`. 
2) Sender Ecdh pubkey: available inside `PSBT_ELEMENTS_IN_UTXO` in txout.nNonce
3) (un)Blinding pk: Typically stored in the CT address that the receiver uses.
 
we also need the public key from the receiver address. Is it also possible to add optional pubkey `PSBT_ELEMENTS_IN_UTXO_BLINDING_PUBKEY` ?

achow101:
> 3. (un)Blinding pk: Typically stored in the CT address that the receiver uses.
>
> we also need the public key from the receiver address. Is it also possible to add optional pubkey `PSBT_ELEMENTS_IN_UTXO_BLINDING_PUBKEY` ?

I'm not sure that updaters will necessarily know this. The blinding pubkey is not stored publicly anywhere and I'm not sure that updaters would necessarily be able to derive the blinding pubkey for an output. AFAIU, blinders should know how their blinding pubkeys are derived so a blinder can try to unwind every rangeproof. The ones where the unwinding is successful are the inputs that the blinder owns.

apoelstra:
@sanket1729 @achow101 think we should merge this since both PSET implementations are nearing completion?

achow101:
Fine by me

sanket1729 (approved):
ACk 8e19dfa35ba6e1f198a3218892aad8f420cadba8

PR #952, opened 2020-12-22
[experimental] [idea] OP_CHECKEXCHANGE

kallewoof:
Feasibility, brokenness, conceptual soundness etc aside...

... this is a new operator called OP_CHECKEXCHANGE, which facilitates automated swaps and/or give-outs of specified assets.

The operator takes two sets of inputs -- withdraw rules  and payment rules.

Withdraw rules are composed of an asset ID and a maximum amount. To satisfy the rules, the spending transaction must take no more than the maximum of the given asset and return the remaining balance to the same script.

There may be (currently) up to 10 withdraw rules in one OP_CHECKEXCHANGE expression.

Payment rules are composed of an asset ID, a minimum, and a destination address. To satisfy the rules, the spending transaction must send no less than the minimum of the given asset to the specified destination.

There may be up to 10 of these, as well, i.e. a maximum of 20 rules.

Things you can do with this operator:
* Automated faucets, handing out some form of token.
* Automated exchange systems, where users can exchange or buy a specified asset using some other asset (or bitcoin)
* ???

Either way, it was a fun experiment. I don't know if it's feasible for anything, but the code turned out to not be super complex (although the requirement for the input amount map is a bit iffy) so I'm presenting it here.

Note: this does not work with CTCA, atm. I don't think it's even possible, but please prove me wrong. :)


dgpv:
> Note: this does not work with CTCA, atm. I don't think it's even possible, but please prove me wrong. :)

The proposed opcode can be implemented with covenant, but cannot work with CTCA.

Basically this is a restriction on the outputs, the only problem would be allowing different destination addresses. This can be achieved either by constricting the sighash in the covenant itself, and checking the constucted sighash with CHECKSIGFROMSTACK, or using new opcodes proposed in https://github.com/ElementsProject/elements/pull/949 (`INSPECTOUTPUT*`)

Covenant can work with CTCA though, by pre-computing all the blindings and using deterministic derivation for blinding keys and deterministic random. The number of transactions in the transaction chain this covenant will enable will be limited, but it can be quite large. Also the address of the covenant will be different each time, but this likely doesn't matter because the script will be revealed in the spend anyway.

kallewoof:
@dgpv Thanks for the feedback! Yeah, I figured covenants could do most of what this does. I'll look at that and see if I can figure it out.

dgpv:
You can find an example of the covenant for partial matching of outputs in https://ruggedbytes.com/articles/ll/#covenant-scripts

It also demonstrates the multiple-choice construction to allow different variants of outputs in one covenant (`OUTPUTS_HASH_LOOKUP_OPS`), which can be used to create the "different withdraw rules" behavior

kallewoof:
Thanks a lot, will look at those :)

dgpv:
> but cannot work with CTCA.

I think I was mistaken. If the opcode is essentially a hard-coded covenant construction, nothing says that it cannot also check the blinded values. These blinded values can be pre-calculated the same way as for the covenants, with deterministic derivation. The code can discriminate blinded/unblined by the size of the data supplied.

dgpv:
Although "return the remaining balance to the same script." won't work with blinding. You would need to provide the next scriptPubKey for the contract explicitly, because blinding will be different each time

dgpv:
Also to work with blinding, the opcode can't rely on input values, it just needs to ensure that the change went to the "contract" address. This way the balance will be checked by the generic rules for the balance of the whole transaction.

dgpv:
Thinking about this more, trying to turn `OP_CHECKEXCHANGE` construction to work with blinding just turns it into a simple OP_CTV covenant, because all the values need to be precomputed. 

> Withdraw rules are composed of an asset ID and a maximum amount. To satisfy the rules, the spending transaction must take no more than the maximum of the given asset and return the remaining balance to the same script.

> Payment rules are composed of an asset ID, a minimum, and a destination address. To satisfy the rules, the spending transaction must send no less than the minimum of the given asset to the specified destination.

There can't be "minimum" or "maximum". Only fixed values, or gradation of values, which need to be present in the script.

For each withdrawal, there should be exactly one change output to the same covenant, because input values cannot be checked

For each payment, there should be exactly one output

The number of outputs must be equal to <number_of_withdrawals>*2+<number_of_payments> (can't be less, because we don't know the sum of inputs)

So all the work will be in pre-computing the output values for the chain of transactions (or a tree of transactions if there's gradation of values)

dgpv:
> but cannot work with CTCA.
> I think I was mistaken. If the opcode is essentially a hard-coded covenant construction

To conclude, I was not mistaken, and @kallewoof's first impression that this won't work with CTCA was correct.

It might be possible to emulate this opcode with covenant on the unblinded transaction. It will be much more complex with sighash-building approach. With new introspection opcodes it should be easier to emulate this, using `OP_INSPECTINPUTVALUE`, `OP_INSPECTOUTPUTVALUE`, and `OP_FOR` loop.

kallewoof:
Very interesting! Thanks for taking the time to put thought into this.

kallewoof:
Actually by knowing the initial balance and accounting for that each time, the entire `input_map` thing can be pulled out and replaced by a simple required value (which drops each time). Perhaps the opcode would have that as an incoming argument and that argument would be lowered every time the contract executed.

dgpv:
> It does remove the input map requirement though.

It also adds requirements that the withdraw and payment values must be discrete, otherwise the number of possible "next state" scriptPubKeys will be very big.

You will also need to ensure that the change goes to the "next state" scriptPubKey and not anywhere else. That requires restricting the outputs.

These requirements make the mechanism very much resemble how it would work with OP_CTV-like covenant.

The `input_map` allows to store the state of the contract in the blockchain itself, in the values of UTXO. Without it, the state of the contract needs to be saved in the utxo itself - it means you only have the "state identifier" (the txid:vout of UTXO), and what this state identifier means needs to be precomputed. With `input_map` your state can be much bigger because you have all the `nValue` bits to store, instead of just one bit of "has state moved to the next state id" ?

kallewoof:
Closing this as I think all the feedback has been given.

PR #959, opened 2021-02-18
Make it clear that the CRangeCheck is not called on explicit values

stevenroose:
That is enforced by the `VerifyAmounts` method for regular transactions
and by the `VerifyIssuanceAmounts` method for issuance values.

### DIFF
--- a/src/confidential_validation.cpp
+++ b/src/confidential_validation.cpp
@@ -52,9 +52,7 @@ CAmountMap GetFeeMap(const CTransaction& tx) {
 }
 
 bool CRangeCheck::operator()() {
-    if (val->IsExplicit()) {
-        return true;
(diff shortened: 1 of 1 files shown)


stevenroose:
Let's await CI to see if I might have missed something, but I'm pretty confident that that assertion always holds.

sanket1729 (approved):
tACK a8a1a933dc211c2efb3c254ccf7c4d4115287b50. 

Verified that all tests pass locally and checked that the only two calls for `CRangeCheck` are never called on explicit amounts. 
Asserting this is safe and helpful.

gwillen:
utACK, although I'm not sure how much good review will do here -- if it tests ok, that seems like the real check.

PR #960, opened 2021-02-18
Add SIGHASH_RANGEPROOF support

stevenroose:
Add a new sighash flag that includes the rangeproof information in the signature. This avoids certain kinds of malleability in PSBT scenarios.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -186,11 +186,17 @@ bool static IsLowDERSignature(const valtype &vchSig, ScriptError* serror) {
     return true;
 }
 
-bool static IsDefinedHashtypeSignature(const valtype &vchSig) {
+bool static IsDefinedHashtypeSignature(const valtype &vchSig, unsigned int flags) {
     if (vchSig.size() == 0) {
         return false;
     }
     unsigned char nHashType = vchSig[vchSig.size() - 1] & (~(SIGHASH_ANYONECANPAY));
+
+    // ELEMENTS: Only allow SIGHASH_RANGEPROOF if the flag is set (after dynafed activation).
+    if ((flags & SCRIPT_SIGHASH_RANGEPROOF) == SCRIPT_SIGHASH_RANGEPROOF) {
+        nHashType = nHashType & (~(SIGHASH_RANGEPROOF));
+    }
+
     if (nHashType < SIGHASH_ALL || nHashType > SIGHASH_SINGLE)
         return false;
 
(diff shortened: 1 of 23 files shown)


gwillen:
concept ACK; I have looked over the code but I'm not sure I feel qualified to review it...

What are the implications of setting the flag when signing? I assume this doesn't cause us to immediately start making bad signatures as soon as we update to the new binary? There was a comment but I wasn't sure what to make of it.

sanket1729 (approved):
tACK b888f42270bb0ef94a3b324e52591da06305a95d.

stevenroose:
> What are the implications of setting the flag when signing? I assume this doesn't cause us to immediately start making bad signatures as soon as we update to the new binary? There was a comment but I wasn't sure what to make of it.

We don't set the sighash flag when signing, we set the "script verification flag" that indicates that the sighash flag should be supported. So users that manually set the sighash flag can use it. This is because in some places we don't have access to the chain to know if dynafed already activated.

We should have a separate PR in a subsequent release that use the new sighash by default. Maybe we better wait with that until after dynafed activated.

src/script/interpreter.cpp
```diff
@@ -1529,11 +1538,23 @@ class CTransactionSignatureSerializer
     /** Serialize an output of txTo */
     template<typename S>
     void SerializeOutput(S &s, unsigned int nOutput) const {
-        if (fHashSingle && nOutput != nIn)
+        if (fHashSingle && nOutput != nIn) {
             // Do not lock-in the txout payee at other indices as txin
             ::Serialize(s, CTxOut());
```
roconnor-blockstream, line 1543: Legacy HashSingle should commit to the range proofs since that is what the BIP143 variant does.

sanket1729: It does right? Or is this something with the case where `nIn > txTo.vout.size()` ?

roconnor-blockstream: Indeed it does.  I was misreading this clause.

src/script/interpreter.cpp
```diff
@@ -1529,11 +1538,23 @@ class CTransactionSignatureSerializer
...
             ::Serialize(s, txTo.vout[nOutput]);
+
+            // Serialize rangeproof
+            if (fRangeproof) {
+                if (nOutput < txTo.witness.vtxoutwit.size()) {
+                    ::Serialize(s, txTo.witness.vtxoutwit[nOutput].vchRangeproof);
+                    ::Serialize(s, txTo.witness.vtxoutwit[nOutput].vchSurjectionproof);
+                } else {
+                    ::Serialize(s, (unsigned char) 0);
+                    ::Serialize(s, (unsigned char) 0);
```
apoelstra, line 1554: nb this branch is not covered by the functional test

PR #958, opened 2021-02-10
Some minor dynafed improvements

stevenroose:
- support blocksigners specifying their own activation to make coordination a bit more flexible
- don't force fedpeg script to change in order to change the block signing script

PR #962, opened 2021-03-02
[elements-0.18] Some minor dynafed improvements

stevenroose:
Backport of #958.


apoelstra:
concept ACK - verified that backport was done correctly (they are the same commits)

PR #963, opened 2021-03-02
[elements-0.18]  Add SIGHASH_RANGEPROOF support

stevenroose:
Backport of #960.


apoelstra:
concept ACK - verified that backport was done correctly (they are the same commits)

PR #954, opened 2021-01-06
Add semi-automated merge-review tool for Elements upstream merge proc…

gwillen:
Note: I'm doing this a bit backwards, in that the script which _generates_ these merges, as well as the documentation of how they are done, is not yet PR'ed. Hopefully that will come next.

PR #964, opened 2021-03-03
Process for elements upstream tracking / merging

gwillen:
This goes along with #954 (which is a tool for assisting the review process) and #935 (which was created using approximately this process.)

It's already had some internal review inside Blockstream, and probably should just get merged now, as a description of the process we're currently trying out; then we can have followup PRs once we know how we feel about it / if we want to revise it.

@stevenroose ?


apoelstra:
This all looks great to me.

PR #967, opened 2021-03-04
Fix qt build issues for Fedora 33

philippem:


doc/build-unix.md
```diff
@@ -134,9 +134,9 @@ Optional:
 
     sudo dnf install miniupnpc-devel
 
-To build with Qt 5 you need the following:
+To build elements-qt you need the following:
```
stevenroose, line 137: Hmm, it seems unnecessary to introduce a diff with upstream here, while the original sentence is correct.

doc/build-unix.md
```diff
@@ -134,9 +134,9 @@ Optional:
...
 
-    sudo dnf install qt5-qttools-devel qt5-qtbase-devel protobuf-devel
+    sudo dnf install qt5-qttools-devel qt5-qtbase-devel protobuf-devel ImageMagick librsvg2-tools
```
stevenroose, line 139: Is this something that is also added upstream? Or is this needed for us because of the icon changes that Luke introduced?

src/qt/trafficgraphwidget.cpp
```diff
@@ -7,6 +7,7 @@
 #include <qt/clientmodel.h>
 
 #include <QPainter>
+#include <QPainterPath>
```
stevenroose, line 10: Is this something that upstream also added?

doc/build-unix.md
apoelstra: This is needed because of the icon changes. These are removed/reverted by #935. Let's please not dig this hole any deeper here.

src/qt/trafficgraphwidget.cpp
apoelstra: Yes. It is needed to compile with recent versions of qt. (And even upstream, all versions of Core from 0.18 and before fail to compile because of this.)

philippem:
Closing in favor of upstream changes in #935

PR #969, opened 2021-03-10
Test duration of fedpeg validity in dynafed test

stevenroose:
I noticed this scenario was not really tested in the test


apoelstra (approved):
lol this logic is pretty rough to follow. But great job covering these cases!

PR #973, opened 2021-03-15
[elements-0.18] Bump version to 0.18.1.11

stevenroose:
Pending on inclusion of https://github.com/ElementsProject/elements/pull/972 and https://github.com/ElementsProject/elements/pull/971.


gwillen:
utACK. (I assume we have a scheme for this which you are follwing.)

PR #982, opened 2021-04-07
Add liquidv1test chainparams.

gwillen:
Add the `liquidv1test` chain, which is meant to more closely track the `liquidv1` chain for realistic testing in CI and on test networks (vs `elementsregtest`, a "custom chain" which has diverged a bit more than I'm comfortable with.)

I also fixed a few stray things that looked like minor mistakes to me (uses of the global `gArgs` instead of the parameter `args`, and the `-chain` help text not mentioning `liquidv1` as the default.)

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -675,14 +675,14 @@ class CCustomParams : public CRegTestParams {
         // ELEMENTS fields
 
         // Determines type of genesis block
-        consensus.genesis_style = gArgs.GetArg("-con_genesis_style", "elements");
+        consensus.genesis_style = args.GetArg("-con_genesis_style", "elements");
 
         // Block signing encumberance script, default of 51 aka OP_TRUE
-        std::vector<unsigned char> sign_bytes = ParseHex(gArgs.GetArg("-signblockscript", "51"));
+        std::vector<unsigned char> sign_bytes = ParseHex(args.GetArg("-signblockscript", "51"));
         consensus.signblockscript = CScript(sign_bytes.begin(), sign_bytes.end());
         // Default signature size is the size of dummy push, and single 72 byte DER signature
-        consensus.max_block_signature_size = gArgs.GetArg("-con_max_block_sig_size", 74);
-        g_signed_blocks = gArgs.GetBoolArg("-con_signed_blocks", true);
+        consensus.max_block_signature_size = args.GetArg("-con_max_block_sig_size", 74);
+        g_signed_blocks = args.GetBoolArg("-con_signed_blocks", true);
 
         // Note: These globals are needed to avoid circular dependencies.
         // Default to true for custom chains.
@@ -700,7 +700,7 @@ class CCustomParams : public CRegTestParams {
         // Custom chains connect coinbase outputs to db by default
         consensus.connect_genesis_outputs = args.GetArg("-con_connect_genesis_outputs", true);
 
-        initialFreeCoins = gArgs.GetArg("-initialfreecoins", 0);
+        initialFreeCoins = args.GetArg("-initialfreecoins", 0);
 
         anyonecanspend_aremine = args.GetBoolArg("-anyonecanspendaremine", true);
 
@@ -742,8 +742,8 @@ class CCustomParams : public CRegTestParams {
 
         // Subsidy asset, like policyAsset, defaults to the pegged_asset
         consensus.subsidy_asset = consensus.pegged_asset;
-        if (gArgs.IsArgSet("-subsidyasset")) {
-            consensus.subsidy_asset = CAsset(uint256S(gArgs.GetArg("-subsidyasset", "0x00")));
+        if (args.IsArgSet("-subsidyasset")) {
+            consensus.subsidy_asset = CAsset(uint256S(args.GetArg("-subsidyasset", "0x00")));
         }
 
         consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].bit = 25;
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -990,7 +990,293 @@ public:
         consensus.hashGenesisBlock = genesis.GetHash();
         assert(consensus.hashGenesisBlock.GetHex() == "1466275836220db2944ca059a3a10ef6fd2ea684b0688d2c379296888a206003");
     }
+};
 
+/**
+ * New: Liquid v1 testing, as close to prod as possible while still being customizable.
+ */
+class CLiquidV1TestParams : public CLiquidV1Params {
+public:
+    explicit CLiquidV1TestParams(ArgsManager& args)
+    {
+        // Our goal here is to override ONLY the things from liquidv1 that make no sense for a test chain / which are pointless and burdensome to require people to override manually.
+
+        strNetworkID = "liquidv1test";
+
+        // required for some tests to consider this 'regtest'
+        fMineBlocksOnDemand = true;
+
+        vSeeds.clear();  // No network seeds
+        vFixedSeeds.clear();  // No network seeds
+
+        // 51 means OP_TRUE, this can be overridden on the commandline
+        std::vector<unsigned char> sign_bytes = ParseHex("51");
+        consensus.signblockscript = CScript(sign_bytes.begin(), sign_bytes.end());
+
+        // Do not mandate a specific destination for fees in testing
+        consensus.mandatory_coinbase_destination = CScript(); // Blank script allows any coinbase destination
+
+        // The bitcoin regtest genesis blockhash is the default, not the mainchain
+        parentGenesisBlockHash = uint256S("0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206");
+        const bool parent_genesis_is_null = parentGenesisBlockHash == uint256();
+        assert(consensus.has_parent_chain != parent_genesis_is_null);
+
+        // This is the regtest limit, not the mainchain limit.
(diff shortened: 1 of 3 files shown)


gwillen:
I tested this by manually editing `functionary/tests/specs/wm-regular-pak.json` to change `sidechain` from `elementsregtest` to `liquidv1test`, which passes. I haven't tried doing this systematically with all available tests.

If we want to switch to this by default, one option would be to name it "elementsregtest" instead of "liquidv1test", since that's the "custom" name we use for most of our testing. However, it's not guaranteed we used it everywhere, and if we gave that name in any external documentation then it may be a bad idea to change its behavior.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+
+        vSeeds.clear();  // No network seeds
+        vFixedSeeds.clear();  // No network seeds
+
+        // 51 means OP_TRUE, this can be overridden on the commandline
+        std::vector<unsigned char> sign_bytes = ParseHex("51");
+        consensus.signblockscript = CScript(sign_bytes.begin(), sign_bytes.end());
+
+        // Do not mandate a specific destination for fees in testing
+        consensus.mandatory_coinbase_destination = CScript(); // Blank script allows any coinbase destination
```
stevenroose, line 1017: I don't think this one necessarily has to be changed.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+        parentGenesisBlockHash = uint256S("0f9188f13cb7b2c71f2a335e3a4fc328bf5beb436012afca590b1a11466e2206");
+        const bool parent_genesis_is_null = parentGenesisBlockHash == uint256();
+        assert(consensus.has_parent_chain != parent_genesis_is_null);
+
+        // This is the regtest limit, not the mainchain limit.
+        consensus.parentChainPowLimit = uint256S("7fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
+        consensus.parent_chain_signblockscript = CScript(); // It has PoW
+
+        // Default to 8, not 100, for expedited testing.
+        consensus.pegin_min_depth = DEFAULT_PEGIN_CONFIRMATION_DEPTH;
```
stevenroose, line 1029: I don't think this one necessarily has to be changed.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+        bech32_hrp = args.GetArg("-bech32_hrp", "ert");
+        blech32_hrp = args.GetArg("-blech32_hrp", "el");
+        base58Prefixes[PUBKEY_ADDRESS] = std::vector<unsigned char>(1, args.GetArg("-pubkeyprefix", 235));
+        base58Prefixes[SCRIPT_ADDRESS] = std::vector<unsigned char>(1, args.GetArg("-scriptprefix", 75));
+        base58Prefixes[BLINDED_ADDRESS] = std::vector<unsigned char>(1, args.GetArg("-blindedprefix", 4));
+        base58Prefixes[SECRET_KEY] =     std::vector<unsigned char>(1, args.GetArg("-secretprefix", 239));
+        base58Prefixes[PARENT_PUBKEY_ADDRESS] = std::vector<unsigned char>(1, args.GetArg("-parentpubkeyprefix", 111));
+        base58Prefixes[PARENT_SCRIPT_ADDRESS] = std::vector<unsigned char>(1, args.GetArg("-parentscriptprefix", 196));
+        parent_bech32_hrp = args.GetArg("-parent_bech32_hrp", "bcrt");
+        parent_blech32_hrp = args.GetArg("-parent_blech32_hrp", "bcrt");
```
stevenroose, line 1047: I don't think these necessarily have to be changed.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+
+        std::string extpubprefix = args.GetArg("-extpubkeyprefix", "043587CF");
+        assert(IsHex(extpubprefix) && extpubprefix.size() == 8 && "-extpubkeyprefix must be hex string of length 8");
+        base58Prefixes[EXT_PUBLIC_KEY] = ParseHex(extpubprefix);
+
+        std::string extprvprefix = args.GetArg("-extprvkeyprefix", "04358394");
+        assert(IsHex(extprvprefix) && extprvprefix.size() == 8 && "-extprvkeyprefix must be hex string of length 8");
+        base58Prefixes[EXT_SECRET_KEY] = ParseHex(extprvprefix);
```
stevenroose, line 1055: I don't think these necessarily have to be changed.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+
+        const std::string magic_str = args.GetArg("-pchmessagestart", "FABFB5DA");
+        assert(IsHex(magic_str) && magic_str.size() == 8 && "-pchmessagestart must be hex string of length 8");
+        const std::vector<unsigned char> magic_byte = ParseHex(magic_str);
+        std::copy(begin(magic_byte), end(magic_byte), pchMessageStart);
+        // END magic numbers
```
stevenroose, line 1061: What would be the result of not changing the magic numbers? We cleared the DNS seeds and addr seeds, so I don't think we'd accidentally mix with the liquid network. Plus the genesis is different so we'd obviously be kicked by all liquidv1 nodes.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
 
+        UpdateFromArgs(args);
+        SetGenesisBlock();
+        consensus.hashGenesisBlock = genesis.GetHash();
+    }
+
+    // As much as possible here, our goal is to:
+    // - Allow overriding anything that can be overridden in CCustomParams;
```
stevenroose, line 1069: I'm not sure we should have this goal. We have CCustomParams for that. I thought this chain's goal was to have something as close as possible to liquidv1 with only the necessary changes to be able to do some testing.

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+    // - Leave everything alone unless an argument / config parameter was given.
+    // This is unlike the CCustomParams UpdateFromArgs method, which has lots of defaults in it.
+    void UpdateFromArgs(ArgsManager& args)
```
stevenroose, line 1072: Hmm so in this method, you allow basically everything to be changed. Making it just another CCustomParams just with different fallback defaults. I'm not sure if that should be our intention here. Personally, I'd be happy with just allowing to change the liquidv1 consensus parts that really make testing impossible (i.e. signblockscript, fedpegscript, PAK and parent chain info). Everything else could stay the same.

If we want to do dynafed testing, maybe one exception could be to allow changing the dynafed activation, so we don't have to create 1,000,000 blocks to activate dynafed :sweat_smile: But I noticed that liquidv1 also allows that:
```c++
        consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime = gArgs.GetArg("-con_dyna_deploy_start", 1000000);
```

src/chainparams.cpp
```diff
@@ -990,7 +990,293 @@ class CLiquidV1Params : public CChainParams {
...
+        // CSV always active by default, unlike regtest
+        if (args.IsArgSet("-con_csv_deploy_start")) {
+            consensus.vDeployments[Consensus::DEPLOYMENT_CSV].bit = 0;
+            consensus.vDeployments[Consensus::DEPLOYMENT_CSV].nStartTime = args.GetArg("-con_csv_deploy_start", Consensus::BIP9Deployment::ALWAYS_ACTIVE);
+            consensus.vDeployments[Consensus::DEPLOYMENT_CSV].nTimeout = Consensus::BIP9Deployment::NO_TIMEOUT;
+        }
+    }
+
+    // XXX: This is copy-and-pasted from CCustomParams; sharing it would be better, but is annoying.
+    void SetGenesisBlock() {
```
stevenroose, line 1260: Since this is a clone of liquidv1, we can instead assert that `genesis_style` is `"elements"` and only have that part be in this method.

stevenroose (commented):
Hmm, so you went on kind of a different route here than I expected we would.

I think my intention was a bit closer to your first line there, saying "Our goal here is to override ONLY the things from liquidv1 that make no sense for a test chain / which are pointless and burdensome to require people to override manually.".

I think we have CCustomParams to create arbitrary chains for testing various things. Which is quite powerful and perhaps it would make sense for us to public an `elements.conf` that makes CCustomParams emulate a Liquid-like environment. Or perhaps even an elements.conf that basically makes Liquidv1 (perhaps we can have a test that asserts that?).

But I thought the goal of `liquidv1test` was to have a chain that is by definition (through inheritance) identical to liquidv1, with really only the minimal changes required to do testing in which we want Liquid behavior? Which IMO is *just* the initial federation setup + parent chain definition (i.e. signblockscript, fedpegscript, PAK, and parent chain hash and PoW info).

stevenroose (approved):
utACK

src/chainparams.cpp
apoelstra: If we somehow did connect to the Liquid network the result would be a lot of mutual peer banning and messy logs.

gwillen:
Fixed build errors that steven pointed out, sorry. (I apparently did not check that it still built, after I rebased onto master. :facepalm:)

PR #986, opened 2021-04-28
Allow setting the block version from the config file

stevenroose:
Instead of only allowing this for regtest chains, we allow it for all chains.


gwillen:
So just to check my understanding -- the reason we have to do this in `CreateNewBlock`, rather than in the ChainParams for LiquidV1, is so we can decline to _signal_ for Dynafed with BIP9 in our own blocks, while still understanding/accepting if OTHER blocks signal? Which is not something that the existing BIP9 signalling code can do?

(E.g. if we changed `consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime`, that would not JUST prevent us from signaling, it would ALSO prevent us from accepting blocks with the dynafed hardfork bit set before that time, even if 75% signalling was achieved? Or is that counted some other way?)

gwillen:
BTW it looks like feature_dynafed.py and mining_basic.py are failing in CI. I assume that is in fact real breakage from this change that we need to figure out.

gwillen:
LMK if I can help with the test issues at all.

gwillen:
That looks great, utACK.

PR #988, opened 2021-04-30
[elements-0.18]  Add liquidv1test chainparams

stevenroose:
Backport of #982.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -542,14 +542,14 @@ class CCustomParams : public CRegTestParams {
         // ELEMENTS fields
 
         // Determines type of genesis block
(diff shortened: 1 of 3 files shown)


gwillen:
Tested this on a test network, it does what I expect (I can switch from a custom chain named "liquidv1test" to using the new liquidv1test chain, by swapping from the .11 binary to this one, and everything still works. If I change an arbitrary parameter, parentgenesisblockhash, in the definition of liquidv1test and rebuild, it fails as expected. If I override it on the commandline back to the correct value, it works again.)

(Technically I tested using my feature-liquidv1test-chainparams branch, but I checked that this appears to be just a rebase of it.)

gwillen:
At least one of the Travis builds timed out, I don't seem to have permission to re-trigger it, @stevenroose?

PR #989, opened 2021-04-30
[elements-0.18]  Undo default signalling behavior for dynafed unless enabled

stevenroose:
Backport of #986.


gwillen:
utACK b104f95 for the backport, after #986 is merged, assuming that b7a116d is the final version that goes in for #986 .

Issue #985, opened 2021-04-21
Add contract_hash to issueasset


allenpiscitello:
rawissueasset can be painful to work with as it requires a previously constructed transaction. This means the user must overestimate fees in the initial transaction, and may overpay.

This is mostly needed because issueasset does not support including a contract_hash, which is needed for the registry. Adding contract_hash as a parameter to issueasset will allow simpler and more efficient asset issuance for assets that will go in the registry.

PR #993, opened 2021-05-07
[elements-0.18] Support supplying contract hash to issueasset RPC

stevenroose:
Relating https://github.com/ElementsProject/elements/issues/985.

Needs forward port.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -6133,16 +6133,16 @@ UniValue issueasset(const JSONRPCRequest& request)
         return NullUniValue;
     }
 
-    if (request.fHelp || request.params.size() < 2 || request.params.size() > 3)
+    if (request.fHelp || request.params.size() < 2 || request.params.size() > 4)
         throw std::runtime_error(
             RPCHelpMan{"issueasset",
                 "\nCreate an asset. Must have funds in wallet to do so. Returns asset hex id.\n"
-                "For more fine-grained control such as non-empty contract-hashes to commit\n"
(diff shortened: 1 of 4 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -6174,6 +6174,12 @@ UniValue issueasset(const JSONRPCRequest& request)
 
     bool blind_issuances = request.params.size() < 3 || request.params[2].get_bool();
 
+    // Check for optional contract to hash into definition
+    uint256 contract_hash;
```
philippem, line 6178: ok to leave unitialized?

stevenroose: Yeah the default value is 0. I am assuming C++ calls the default constructor so that this is equivalent to `uint256 contact_hash = uint256();`..

stevenroose: Though we usually don't do that though

```
$ rgcpp "uint256 [a-z_]+;" | wc -l
147
$ rgcpp "uint256 [a-z_]+\{\};" | wc -l
6
$ rgcpp "uint256 [a-z_]+ = uint256\(\);" | wc -l
1
```

stevenroose: So would it be the same and is it more a principle difference? Or is the uninitiated actually uninitiated and random memory?

philippem:
utack

src/wallet/rpcwallet.cpp
apoelstra: It is just a principle difference but Core is increasingly moving toward using `{}` everywhere to reduce the cognitive load associated with C++ ambiguity.

PR #992, opened 2021-05-07
Update URL to download old QT SDK

gwillen:

PR #995, opened 2021-05-07
[elements 0.18]  Update URL to download old QT SDK

stevenroose:
Backport of #992.


stevenroose:
Not sure if this is needed, though. It might be so that the gitian builder scripts always take the descriptors from the master branch..

PR #1000, opened 2021-05-14
Don't embed a build-id when building libdmg-hfsplus

mword:
This is a change that is included in more advanced branches that should help with gitian reproducibility on osx.


delta1:
Reproducible builds are now done with Guix, which doesn't seem to have this issue

PR #1002, opened 2021-05-20
Implement Taproot Sighash

sanket1729:

PR #1003, opened 2021-05-21
Taproot tests using python bindings for secp256k1-zkp

sanket1729:
#1002 only implements tests for taproot sighash that spend from non-confidential txouts. Since, we do not yet have wallet support, it is not possible to use regular blinding rpcs to spend from confidential txouts. This adds support for python test framework for creating blinding transactions using bindings from @dgpv. 

@apoelstra suggested that we should not be merging these bindings here because 1) they will create platform/architecture dependency for tests. and 2) the bindings rely on .so file and we do not want to touch the build system to enforce this in CI. 

Regardless, this PR can aid in the review of #1002 or can help writing blinding/CT logic in the future too. The test has instructions on how to run it. 

The code quality of the test is not the best, but if we want to merge this I can work on improving the test. 

Marking as draft because we don't expect these tests to be merged.


apoelstra:
Now that #1002 is merged I am going to close this.

For the avoidance of doubt, this is a no-merge because it includes Python bindings to libsecp256k1-zkp which put additional system requirements on build machines, and which are forked off of a 3rd-party (Dmitry)'s project, which would mean an ongoing maintanance/syncing requirem.nt

PR #1006, opened 2021-06-03
set nNonce to Null in CoinsView

### LINKED ISSUE #1004: txout nonce is inconsistently stored in coinsview
We currently do not de/serialize the txout nonce in the `Coin` class, meaning that after restarting the node, all nonces in the utxoset are cleared, resulting in a network where everyone has an (inconsequentially) different version of the UTXO set.

Currently the nonce is not used for anything consensus-critical, but it may be in the future --- for example, we might want input nonces covered by a future sighash version --- so this is a massive footgun.

sanket1729:
Fixes #1004


apoelstra:
Closing in favor of #1008; I think the assertions are in the wrong place and I have a slightly cleaner way of nulling out the nonce.

PR #1008, opened 2021-06-08
set nNonce to Null in CoinsView

apoelstra:
Alternative to #1006 

Fixes #1003

### DIFF
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -51,8 +51,10 @@ static inline CCoinsMapKey native_key(const COutPoint& outpoint) {
 
 CCoinsMap::iterator CCoinsViewCache::FetchCoin(const COutPoint &outpoint) const {
     CCoinsMap::iterator it = cacheCoins.find(native_key(outpoint));
-    if (it != cacheCoins.end())
+    if (it != cacheCoins.end()) {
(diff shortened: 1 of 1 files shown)


sanket1729 (approved):
utACK a70ff3cfd7a2cd349c71cd229cf7a642e6070893

apoelstra:
Lol, some of the new assertions are triggering in CI

apoelstra:
Sigh, so, the `CCoinsView` is a base class of two things -- the "UTXO set" which is saved to disk in the `coinstate/` directory, where we want to erase the nonce, and the mempool tx db, where we do **not** want to erase the nonce (and there are existing assertions that prevent us from doing so).

stevenroose:
Any update on solving this?

roconnor-blockstream:
Why don't we want to erase it from the mempool?

apoelstra:
@stevenroose no update. I think it might be practically impossible.

@roconnor-blockstream because of tangled inheritance relationships between various mempool-like structures. If you do what you're suggesting you will get far away assertion failures in consensus code where they might be meaningful.

apoelstra:
I think we should just live with this bug.

PR #996, opened 2021-05-10
[forward port] Add logging for dynafed activation and transitions

stevenroose:
Forward port and fixup of https://github.com/ElementsProject/elements/pull/994.
Closes https://github.com/ElementsProject/elements/issues/981.

PR #1009, opened 2021-06-10
Fix dynafed logging to not duplicate messages

stevenroose:
This brings it up to date with the fixed forward port in https://github.com/ElementsProject/elements/pull/996.


apoelstra (approved):
utack 2af9a391e2be95545ddf0bec5a94cf34195b32f4

apoelstra:
Would be good to rebase this when #1012 lands so we can get CI passing.

PR #1012, opened 2021-06-18
fuzz: backport fixes to fuzzer errors

apoelstra:
Backport of https://github.com/bitcoin/bitcoin/pull/22279 and https://github.com/bitcoin/bitcoin/pull/22002 and https://github.com/bitcoin/bitcoin/pull/22137

### DIFF
--- a/src/test/fuzz/base_encode_decode.cpp
+++ b/src/test/fuzz/base_encode_decode.cpp
@@ -14,6 +14,11 @@
 #include <string>
 #include <vector>
 
+void initialize()
+{
+    static const ECCVerifyHandle verify_handle;
+}
+
 void test_one_input(const std::vector<uint8_t>& buffer)
 {
     const std::string random_encoded_string(buffer.begin(), buffer.end());
--- a/src/test/util_tests.cpp
+++ b/src/test/util_tests.cpp
@@ -318,6 +318,25 @@ BOOST_FIXTURE_TEST_CASE(util_CheckValue, CheckValueTest)
     CheckValue(M::ALLOW_ANY, "-value=abc", Expect{"abc"}.String("abc").Int(0).Bool(false).List({"abc"}));
 }
 
+struct NoIncludeConfTest {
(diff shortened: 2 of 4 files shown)


apoelstra:
Our CI uses Bitcoin's fuzzer corpus, meaning that when they add test vectors to it that they've fixed, our CI breaks until we fix them too :).

apoelstra:
For some reason I needed to edit Marco's last commit with 
```
diff --git a/src/test/util_tests.cpp b/src/test/util_tests.cpp
index ec996dbe7f6..c7bb028b111 100644
--- a/src/test/util_tests.cpp
+++ b/src/test/util_tests.cpp
@@ -323,7 +323,7 @@ struct NoIncludeConfTest {
     {
         TestArgsManager test;
         test.SetupArgs({{"-includeconf", ArgsManager::ALLOW_ANY}});
-        std::array argv{"ignored", arg};
+        std::array<const char*, 2> argv{"ignored", arg};
         std::string error;
         (void)test.ParseParameters(argv.size(), argv.data(), error);
         return error;
```
to avoid compiler errors "missing template arguments before `argv`". I don't know why this is needed here but not upstream.

gwillen:
Driveby quick review: utACK.

The `std::array` template thing appears to be a C++14 vs C++17 change. Is upstream on 17? Are we on 14 and should switch to 17?

jonasnick (commented):
Not familiar with area of the code, but I can confirm that these are backports

utACK 0896d0a278c2e03cba8e9def912725387538650e

apoelstra:
@gwillen if upstream has switched to 17, it was between 0.21 and 22.0 and we'll follow them as part of the 22.0 rebase.

gwillen:
> @gwillen if upstream has switched to 17, it was between 0.21 and 22.0 and we'll follow them as part of the 22.0 rebase.

Oh, right, and that makes perfect sense because this is a backport. 👍

PR #1017, opened 2021-07-09
test: update `createrawtransaction` call in feature_taphash_pegins_issuances.py

apoelstra:
Should describe the outputs as an array rather than as an object. (The old behavior has long been deprecated but was eliminated entirely in #900.)

Fixes functional tests which were broken in master by simultaneous merge of #1002 and #900.


stevenroose:
utACK 1c883db83ea564129cf0241797ea231b81e7f744

stevenroose:
utACK 1c883db83ea564129cf0241797ea231b81e7f744

PR #1022, opened 2021-07-20
When validation is waiting for parent chain daemon, "stall".

gwillen:
Currently, if -validatepegin is given, and block validation can't proceed
because the parent chain is not synced, we mark the block invalid and put
it in a queue to be "revalidated" later. Unfortunately, marking a block
invalid has downstream consequences, in particular causing descendant blocks
to be marked invalid, which are not currently fixed by the queue.

Instead, we'll use a different strategy: if the mainchain daemon isn't
sufficiently synced to validate a block, we will "stall" connecting that
block to the chain, and have ActivateBestChain simply keep the tip at the
previous block until we're ready.

We can still download and validate (partly) blocks past this point while
we're waiting. They will be connected once the parent chain daemon catches
up.


src/validation.cpp
```diff
@@ -1936,6 +1937,35 @@ static int64_t nTimeCallbacks = 0;
...
+        const CTransaction &tx = *(block.vtx[i]);
+
+        if (!tx.IsCoinBase()) {
+            for (unsigned int i = 0; i < tx.vin.size(); ++i) {
+                if (tx.vin[i].m_is_pegin) {
+                    std::string err;
+                    // IsValidPeginWitness technically checks more than just "ripeness". As a result,
+                    // if we get a validly-signed block which contains an ill-formed pegin witness,
+                    // this test will fail, and it will be treated as "unripe" rather than invalid,
+                    // resulting in a stall rather than rejecting the block. This is reasonable, because
```
apoelstra: trailing whitespace

apoelstra: Also, I think this comment is incorrect because in the case that `IsValidPeginWitness` fails, we match on the error string.

src/init.cpp
```diff
@@ -1278,6 +1277,61 @@ bool AppInitLockDataDirectory()
...
+    // Retry until a non-RPC_IN_WARMUP result
+    while (true) {
+        try {
+            // The first thing we have to check is the version of the node.
+            UniValue params(UniValue::VARR);
+            UniValue reply = CallMainChainRPC("getnetworkinfo", params);
+            UniValue error = reply["error"];
+            if (!error.isNull()) {
+                // On the first call, it's possible to node is still in
+                // warmup; in that case, just wait and retry.
```
apoelstra, line 1295: Should drop `On the first call` from this comment since you removed the `init` flag

src/validation.cpp
apoelstra: Ah, the comment is removed 2 commits later

apoelstra (approved):
ack 9c5ea27f99b119cf4b1bb5e1dfaf9577f6e3efa4

apoelstra:
rescinding ACK

src/validation.cpp
```diff
@@ -5337,63 +5337,3 @@ class CMainCleanup
...
 } instance_of_cmaincleanup;
-
-// ELEMENTS:
-/* This function checks that the RPC connection to the parent chain node
- * can be attained, and is returning back reasonable answers.
- */
-bool MainchainRPCCheck(const bool init)
-{
-    // Check for working and valid rpc
-    if (gArgs.GetBoolArg("-validatepegin", Params().GetConsensus().has_parent_chain)) {
```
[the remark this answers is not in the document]
gwillen: Because it's only ever called inside a conditional that already checks the same condition, so it was redundant.

(It sounded like you had an example of this failing though -- can you provide details?)

apoelstra:
I rescind my rescindation. My previous ACK is still good.

apoelstra:
I rescind my ACK for real this time.

We can't combine the schedulers, cf https://github.com/ElementsProject/elements/pull/647

apoelstra:
To expand on the need for two schedulers: this code puts `ActivateBestChain` on a 30-second timer. The old code did something similar though more involved with its 2-minute "reverification scheduler". However, `ActivateBestChain` calls `LimitValidationInterfaceQueue` before taking any other action, which blocks on the validation queue being cleared. But the actions in this queue are executed on the scheduler thread ... the same thread that is sitting blocked! So if we want to call ABC periodically we need to give it its own thread.

apoelstra:
cc @gwillen I squashed a couple commits and un-removed the second scheduling thread

gwillen:
@apoelstra Thanks, looks great. If that fixes your tests then I'm 100% happy with it.

PR #1025, opened 2021-08-05
Introduce blech32m format for v1+ witness programs

apoelstra:
Includes backports of https://github.com/bitcoin/bitcoin/pull/20832 (1 commit) and https://github.com/bitcoin/bitcoin/pull/20861 (5 commits)


real-or-random:
I'm late to the party but I still think "blech" is a bad name because it means "(metal) sheet" in Germany and is associated with cheap metal and often has a negative connotation (e.g., sometimes people ironically call the worthless fourth place in a competition "Blech", i.e., you missed the gold/silver/bronze medals.)

apoelstra:
It was something of a joke, because it also has a negative connotation in English (it's basically a gross sound you might make, e.g. see [this picture of Bill the Cat from Bloom County](https://bloomcounty.fandom.com/wiki/Bill_the_Cat))

src/blech32.cpp
```diff
@@ -122,21 +139,24 @@ data ExpandHRP(const std::string& hrp)
...
-bool VerifyChecksum(const std::string& hrp, const data& values)
+Encoding VerifyChecksum(const std::string& hrp, const data& values)
 {
     // PolyMod computes what value to xor into the final values to make the checksum 0. However,
     // if we required that the checksum was 0, it would be the case that appending a 0 to a valid
-    // list of values would result in a new valid list. For that reason, Blech32 requires the
-    // resulting checksum to be 1 instead.
-    return PolyMod(Cat(ExpandHRP(hrp), values)) == 1;
+    // list of values would result in a new valid list. For that reason, Bech32 requires the
+    // resulting checksum to be 1 instead. In Bech32m, this constant was amended.
```
gwillen, line 147: Nit: Bech->Blech on this line and above (but it doesn't really matter.)

src/blech32.cpp
```diff
@@ -158,7 +182,7 @@ std::string Encode(const std::string& hrp, const data& values) {
 }
 
 /** Decode a Blech32 string. */
```
gwillen, line 184: Nit: missed comment change ("Blech32" -> "Blech32 or Blech32m")

gwillen:
Ugh, I got confused about what diff I was looking at and some of my comments are wrong, stand by and ignore them for now. Sorry!

src/blech32.h
```diff
@@ -19,16 +20,33 @@
...
 {
 
+enum class Encoding {
+    INVALID,  //!< Failed decoding
+
+    BLECH32,  //!< Original blech32
+    BLECH32M, //!< Blech32m tweaked a la bech32
+};
+
 /** Encode a Bech32 string. Returns the empty string in case of failure. */
```
gwillen, line 30: Nit: Bech->Blech.

(Upstream has a longer comment here, but this was already true before this change -- I don't know if it's untrue for us for some reason. Converted for our use it would be:

`/** Encode a Blech32 or Blech32m string. If hrp contains uppercase characters, this will cause an
 *  assertion error. Encoding must be one of BLECH32 or BLECH32M. */`)

src/key_io.cpp
```diff
@@ -114,7 +116,7 @@ class DestinationEncoder : public boost::static_visitor<std::string>
             bytes.insert(bytes.end(), id.program, id.program + id.length);
             ConvertBits<8, 5, true>([&](unsigned char c) { data.push_back(c); }, bytes.begin(), bytes.end());
             const std::string& hrp = for_parent ? m_params.ParentBlech32HRP() : m_params.Blech32HRP();
-            return blech32::Encode(hrp, data);
+            return blech32::Encode(blech32::Encoding::BLECH32, hrp, data);
```
gwillen, line 119: Should this be BLECH32M? Compare to 5 lines down.

src/key_io.cpp
```diff
@@ -241,14 +243,29 @@ CTxDestination DecodeDestination(const std::string& str, const CChainParams& par
     data.clear();
     auto blech = blech32::Decode(str);
     const std::string& bl_hrp = for_parent ? params.ParentBlech32HRP() : params.Blech32HRP();
-    if (blech.second.size() > 0 && blech.first == bl_hrp) {
+    if (blech.data.size() > 0) {
```
gwillen, line 246: Lost the check here for `blech.encoding == blech32::Encoding::BECH32 || blech.encoding == blech32::Encoding::BECH32M` ?

gwillen: (Added on the bech32 side but not correspondingly here.)

gwillen:
FYI, It would hard for me to catch any blech32-specific stuff around blinding keys in DecodeDestination, the bit-fiddling is extremely fiddly and doesn't parallel anything in bech32. So maybe double-check that part.

gwillen:
Ok, I have checked all the bits that weren't backports, comments above.

gwillen:
Checked the cherry-pick of bitcoin/bitcoin#20832 using my usual trick. (Redo it myself, with `merge.conflictstyle=diff3`, make a temporary commit with the conflict markers, diff against your resolution.) Looks good.

gwillen:
Checked the other backport the same way, looks good. 

I notice that we removed an anonymous namespace in `bech32.h`, present in the upstream version, which was hiding some private helper functions. It didn't change in this version, it was already like that, so I'm content to leave it alone; but I'm curious why we did that.

I've now looked at everything, and it's all good other than my comments above.

src/key_io.cpp
apoelstra: Yep, good catch, thanks

src/key_io.cpp
apoelstra: Yep, good catch. I think I confused myself here thinking that the extra address formats in Elements meant that I didn't need/want this check. But we do, the check is there to make sure that `blech32::Decode` doesn't return an error.

apoelstra:
Addressed Glenn's comments and changed the constant as per @roconnor-blockstream's recommendation.

gwillen:
utACK c72c949.

PR #1029, opened 2021-09-01
Elements 0.21: backports from Core 0.21.0

apoelstra:
This is a cherry-pick of all the commits from Bitcoin Core 0.21.1 that are not on master, **except** those related to bech32m and Speedy Trial.

**TODO** (will do each in a separate additional commit, no need to delay review)
* [x] regenerate manpages
* [x] bump version in configure.ac


apoelstra:
Rebased.

cc @gwillen this is ready for review

gwillen:
Checked that all the commits (up to 321ff88) are the thing they claim to be. (I reviewed the `git range-diff` from the commit here, against the `(cherry-picked from commit X)` commit ID from upstream's release branch (I assume); except for the last four commits which were missing that line, and I instead diffed against the upstream `Rebased-From: X` commit ID, presumably referring to the original pull request.)

I didn't check yet that the things they claim to be are reasonable.

gwillen:
Oh, except e45e7c15e7928e6ad63bb214164367e1763bc4dc, where our diff is empty, whereas upstream's diff is adding a test case for transaction decoding (ambiguity between segwit and pre-segwit encoding format.) Possibly the offending issue just doesn't exist in elements, and you did that on purpose?

apoelstra:
Thank you! Let me redo the last four commits. I forgot to use `-x` with `git-cherry-pick`. And I will double-check the empty diff -- most likely the code was already in Elements due to an earlier backport. (These can't happen by accident, `git commit` won't work unless you explicitly give it `--allow-empty-commit`, so I definitely was aware of this at the time that I did it..)

apoelstra:
Replaced last four commits.

apoelstra:
Added doc update and release version

gwillen:
Rechecked that last four cherry-picks now look good (they match what they say they were cherry-picked from.)

Checked that the set of cherry-picked commits matches the non-merge commits in:

`git log v0.21.1 --not 831675c8dcc --graph`

after manually excluding the speedy trial ones, the bech32 ones, and the one-off ones by Wladimir doing release activities.

gwillen:
There's a small issue I spotted in the manpages, probably implying the same issue exists in the help text, which we're going to leave for -rc2. (The port looks wrong.)

```
-Connect to JSON\-RPC on <port> (default: 8332, testnet: 18332, regtest:
-18443)
+Connect to JSON\-RPC on <port> (default: 7041, testnet: 18332, signet:
+38332, regtest: 18443)
```

PR #1030, opened 2021-09-01
[forward port] When validation is waiting for parent chain daemon, "stall"

apoelstra:
Forward-port of #1022


apoelstra:
Blocked on fix to #1022.

apoelstra:
Ready for review @gwillen .. can probably just check that it is a forward-port of #1022

src/rpc/rawtransaction_util.cpp
```diff
@@ -545,8 +545,9 @@ bool ValidateTransactionPeginInputs(const CMutableTransaction& mtx, std::map<int
             continue;
         }
         // Report warning about immature peg-in though
-        if(txin.m_is_pegin && !IsValidPeginWitness(mtx.witness.vtxinwit[i].m_pegin_witness, fedpegscripts, txin.prevout, err, true)) {
-            CHECK_NONFATAL(err == "Needs more confirmations.");
+        bool depth_failed = false;
+        if(txin.m_is_pegin && !IsValidPeginWitness(mtx.witness.vtxinwit[i].m_pegin_witness, fedpegscripts, txin.prevout, err, true, &depth_failed)) {
+            assert(depth_failed);
```
gwillen, line 550: This check (the string error version) changed from `assert` to `CHECK_NONFATAL` since 0.18, apparently. I can't tell whether you changed it back to assert intentionally when rebasing my changes.

src/init.cpp
```diff
@@ -2079,29 +2133,34 @@ bool AppInitMain(const util::Ref& context, NodeContext& node, interfaces::BlockA
...
-        const std::string err_msg = "ERROR: elements is set to verify pegins but cannot get a valid response from the mainchain daemon. Please check debug.log for more information.\n\nIf you haven't setup a bitcoind please get the latest stable version from https://bitcoincore.org/en/download/ or if you do not need to validate pegins set in your elements configuration validatepegin=0";
-        // We fail immediately if this node has RPC server enabled
-        if (gArgs.GetBoolArg("-server", false)) {
-            InitError(Untranslated(err_msg));
-            return false;
-        } else {
-            // Or gently warn the user, and continue
-            InitError(Untranslated(err_msg));
-            gArgs.SoftSetArg("-validatepegin", "0");
+        if (!MainchainRPCCheck()) { //Initial check only
```
gwillen, line 2136: I'm having trouble reading the `range-diff` here, but I think you might have removed this (now outdated) comment yourself when rebasing my original PR, and then failed to carry that over to this PR.

gwillen:
utACK  da11d7b. I ran:

`git range-diff -w --color-moved 5b1f3cc00b3c279d94b29b4a1d8c41a7e2b25ad1..upstream/pr/1022 904054ffcb4a67905577a4eb8b256369111be387..upstream/pr/1030`

and checked that this appeared to be a faithful rebase of my work, which it is, subject to comments above (which are all nits.)

(It turns out that `-w` and `--color-moved` don't actually _work_ right with `git range-diff`, but that's a complaint for another day.)

src/rpc/rawtransaction_util.cpp
apoelstra: I did consciously change this to an `assert`, though I didn't think much about the code history. Let me change it back because I don't remember my reasoning.

gwillen:
utACK b2e1cc38de, verified that it contains only the requested changes from da11d7b6fd.

```
[nix-shell:~/src/blockstream/elements/elements]$ git range-diff 904054ffcb4a67905577a4eb8b256369111be387..da11d7b6fd  532d55d059181790eb3b39c734a20f643017b106..b2e1cc38de 
1:  8195a720cc ! 1:  313f73d5b2 When validation is waiting for parent chain daemon, "stall".
    @@ src/rpc/rawtransaction_util.cpp: bool ValidateTransactionPeginInputs(const CMuta
     -            CHECK_NONFATAL(err == "Needs more confirmations.");
     +        bool depth_failed = false;
     +        if(txin.m_is_pegin && !IsValidPeginWitness(mtx.witness.vtxinwit[i].m_pegin_witness, fedpegscripts, txin.prevout, err, true, &depth_failed)) {
    -+            assert(depth_failed);
    ++            CHECK_NONFATAL(depth_failed);
                  immature_pegin = true;
              }
          }
2:  f1764c1b14 ! 2:  d5042b41c8 Finish removing 'recheckpeginblockinterval'; move MainchainRPCCheck
    @@ src/init.cpp: bool AppInitMain(const util::Ref& context, NodeContext& node, inte
     -            // Or gently warn the user, and continue
     -            InitError(Untranslated(err_msg));
     -            gArgs.SoftSetArg("-validatepegin", "0");
    -+        if (!MainchainRPCCheck()) { //Initial check only
    ++        if (!MainchainRPCCheck()) {
     +            const std::string err_msg = "ERROR: elements is set to verify pegins but cannot get a valid response from the mainchain daemon. Please check debug.log for more information.\n\nIf you haven't setup a bitcoind please get the latest stable version from https://bitcoincore.org/en/download/ or if you do not need to validate pegins set in your elements configuration validatepegin=0";
     +            // We fail immediately if this node has RPC server enabled
     +            if (gArgs.GetBoolArg("-server", false)) {
3:  da11d7b6fd = 3:  b2e1cc38de Regression test for pegin validation issues during sync
```

Issue #1031, opened 2021-09-02
Weight needs to be cached for OP_TXWEIGHT


roconnor-blockstream:
The operation `GetSerializeSize` is `O(transaction_size)` instead of `O(1)`.  Thus this value need to be cached.

https://github.com/ElementsProject/elements/commit/61b5152d445508f12155b477d77fa0940ae7fba7#diff-a0337ffd7259e8c7c9a7786d6dbd420c80abfa1afdb34ebae3261109d9ae3c19R2426-R2431

apoelstra:
Closed by #1030

PR #1032, opened 2021-09-03
script: cache transaction weight per-transaction

apoelstra:
This is an O(N) operation so we want to make sure that we're not recomputing
it every time we hit the INSPECTTXWEIGHT opcode.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -2020,7 +2020,12 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript& 
                         }
                         case OP_TXWEIGHT:
                         {
-                            push8_le(stack, checker.GetTxWeight());
+                            const PrecomputedTransactionData *cache = checker.GetPrecomputedTransactionData();
+                            // Return error if the evaluation context is unavailable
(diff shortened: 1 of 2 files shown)


src/script/interpreter.cpp
```diff
@@ -2993,13 +2999,6 @@ const std::vector<CTxOut>* GenericTransactionSignatureChecker<T>::GetTxvOut() co
     return &(txTo->vout);
 }
 
-template <class T>
-uint64_t GenericTransactionSignatureChecker<T>::GetTxWeight() const
-{
-    // line copied from GetTransactionWeight() in src/consensus/validation.h
```
roconnor-blockstream, line 2999: This comment got left out.

apoelstra: thanks, i'll re-add it

roconnor-blockstream:
LGTM.

sanket1729 (approved):
CR ACK 4aad8c1a145a814b31aec39d57aa2056f2773e45

roconnor-blockstream:
Fixes https://github.com/ElementsProject/elements/issues/1031

PR #1033, opened 2021-09-06
Fix another instance of bitcoin #20347 nailing us

apoelstra:
This is fixed completely upstream in 22.0, but for now we need to patch around it.

Discovered while cleaning up the assets tutorial.


achow101:
ACK 2ec356a1851c72fb543c260ad84634e090d8d34b

PR #1036, opened 2021-09-07
Rewrite assets tutorial

apoelstra:
Clean up and rewrite the assets tutorial

* Updates RPC to handle 0.21 changes (you now need createwallet/loadwallet)
* Cleans up stray processes and temp directories even when it crashes
* Significantly expands comments
* Modernizes the Python, removes weird idioms like calling `raise` in an `except: pass` block

### DIFF
--- a/contrib/assets_tutorial/assets_tutorial.py
+++ b/contrib/assets_tutorial/assets_tutorial.py
@@ -56,7 +56,10 @@ class Daemon():
 
     def shutdown(self):
         if self.proc is not None:
-            self.proc.kill()
+            print ("Shutting down %s" % self.name)
+            self.proc.terminate()
+            ## FIXME determine why we need 30+ seconds to shut down with a tiny regtest chain
+            self.proc.wait(120)
             self.proc = None
 
         if self.datadir_path is not None:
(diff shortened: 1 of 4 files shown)


contrib/assets_tutorial/assets_tutorial.py
```diff
@@ -3,111 +3,185 @@
...
+
+def sync_all(nodes, timeout_sec = 10):
+    totalWait = timeout_sec
+
+    stop_time = time.time() + timeout
+    while time.time() <= stop_time:
+        best_hash = [x.getbestblockhash() for x in nodes]
+        if best_hash.count(best_hash[0]) == len(nodes):
+            return
+    raise Exception("Nodes cannot sync blocks or mempool!")
```
gwillen, line 125: You're no longer actually checking the mempool so this message became misleading.

gwillen: Nevermind, fixed in later commit.

gwillen:
utACK 1d2dacb.

I reviewed the first two commits relatively carefully, and the third one up to `## 2. Basic wallet usage`. After that I was skimming somewhat in the body of the tutorial itself.

PR #1038, opened 2021-09-09
PSET: Fix blinded coinjoins with 3 or more parties

### LINKED ISSUE #1037: PSET blinding implementation is missing pieces
* Blinding a PSET should not change its unique ID; the unique ID should be computed from the explicit version of the transaction always
* When a PSET is blinded, the asset and value commitments need to be accompanied by cryptographic proofs that they match the explicit value (for rangeproofs set `ct_bits` to -1 to make an explicit value proof, for surjection proofs just do a proof with an input set of size 1)
* Before signing a PSET, these proofs need to be verified

achow101:
In order for blinded coinjoins with 3 or more parties to work, some fields need to be added, amounts cannot be removed, and some bugs need fixing.

First and foremost is to no longer remove amounts after blinding. Due to a miscommunication, I had believed that part of the goal of PSET was to hide semi private information (such as output amounts) from other parties in the transaction. However this causes the combiner to fail because the unique ID is dependent on those amounts and their commitments. If multiple parties had blinded just their own outputs, then the resulting PSETs would not combine because the amounts had be removed and so the unique ID calculated was incorrect. In order for this combining to work, amounts must be kept after blinding and the unique ID calculation must only use the computed unblinded transaction. This change has also been made to the spec document.

Second is the addition of explicit value and asset proofs. In order to prove that the commitments commit to the given explicit value or asset, explicit value rangeproofs and explicit asset surjection proofs are added to PSET. Each blinded output must have an explicit value rangeproof and explict asset surjection proof after blinding. For issuances, explicit issuance value proofs and explicit reissuance value proofs must be added after blinding. These proofs are verified prior to signing.

Thirdly, in order to make the coinjoin workflow work, a `blind` option is added `walletprocesspsbt` so that users can tell `walletprocesspsbt` to not attempt to blind. This is important because blinding requires all UTXOs to be present, and UTXOs can only be added via a call to `walletprocesspsbt`. Error messages have been added and improved so that users who do things in the wrong order will be less likely to end up with an unusable PSET.

Lastly, a test case has been added for a 3 party coinjoin workflow.

Fixes #1037

### DIFF
--- a/src/blindpsbt.cpp
+++ b/src/blindpsbt.cpp
@@ -64,6 +64,21 @@ bool CreateAssetSurjectionProof(std::vector<unsigned char>& output_proof, const 
     return true;
 }
 
+bool VerifyBlindAssetProof(const std::vector<unsigned char>& proof, const CConfidentialAsset& conf_asset)
+{
+    secp256k1_surjectionproof surj_proof;
+    if (secp256k1_surjectionproof_parse(secp256k1_blind_context, &surj_proof, proof.data(), proof.size()) == 0) {
+        return false;
+    }
+
+    secp256k1_generator gen;
+    if (secp256k1_generator_parse(secp256k1_blind_context, &gen, conf_asset.vchCommitment.data()) == 0) {
+        return false;
+    }
+
+    return secp256k1_surjectionproof_verify(secp256k1_blind_context, &surj_proof, &gen, 1, &gen) == 0;
+}
+
 uint256 GenerateRangeproofECDHKey(CPubKey& ephemeral_pubkey, const CPubKey blinding_pubkey)
 {
     // Generate ephemeral key for ECDH nonce generation
@@ -113,6 +128,26 @@ bool CreateBlindValueProof(std::vector<unsigned char>& rangeproof, const uint256
     int res = secp256k1_rangeproof_sign(secp256k1_blind_context, rangeproof.data(), &rangeproof_len, /* min_value */ amount, &value_commit, value_blinder.begin(), nonce.begin(), /* exp */ -1, /* min_bits */ 0, amount, /* message */ nullptr, /* message_len */ 0, /* extra_commit */ nullptr, /* extra_commit_len */ 0, &gen);
     rangeproof.resize(rangeproof_len);
     return res == 1;
+}
+
+bool VerifyBlindValueProof(CAmount value, const CConfidentialValue& conf_value, const std::vector<unsigned char>& proof, const CConfidentialAsset& conf_asset)
+{
+    secp256k1_pedersen_commitment value_commit;
+    if (secp256k1_pedersen_commitment_parse(secp256k1_blind_context, &value_commit, conf_value.vchCommitment.data()) == 0) {
+        return false;
+    }
+
+    secp256k1_generator gen;
+    if (secp256k1_generator_parse(secp256k1_blind_context, &gen, conf_asset.vchCommitment.data()) == 0) {
+        return false;
+    }
+
+    uint64_t min_value;
+    uint64_t max_value;
+    if (secp256k1_rangeproof_verify(secp256k1_blind_context, &min_value, &max_value, &value_commit, proof.data(), proof.size(), /* extra_commit */ nullptr, /* extra_commit_len */ 0, &gen) == 0) {
+        return false;
+    }
+    return min_value == (uint64_t)value;
 }
 
 void CreateAssetCommitment(CConfidentialAsset& conf_asset, secp256k1_generator& asset_gen, const CAsset& asset, const uint256& asset_blinder)
--- a/src/blindpsbt.cpp
+++ b/src/blindpsbt.cpp
@@ -34,11 +34,11 @@ std::string GetBlindingStatusError(const BlindingStatus& status)
 }
 
 // Create surjection proof
-bool CreateAssetSurjectionProof(std::vector<unsigned char>& output_proof, const std::vector<secp256k1_fixed_asset_tag>& fixed_input_tags, const std::vector<secp256k1_generator>& ephemeral_input_tags, const std::vector<uint256>& input_asset_blinders, const uint256& output_asset_blinder, const secp256k1_generator& output_asset_tag, const CAsset& asset)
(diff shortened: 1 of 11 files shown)


src/blindpsbt.cpp
```diff
@@ -98,6 +98,23 @@ bool CreateValueRangeProof(std::vector<unsigned char>& rangeproof, const uint256
     return (res == 1);
 }
 
+// Create an explicit value rangeproof which proves that the commitment commits to an explicit value
+bool CreateBlindValueProof(std::vector<unsigned char>& rangeproof, const uint256& value_blinder, const CAmount amount, const secp256k1_pedersen_commitment& value_commit, const secp256k1_generator& gen)
+{
+    // Prep rangeproof
+    size_t rangeproof_len = 5134;
```
apoelstra, line 122: This should be reduced (I will find the right value and revisit this comment).

Similarly it would be good to assert the value of this variable after the proof is created.

src/rpc/rawtransaction.cpp
```diff
@@ -1649,6 +1663,16 @@ static RPCHelpMan decodepsbt()
             out.pushKV("blinder_index", (int64_t)*output.m_blinder_index);
         }
 
+        // Blind value proof
+        if (!output.m_blind_value_proof.empty()) {
+            out.pushKV("blind_value_proof", HexStr(output.m_blind_value_proof));
+        }
+
+        // Blind asset proof
+        if (!output.m_blind_value_proof.empty()) {
```
apoelstra, line 1672: copy/paste error: `m_blind_value_proof` should be `m_blind_asset_proof`

src/blindpsbt.cpp
```diff
@@ -115,6 +130,26 @@ bool CreateBlindValueProof(std::vector<unsigned char>& rangeproof, const uint256
...
+    if (secp256k1_generator_parse(secp256k1_blind_context, &gen, conf_asset.vchCommitment.data()) == 0) {
+        return false;
+    }
+
+    uint64_t min_value;
+    uint64_t max_value;
+    if (secp256k1_rangeproof_verify(secp256k1_blind_context, &min_value, &max_value, &value_commit, proof.data(), proof.size(), /* extra_commit */ nullptr, /* extra_commit_len */ 0, &gen) == 0) {
+        return false;
+    }
+    return min_value == (uint64_t)value;
```
apoelstra, line 152: Let me double-check on this but I believe you also need to check that `min_value == max_value`.

apoelstra:
This looks great! I especially like the new argument to `GetUnsignedTx` which turned out to be a very low-diff way to get the two versions of the transaction.

Aside from the nits, my only concern is about the `blind=false` option to `walletprocesspsbt`. From private conversation with you, I understand that the purpose of this is so that we don't create asset surjection proofs before all the input data has been provided (if we did it using only our own inputs, this would reduce privacy). But wouldn't it work just as well for `walletprocesspsbt` to analyze the transaction and only blind if it can find asset IDs for every input? Then the onus is no longer on the user to provide this option.

As a followup, it would probably be useful to add another field to the `walletprocesspsbt` output, aside from just `complete`, which indicates the current state of the PSET (missing input data, not yet blinded, will not be blinded, blinded, signed/complete). It would be fine to do this in a separate PR or even as part of a later Elements release.

src/rpc/rawtransaction.cpp
achow101: Fixed

achow101:
> Aside from the nits, my only concern is about the `blind=false` option to `walletprocesspsbt`. From private conversation with you, I understand that the purpose of this is so that we don't create asset surjection proofs before all the input data has been provided (if we did it using only our own inputs, this would reduce privacy). But wouldn't it work just as well for `walletprocesspsbt` to analyze the transaction and only blind if it can find asset IDs for every input? Then the onus is no longer on the user to provide this option.

It was simpler than I had anticipated to be "smart" about this, so I have removed `blind` and implemented the "smart" method.

> As a followup, it would probably be useful to add another field to the `walletprocesspsbt` output, aside from just `complete`, which indicates the current state of the PSET (missing input data, not yet blinded, will not be blinded, blinded, signed/complete). It would be fine to do this in a separate PR or even as part of a later Elements release.

This is what `analyzepsbt` is for (I don't remember if it works currently).

apoelstra:
>You areIt was simpler than I had anticipated to be "smart" about this, so I have removed blind and implemented the "smart" method.

lol, on a roll with these small-diff solutions.

code review ACK. Let me do some more testing and I will merge this.

test/functional/rpc_psbt.py
```diff
@@ -637,6 +638,65 @@ def run_ca_tests(self):
...
+            ]
+        )
+        # Update all but don't blind
+        up_psbt1 = w1.walletprocesspsbt(psbt=created_psbt, sign=False)["psbt"]
+        up_psbt2 = w2.walletprocesspsbt(psbt=created_psbt, sign=False)["psbt"]
+        up_psbt3 = w3.walletprocesspsbt(psbt=created_psbt, sign=False)["psbt"]
+        # Combine updated
+        comb_psbt1 = self.nodes[0].combinepsbt([up_psbt1, up_psbt2, up_psbt3])
+        # 1 and 2 blind
+        blind_psbt1 = w1.walletprocesspsbt(psbt=comb_psbt1, sign=False)["psbt"]
```
apoelstra, line 695: If you double-up this line, e.g. by writing
```
         blind_psbt1 = w1.walletprocesspsbt(psbt=blind_psbt1, sign=False)["psbt"]
```
after it, you will get a failure of the form `Transaction has blind inputs belonging to this blinder but does not have outputs to blind`.

The reason is that at the beginning of `BlindPSBT` in blindpsbt.cpp, we check "is everything done" by comparing the number of fully blinded outputs with the number of outputs -- but we do this **before** filtering out outputs that aren't ours. Then when we do the actual blinding, skipping all the outputs that (a) weren't ours, and (b) were already blinded, so we skip all of them. Finally at the bottom of the function we check whether we had blinded inputs (yes) and whether we blinded no outputs (also yes) and then throw this error.

Can you add this extra test line, perhaps with a comment explaining its purpose (it should be a no-op and not throw any errors), and change the logic in blindpsbt.cpp?

achow101: Fixed, with test.

apoelstra:
I have a couple more ntis but I'll address them in a followup PR.

PR #1039, opened 2021-09-09
claimpegin: sanity-check fee in case fallbackfee is disabled, throw explicit exception

apoelstra:
Produce a much less confusing error message in `claimpegin` in the case that estimatesmartfee fails and no fallbackfee is set.


stevenroose:
ACK fc8bfc1f2cda0844adbf1d831a1a5e1888f8949b

PR #1040, opened 2021-09-10
regtest: change network magic away from liquid v1

apoelstra:
We don't want connectivity issues between Liquid and custom test networks.


stevenroose:
ACK e57b4c51ad2ee43897d2fcac4e7cd6b8858c674f

apoelstra:
This change is actually pretty invasive.

* On existing regtest networks, or 0.18 networks where you try to test upgrading to 0.21, you will get silent inability to connect
* Worse(?) when starting up, you get assertion errors loading blocks from disk as the network magic appears in the disk serialization

I think we should probably revert this, even though it is regtest-only.

cc @psgreco

apoelstra:
Or maybe we should put output a warning on startup for a couple of versions.

apoelstra:
Also, if we keep this change, we should've done it to` LiquidV1Test` as well as regtest/custom.

PR #1041, opened 2021-09-10
re-enable fallbackfee by default

apoelstra:
Upstream disabled the default `fallbackfee` option in 0.17, which caused us some user error reports in 0.18. Re-enable the option in 0.21.


stevenroose:
Should it be 1sat/vb or 0.1 sat/vb. Currently the minrelayfee is 0.1 sat/vb and that's kinda what we advertise as Liquid's feerate, isn't it?

apoelstra:
Oops, yeah, I had 1, corrected it, and forgot to fix the comment.

stevenroose:
utACK 1204b91c2d5c36c12000fca1414726f19d2e7354

PR #1043, opened 2021-09-14
fuzz: update psbt fuzztest for PSBT2 changes

apoelstra:
The `psbt.tx` Optional is no longer safe to dereference just because decoding succeeded.

### DIFF
--- a/src/test/fuzz/psbt.cpp
+++ b/src/test/fuzz/psbt.cpp
@@ -42,11 +42,8 @@ void test_one_input(const std::vector<uint8_t>& buffer)
 
     (void)psbt.IsNull();
 
-    Optional<CMutableTransaction> tx = psbt.tx;
-    if (tx) {
-        const CMutableTransaction& mtx = *tx;
-        const PartiallySignedTransaction psbt_from_tx{mtx};
-    }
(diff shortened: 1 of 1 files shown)


src/test/fuzz/psbt.cpp
```diff
@@ -57,7 +54,7 @@ void test_one_input(const std::vector<uint8_t>& buffer)
         (void)output.IsNull();
     }
 
-    for (size_t i = 0; i < psbt.tx->vin.size(); ++i) {
+    for (size_t i = 0; i < mtx.vin.size(); ++i) {
         CTxOut tx_out;
         if (psbt.inputs.at(i).GetUTXO(tx_out)) {
```
achow101, line 59: It's simpler to just iterate over the `PSBTInput`s:

```suggestion
    for (const auto& input : psbt.inputs) {
        CTxOut tx_out;
        if (input.GetUTXO(tx_out)) {
```

apoelstra: Thanks, fixed

achow101:
ACK bcd7de55a40b8f9e4e2f1993a602b868e56d06b8

PR #1044, opened 2021-09-15
Define Taproot activation parameters for Liquid

apoelstra:
Sets Taproot to start signalling around noon (California time) on Nov 1, 2021, assuming 95% of blocks are produced between now and then.

Will activate after one week of 100% signalling. If we can pull this off on the first or second try, we will beat Bitcoin which currently looks like it will activate on Nov 16.

**Edit:** actually, even without Speedy Trial, there is one full period (week) where Taproot will be "locked in" but not "active". So it will activate on Nov 15 at the earliest.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -478,8 +478,10 @@ public:
         consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime = 1199145601; // January 1, 2008
         consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nTimeout = 1230767999; // December 31, 2008
         consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].bit = 2;
-        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = Consensus::BIP9Deployment::ALWAYS_ACTIVE;
+        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = gArgs.GetArg("-con_taproot_signal_start", Consensus::BIP9Deployment::ALWAYS_ACTIVE);
         consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nTimeout = Consensus::BIP9Deployment::NO_TIMEOUT;
+        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nPeriod = 128; // test ability to change from default
+        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nThreshold = 128;
 
         consensus.nMinimumChainWork = uint256{};
         consensus.defaultAssumeValid = uint256{};
@@ -1249,6 +1251,10 @@ public:
             consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].bit = 25;
             consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime = args.GetArg("-con_dyna_deploy_start", Consensus::BIP9Deployment::ALWAYS_ACTIVE);
             consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nTimeout = Consensus::BIP9Deployment::NO_TIMEOUT;
+        }
+
+        if (args.IsArgSet("-con_taproot_signal_start")) {
+            consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = gArgs.GetArg("-con_taproot_signal_start", 0);
         }
 
         // END ELEMENTS fields
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -978,6 +978,12 @@ public:
         consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY].nStartTime = 0;
         consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY].nTimeout = Consensus::BIP9Deployment::NO_TIMEOUT;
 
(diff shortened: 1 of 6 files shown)


gwillen:
I don't feel like I understand the activation machinery well enough to truly ack this without a bunch more thought, but it seems reasonable to me.

stevenroose:
utACK 5291c0d9d9c09b3613d2fe3fb73311edab7491f3
checked the values, was confused for a while because the nStartTime is a block height instead of a timestamp :D

src/chainparams.cpp
```diff
@@ -978,6 +980,12 @@ class CLiquidV1Params : public CChainParams {
         consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY].nStartTime = 0;
         consensus.vDeployments[Consensus::DEPLOYMENT_TESTDUMMY].nTimeout = Consensus::BIP9Deployment::NO_TIMEOUT;
 
+        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].bit = 2;
+        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = 1554500; // November 1, 2021
```
ajtowns, line 984: Isn't the current code (prior to this patch) buggy in that it leaves `consensus.vDeployments[TAPROOT]` uninitialised, so that it could be anything?

apoelstra: Yep, oops. This has only been true on `master` and for a day or two (during which time we've been merging a lot of taproot-related PRs), but it shouldn't have happened nonetheless.

apoelstra: BTW -- to clarify your comment on twitter, this won't cause Taproot to be enforced on the LIquid network (except in the sense that leaving these variables undefined is UB so theoretically anything could happen). It would, at worst, cause the functionaries to start signalling early. But even that won't happen, because ofc the functionaries are not running `master`.

psgreco:
Looks good, utACK 0df8895b0ba53871fde3d42a766442b58439e389

src/chainparams.cpp
```diff
@@ -478,8 +478,10 @@ class CRegTestParams : public CChainParams {
         consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime = 1199145601; // January 1, 2008
         consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nTimeout = 1230767999; // December 31, 2008
         consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].bit = 2;
-        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = Consensus::BIP9Deployment::ALWAYS_ACTIVE;
+        consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = gArgs.GetArg("-con_taproot_signal_start", Consensus::BIP9Deployment::ALWAYS_ACTIVE);
```
achow101, line 481: In 5291c0d9d9c09b3613d2fe3fb73311edab7491f3 "chainparams: add undocumented regtest/testnet only -con_taproot_signal_start option"

This is entirely unnecessary as there already exists a way to set regtest activation params via `-vbparams`. See the function `CRegTestParams::UpdateActivationParametersFromArgs`.

test/functional/feature_elements_taproot_activation.py
```diff
@@ -0,0 +1,121 @@
...
+        self.nodes[node_idx].submitblock(block)
+        assert_equal(self.nodes[node_idx].getblockchaininfo()["softforks"]["taproot"]["bip9"]["status"], "started")
+
+        # Run through another 128 blocks, without failing to signal
+        blocks = self.nodes[node_idx].generatetoaddress(127, self.nodes[node_idx].getnewaddress())
+        for n, block in enumerate(blocks):
+            decode = self.nodes[node_idx].getblockheader(block)
+            assert_equal (decode["versionHex"], "20000004")
+        assert_equal(self.nodes[node_idx].getblockchaininfo()["softforks"]["taproot"]["bip9"]["status"], "started")
+        # The 128th block then switches from "locked_in" to "active"
```
achow101, line 79: In 0df8895b0ba53871fde3d42a766442b58439e389 "test: add test for Taproot activation"

```suggestion
        # The 128th block then switches from "started" to "locked_in"
```

test/functional/feature_elements_taproot_activation.py
```diff
@@ -0,0 +1,121 @@
...
+
+class TaprootActivationTest(BitcoinTestFramework):
+    def set_test_params(self):
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+
+    def skip_test_if_missing_module(self):
+        self.skip_if_no_wallet()
+
+    def test_activation(self, node_idx, activation_height):
```
achow101, line 32: In 0df8895b0ba53871fde3d42a766442b58439e389 "test: add test for Taproot activation"

It would be easier to read this test if instead of passing in the node index, the rpc object to use was passed in directly, e.g.:

```suggestion
    def test_activation(self, rpc, activation_height):
```

and the caller would be

```
        self.test_activation(self.nodes[0], 500)
```

achow101 (commented):
ACK 0df8895b0ba53871fde3d42a766442b58439e389

The parameters themselves don't make a whole lot of sense to me, but the code itself is mostly fine.

achow101:
ACK 3cb9612faa21e9bc6eeaf398489f0aa80dd95a15

src/chainparams.cpp
apoelstra: Thank you! I didn't know about this.

I'm going to leave it in because this code is inherently temporary and we're already using it for internal testing, but in future we can avoid these kind of hacky undocumented options.

stevenroose: Eh, also didn't know that. We did the same thing for dynafed :)

PR #1051, opened 2021-10-01
combinepsbt: do safety checks that explicit values are only used if they are proven

apoelstra:
Also update the PSET tutorial


achow101:
ACK 187c8094bb662599bf946d4ad27b22559581512b

achow101:
ACK fec0b59112a57038499f5ee060bdc37cbdd5c676

PR #1052, opened 2021-10-01
chainparams: change magic bytes for liquidv1test

apoelstra:
This is basically a Liquid-prod-ish version of regtest. Its default magic should not collide with that for Liquid prod.


psgreco:
UTACK 0628311b532706262d8612de9fc5c1644035053d

PR #1053, opened 2021-10-02
rc2 backports

apoelstra:
.


src/blindpsbt.cpp
```diff
@@ -311,8 +311,17 @@ bool ComputeAndAddToScalarOffset(uint256& a, CAmount value, const uint256& asset
     if (a.IsNull()) {
         a = scalar;
     } else {
-        // If we have a, then add the scalar to it.
-        if (secp256k1_ec_privkey_tweak_add(secp256k1_blind_context, a.begin(), scalar.begin()) != 1) return false;
+        uint256 scalar_negated = scalar;
+        if (secp256k1_ec_seckey_negate(secp256k1_blind_context, scalar_negated.begin()) != 1) {
+            return false;
+        }
+        // Special-case zero, which would otherwise cause `secp256k1_ec_privkey_tweak_add` to fail
```
gwillen, line 318: I don't really want to study this in the depth that would be required for me to check it, so: promise me you've thought really hard about why it's okay for this to be zero? (I assume it's not a privkey and can never be accidentally used as a privkey, despite the name of the function being used to do math on it.)

src/psbt.cpp
```diff
@@ -138,6 +138,7 @@ CMutableTransaction PartiallySignedTransaction::GetUnsignedTx(bool force_unblind
         txout.scriptPubKey = *output.script;
 
         bool exp_value = output.m_value_commitment.IsNull() || force_unblinded;
+        exp_value = exp_value && output.amount != nullopt;
```
gwillen, line 141: Just an observation, not sure if there's anything to change, but: It feels like the CTxOutput interface is more intuitive here. (Where you have a field that is a three-way union of null/explicit/commitment.) And less prone to this kind of omission.

src/blindpsbt.cpp
apoelstra: Yes.

src/psbt.cpp
apoelstra: I do not thing that a three-way tagged union implemneted as a byte vector is more intuitive than boolean logic on booleans.

gwillen:
(Andrew has pointed out to me that I was reading the commit list wrong, and all these commits are part of MRs that were already merged into master. So any issues should be dealt with in future MRs, and not fiddling around with what's here.)

apoelstra:
Reconstructed this PR after merging #1054 and #1055

psgreco:
UTACK 84a92533442ad07cb059457b344ab6b5f4a798b5

PR #1055, opened 2021-10-06
minor PSET fix

apoelstra:
Noticed because it triggered a compiler warning.

No test -- we will add one in a future "end-to-end issuance via PSET" tutorial script which is out of scope for the 0.21 release.


achow101:
ACK 0e1007fa7b86de50ffcfa30f98e09ebc06ce2a72

Issue #1061, opened 2021-11-03
Implement vbparams instead of using custom flags for elementsregtest


psgreco:
Currently elements uses custom config flags like `con_dyna_deploy_start` or `con_dyna_taproot_start` to signal things, when this could be done the "bitcoin" way using something like vbparams (maybe evbparams)

PR #1062, opened 2021-11-03
Implement vbparams equivalent for elements

psgreco:
Fixes: https://github.com/ElementsProject/elements/issues/1061

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -86,6 +86,89 @@ static CBlock CreateGenesisBlock(uint32_t nTime, uint32_t nNonce, uint32_t nBits
 }
 
 /**
+ * Allows modifying the Version Bits Elements regtest parameters (and liquidv1test).
+ * Ideally, this would be a method in the base class, inherited everywhere, but that might complicate future merges,
+ * so we settle for this static function.
+ */
+static void UpdateElementsActivationParametersFromArgs(Consensus::Params& consensus, const ArgsManager& args)
+{
+    if (!args.IsArgSet("-evbparams")) return;
+
+    for (const std::string& strDeployment : args.GetArgs("-evbparams")) {
+        std::vector<std::string> vDeploymentParams;
+        boost::split(vDeploymentParams, strDeployment, boost::is_any_of(":"));
+        if (vDeploymentParams.size() != 5) {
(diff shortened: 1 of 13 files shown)


src/chainparams.cpp
```diff
@@ -1275,9 +1360,7 @@ class CLiquidV1TestParams : public CLiquidV1Params {
             consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nTimeout = Consensus::BIP9Deployment::NO_TIMEOUT;
         }
 
-        if (args.IsArgSet("-con_taproot_signal_start")) {
-            consensus.vDeployments[Consensus::DEPLOYMENT_TAPROOT].nStartTime = gArgs.GetArg("-con_taproot_signal_start", 0);
-        }
+        UpdateElementsActivationParametersFromArgs(consensus, args);
```
stevenroose, line 1368: Why didn't you use the same method name as upstream? It's a trade-off between not getting conflicts when they make trivial changes to that method and not getting conflicts when they make critical changes to that method :sweat_smile:

stevenroose:
Looks good, utACK 85c3c2b

src/chainparams.cpp
psgreco: Because even though it is the same idea, it does more things than the method upstream, it uses more arguments (Period and Threshold) and makes them optional. Also in upstream it is a method of CRegTestParams and here we also need to call it from liquidv1test

src/validation.cpp
```diff
@@ -1982,7 +1982,7 @@ int32_t ComputeBlockVersion(const CBlockIndex* pindexPrev, const Consensus::Para
     }
 
     // Undo default signalling behavior for dynafed unless explicitly enabled.
```
stevenroose, line 1984: This comment should go

stevenroose (changes requested):
One comment that needs to go

src/validation.cpp
psgreco: Agreed

stevenroose:
utACK da6ebdf

stevenroose:
Accidentally merged older version missing one change, @psgreco is going to PR the change separately.

PR #1066, opened 2021-11-11
Reevaluate descriptor post parse

jhfrontz:
In `sendtomainchain_pak()` a check is made if the associated `descriptor` is null and an attempt to remedy is made -- but the remedy is never applied to `descriptor` (which results in a SEGV later).  This code fixes that.

Note: unfortunately it also removes the `const` qualifier.  I'm not sure what/if there are conventions for dealing with that (e.g., creating a new `const` of, like, `sanitized_descriptor` for subsequent use in the function?).


stevenroose:
utACK 6c1d0d3ca80fe1a9d14f2559d11ba9e89f45fd72

PR #1067, opened 2021-11-11
[elements-0.21] Re-evaluate peg-out descriptor if wallet is reparsed

jhfrontz:
v0.21.0 version of #1066:

In sendtomainchain_pak() a check is made if the associated descriptor is null and an attempt to remedy is made -- but the remedy is never applied to descriptor (which results in a SEGV later). This code fixes that.

Note: unfortunately it also removes the const qualifier. I'm not sure what/if there are conventions for dealing with that (e.g., creating a new const of, like, sanitized_descriptor for subsequent use in the function?).


stevenroose:
Apparently we're going to merge master into the 0.21 branch, so this backport is not needed.

Issue #1013, opened 2021-06-21
Windows installation crashes while downloading blockchain - out of memory error - elements-0.18.1.12


blankscreengithub:
Program crashes during initial download of the blockchain. 

I expected it to download the block chain.

I can only download about 80MB to 100MB and then the program crashes. If I restart the application, it restarts the block download from the initial block.

This happens with all versions of the windows applications.

v0.21.1

Intel(R) Core(TM) i7-5500U CPU @ 2.40GHz   2.39 GHz
Ram 12.0 GB
256 GB SSD Samsung

Not a GUI related issue
Windows 10 Home
21H1
OS Build 19043.1055

From debug.log:
2021-06-20T23:38:20Z UpdateTip: new best=fca95bc6c01481162249bbd09508d28299bc9bacc8a586fef4297522cb6ddf5f height=50191 version=0x20000000 tx=50347 date='2018-11-09T23:00:59Z' progress=0.036000 cache=0.0MiB(212txo)
2021-06-20T23:38:20Z UpdateTip: new best=7041420d2caecaf40bc09b05de2959453ac1ca9ff9cdb57f7a95188aa988bf0a height=50192 version=0x20000000 tx=50349 date='2018-11-09T23:02:59Z' progress=0.036000 cache=0.0MiB(213txo)
2021-06-20T23:38:20Z UpdateTip: new best=49f6d3927b45b6f29532b5c3b41ce77b50c7bd56ae30b832abfd219829356040 height=50193 version=0x20000000 tx=50350 date='2018-11-09T23:03:59Z' progress=0.036000 cache=0.0MiB(213txo)
2021-06-20T23:38:27Z Error: Out of memory. Terminating.
2021-06-21T01:14:13Z

blankscreengithub:
Update:
I was able to download 1GB of the blockchain before the program crashed. I started with 6GB of free RAM and I still received the out of memory error. How much memory on Windows does this program need to do the initial blockchain download?

blankscreengithub:
Update: I downloaded about 8 GB before it crashed, but my RAM memory still had some available. Technically, I did not run out of memory. Maybe I need to make a larger swap file size?

BitcoinPizza:
same here on my raspbian buster 10. Near block 30k it eat all my 2gi of free mem and die. starting it using -dbcache=4 does not help at all.

PR #1069, opened 2021-11-17
Fix elements multiple-header-download issue.

gwillen:
This fixes an issue which causes Elements to download the blockchain headers
multiple times during initial block download.

In particular: each time we receive an INV P2P message with a new block
(about once a minute), we start downloading the headers, again, in parallel
with any existing download(s) in progress.

With this change, after we receive each batch of headers, we check whether
any of the headers in it were new to us. If not (they were all duplicates),
we stop there, and do not ask the peer for another batch. This reduces the
maximum amount of duplication to about 2x, which is not ideal, but a HUGE
improvement.

With luck, this may help with #1013. But it's primarily directed at fixing slow initial block download, which I can't immediately find an open issue for on this repo, but has been a long-standing problem.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -1843,7 +1843,8 @@ void PeerManager::ProcessHeadersMessage(CNode& pfrom, const std::vector<CBlockHe
     }
 
     BlockValidationState state;
-    if (!m_chainman.ProcessNewBlockHeaders(headers, state, m_chainparams, &pindexLast)) {
+    bool all_duplicate = false;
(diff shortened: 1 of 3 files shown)


gwillen:
A note of context: I have not been able to figure out how our behavior on this differs from upstream Bitcoin Core. It is not out of the question that Core might also have this issue, but because Core's headers are much smaller than ours, and blocks much less frequent, it may not actually trigger for them, or else it may result in a totally insignificant amount of extra work and have never been noticed.

stevenroose (approved):
looks good, utACK cdfb4c9c6cd8909efeb59c8c47cdcef7c1822b26

stevenroose:
Could you backport this to 0.18.

PR #1072, opened 2021-11-22
Remove redundant log message in IsValidPeginWitness

gwillen:
The value it logs, very frequently, is essentially static, coming straight from the chainparams, and should never change during runtime anyway. (With a small exception that doesn't really matter and is probably not the reason for the log.)


stevenroose:
utACK b2dd758ce4b9fbff88205c49ef6392862b15d6d1

PR #1094, opened 2022-03-01
Raise max P2P message size to account for larger elements dynafed header size.

gwillen:

### DIFF
--- a/src/net.h
+++ b/src/net.h
@@ -55,7 +55,7 @@ static const int FEELER_INTERVAL = 120;
 /** The maximum number of addresses from our addrman to return in response to a getaddr message. */
 static constexpr size_t MAX_ADDR_TO_SEND = 1000;
 /** Maximum length of incoming protocol messages (no message over 4 MB is currently acceptable). */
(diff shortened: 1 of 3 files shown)


gwillen:
Fixes an issue we're starting to see with dynafed headers causing GETHEADERS responses to exceed the maximum P2P message size. 

This is happening because we inherited a constant from bitcoin, which hardcodes the assumption that 2000 headers will fit in a maximum-sized P2P packet and can't easily be changed:

```
/** Number of headers sent in one getheaders result. We rely on the assumption that if a peer sends
 *  less than this number, we reached its tip. Changing this value is a protocol upgrade. */
static const unsigned int MAX_HEADERS_RESULTS = 2000;
```

gwillen:
Updated to also fix some hardcoded copies of the same constant in the python tests.

philippem:
tACK

psgreco:
ACK bf7f050aeb7c4f863af8ed35c0ee11f3c85991a5

stevenroose:
Our headers are way more than 4x the size of bitcoin headers, though. It might make sense to also somehow influence how much headers the p2p logic will request in a single getdata.

But ACK on this part.

PR #1098, opened 2022-03-03
[BACKPORT 0.18] Fix elements multiple-header-download issue.

gwillen:
This fixes an issue which causes Elements to download the blockchain headers
multiple times during initial block download.

In particular: each time we receive an INV P2P message with a new block
(about once a minute), we start downloading the headers, again, in parallel
with any existing download(s) in progress.

With this change, after we receive each batch of headers, we check whether
any of the headers in it were new to us. If not (they were all duplicates),
we stop there, and do not ask the peer for another batch. This reduces the
maximum amount of duplication to about 2x, which is not ideal, but a HUGE
improvement.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -1423,7 +1423,8 @@ bool static ProcessHeadersMessage(CNode *pfrom, CConnman *connman, const std::ve
 
     CValidationState state;
     CBlockHeader first_invalid_header;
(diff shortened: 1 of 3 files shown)


gwillen:
This was a slightly complex cherry-pick, and I'm having trouble getting the tests to run on my system. Hopefully CI will pass.

psgreco:
Side-by-side with the original fix makes sense, and local build works as expected. Also tests run fine here.
ACK f5d33d8676bd8dffeb93da9f35cc53ba114310cd

PR #1104, opened 2022-03-25
Docker containers

AaronDewes:
This adds an automatically build Docker container using GitHub actions and the GitHub container registry.

This will make it easier to use elements in node packages like Citadel or Umbrel.

Inspired by the lncm docker bitcoin container: https://github.com/lncm/docker-bitcoind


AaronDewes:
The docker container now has been tested on Citadel and seems to work successfully, I am syncing the chain with it currently.

I'll now also add a workflow to make sure the container is also built when a new tag is created.

wtogami:
Hi Aaron,
I appreciate you put effort into this. Similarly to Bitcoin Core, Elements is very cautious when it comes to official distributed binaries. Both projects distribute only reproducible binaries to reduce the risk of supply chain attack. Both projects would not distribute official docker containers increasing the attack surface.

For these reasons I don't expect this PR to ever be accepted into this repo. Instead we strongly recommend that distributors be responsible for their own builds. It would be best to maintain your docker build definitions elsewhere.

wtogami:
Aaron informed me that he has reproducible docker containers working for other components. That would be interesting to see here.

wtogami:
If I understand it correctly, Debian these days is reproducible but not clean bootstrap verified?

philippem:
Hello @AaronDewes 
Blockstream publishes docker images that package the official gitian-built Elements binaries.
https://hub.docker.com/r/blockstream/elementsd

The image is very simple, you will need to specify an entrypoint and volumes for conf and data.

Issue #259, opened 2017-09-13
reissuance tokens can not be exercised from unblinded outputs


instagibbs:
In the current logic, `CAssetIssuance`'s `assetBlindingNonce` being null implies that the issuance is an initial issuance. When re-issuing, this field is re-used to expose the asset blinder for the spent output.

Clearly this is in direct conflict when the output itself has a null blinder.

maaku:
I'm going to think on if there is a clever way of handling this.

nkostoulas:
Hello, I too came across this issue when trying to reissue from an unblinded issuance, leading to validation (VerifyAmounts()) failing and TX getting rejected.

Do you think there is any better way of doing this other than maybe having an additional flag in CAssetIssuance to differentiate between issuance and reissuance?

instagibbs:
@nkostoulas No great answers on this yet. Note that the outputs during an issuance should generally be blinded, it's just the issuance itself that will be unblinded(input-side). So ideally in an "all spendable outputs are blinded" world this won't even occur.

This is why I'm a little hesitant to change serializations just for this. Thoughts?

nkostoulas:
@instagibbs I guess I've introduced this by allowing unblinded outputs through configuration. In the standard config though where this is not the case, how does this become an issue given that the issuance token output should always be blinded and `assetBlindingNonce ` (when reissuing) non zero?

instagibbs:
You could send the reissuance token to an unblinded output if you strip the blinding pubkey from a destination address using `validateaddress`.

nkostoulas:
Ah okay now I understand! I'll try and look into it, but I'll be keen to fix this if there's no alternative.

instagibbs:
Right, this is pretty unfortunate in the completely unblinded chain case :-1: 

Definitely will get fixed if/when serialization changes are on the table.

stevenroose:
So once upside of this issue is that reissuance tokens amounts don't really matter at all. The only thing that matters is whether they are 0 or non-0. So even in a situation where one wants a chain with total transparency (Elements isn't the greatest option for that, though), blinded token amounts wouldn't change that much.

altafan:
Hello, I just bumped into this without knowing it was a known issue already.

Since I made some attempts using the CLI, that of course failed, maybe it could be useful having a script for anyone interested in reproducing this paritcular case.

Check out [this gist](https://gist.github.com/altafan/db02db3d0902389dd2cce35a648e6f17).

[...]

Issue #1103, opened 2022-03-18
`bad-txns-in-ne-out, value in != value out` Combine unblinded issuance with an unblinded re-issuance does not work  


tiero:
I am trying to make in the same transaction 

- an unblinded issuance of a new asset (amount 1, precision 0 and NO reissuance tokens)
- a unblinded reissuance of a previously issued asset (amount whatever) 

In this transaction the re-issuance token input and change output are blinded (and eventually the L-BTC input and change output)

### Bash script to replicate 

https://github.com/vulpemventures/liquidjs-lib/blob/issuance-mixed-reissuance/issuance-mixed-reissuance.sh

### Decoded raw transaction

```json
{
  "txid": "bd5ba95822b15979567655c4b9d1e5d886868548d19a2c22698f7af44afea069",
  "hash": "027ca81ab73f51f65284337cad90f5bbae554c59f693f4c19ef814a78e7ee459",
  "wtxid": "027ca81ab73f51f65284337cad90f5bbae554c59f693f4c19ef814a78e7ee459",
  "withash": "0702e167c7468291ff37b0f965669deea3244a574e10ab8d7e302e27265fcb3f",
  "version": 2,
  "size": 17957,
  "vsize": 5071,
  "weight": 20282,
  "locktime": 0,
  "vin": [
    {
      "txid": "77241409eaa7ef6663bd9ea9df6a2834e288c0d0368591304e87db365643b7f1",
      "vout": 1,
      "scriptSig": {
        "asm": "",
        "hex": ""
      },
      "is_pegin": false,
      "sequence": 4294967295,
      "txinwitness": [
        "304402207e193a17f501f5659ebcb7d93f716158b30876561a6590cfdc159a40d2f50c3102205ca8d81ab4c74cc5d2621d0d5a1ac275671dca208fbbd1c4c614e9f4e6f3305b01",
        "0387992ef847f40f0d21c1d800767a9a3c1c026b29034fcf5a658cfac0214cd87f"
      ],
      "issuance": {
        "assetBlindingNonce": "9a64d614f476cc765e83b4bb8838f8735cef8efd9d240d5148b88b0d83d722de",
        "assetEntropy": "cd84bbf7f5a71fb106b6a068ce14e9dd2eea9fc8d3574810cbdbe6b2241e9a1a",
        "isreissuance": true,
        "asset": "94de85cdbffc86809cf6f75b467eb15d34b4c514a0fa6bd328080eefaa447802",
[...]

PR #1119, opened 2022-06-12
Add a failing test case for reissuance from an unblinded output (#259)

shesek:
A failing test case for #259.

@instagibbs [mentioned](https://github.com/ElementsProject/elements/issues/259#issuecomment-645392524) that it would be good to have a test for this, hopefully this might nudge someone to give this issue some more thought :-)

(To give some context, I ran into this while experimenting with 'scripted assets', where the reissuance token is under an 'anyone can spend if you follow the rules' covenant rather than managed by an issuing entity. It seems that not blinding the token makes more sense for this use-case?)

### DIFF
--- a/test/functional/feature_issuance.py
+++ b/test/functional/feature_issuance.py
@@ -500,5 +500,14 @@ class IssuanceTest(BitcoinTestFramework):
         txid = self.nodes[1].issueasset(0, 1, False)["txid"]
         tx = self.nodes[1].getrawtransaction(txid, True)
 
+        # Test reissuing with a reissuance token spent from an unblinded output
+        # https://github.com/ElementsProject/elements/issues/259
+        issued = self.nodes[0].issueasset(0, Decimal('0.00000001'), False)
+        # Get a non-confidential address and send the (only) reissuance token to it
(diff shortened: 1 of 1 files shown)


sanket1729:
To the best of my knowledge, this is not possible to do. The workaround for my use-case was to use blinding, but fix the blinding factors and assert that the blinded asset is some fixed asset. There is more nuance to this as the input and output confidential assets cannot be the same according to elements consensus rules. So, it required oscillating between two fixed confidential assets with initially decided blinding factors.

If something like this is of interest, I can elaborate more.

tiero:
It may be slightly related, but I can't make a (blinded) re-issuance in the same tx where I am issuing a control asset.

Also this useful in covenant scripts :)

https://github.com/ElementsProject/elements/issues/1103

apoelstra:
@tiero my understanding is that your other issue is caused by the `blindrawtransaction` RPC doing dumb stuff and not having enough flexibility. The PSET also has these problems :) but with PSET we're actively developing and have the ability to basically add arbitrary extra fields to handle new requirements. So that one I hope we can address just by making PSET more powerful.

The problem in this issue unfortunately seems to be caused by bad consensus logic, and there's nothing we can do about that.

PR #1121, opened 2022-07-01
PSET: Various fixes

### LINKED ISSUE #1065: Pset output scalar decoding bug.
The pset spec says that global scalars PSBT_ELEMENTS_GLOBAL_SCALAR must have value None, but as implemented they have value 0100 (a single vector containing the value zero)

In particular, https://github.com/ElementsProject/elements/blob/04cedf3f6be9e8eeb76af6206656ad2516d16f5a/src/psbt.h#L1425 should directly write 0 instead of SerializeToVector(s, std::vector<unsigned char>())

jgriffiths:
- Fix PSET blinding crash
- Fix failure to handle unknown extension fields
- Fix PSBT field serialization order to match core; update test serializations to current expected format
- Add `parsepsbt` rpc for testing
- Fixes serialization of scalars to follow the PSET spec. To reduce interop concerns with 3rd party implementations, accept both the old and new scalar serialization when parsing.

Fixes #1065


sanket1729 (approved):
code review ACK 32c3a98a09ac51365b8ad7d34a875a311a525e41. Double checked that rust-elements parser and parser that allen implemnted https://tools.blockstream.com/psbt parses the new serialization correctly.

src/psbt.h
```diff
@@ -1423,7 +1426,7 @@ struct PartiallySignedTransaction
             // Scalar offsets
             for (const uint256& scalar : m_scalar_offsets) {
                 SerializeToVector(s, CompactSizeWriter(PSBT_GLOBAL_PROPRIETARY), PSBT_ELEMENTS_ID, CompactSizeWriter(PSBT_ELEMENTS_GLOBAL_SCALAR), scalar);
-                SerializeToVector(s, std::vector<unsigned char>());
+                s << PSBT_SEPARATOR; /* Zero length data value */
```
apoelstra, line 1429: I actually don't understand the difference between this line and its replacement. Don't they both just serialize a 0x00?

src/psbt.h
```diff
@@ -137,6 +137,9 @@ template<typename Stream, typename... X>
 void UnserializeFromVector(Stream& s, X&... args)
 {
     size_t expected_size = ReadCompactSize(s);
+    if (!expected_size) {
+        return; /* Zero size = no data to read */
+    }
```
apoelstra, line 142: I'm curious why this isn't needed anywhere upstream. What would happen without this line?

src/psbt.h
apoelstra: Oh no, I'm an idiot, the original line serilaizes two `00`s, because it's serializing a vector in a vector. Okay.

src/psbt.h
apoelstra: Okay, I guess the answer is just that no other expected-to-be-empty fields are de/serialized as vectors. Alright, I'm happy.

apoelstra (approved):
ACK 32c3a98a09ac51365b8ad7d34a875a311a525e41

apoelstra (approved):
utACK b4378ec6b3c4b0787d07c469b566a5745c369f6e

apoelstra (approved):
ACK 32ac36a218183e9a99d1b5fbe70e7a5e1854daf5

apoelstra (approved):
ACK 3be6e80f90b0582316fed22e29da2908c3b6dead

PR #1127, opened 2022-08-01
Fix tarball names in guix and gitian builds

psgreco:
guix builds are still named `bitcoin-` instead of `elements-`, which is fixed by this PR.

Also, since 21 we've had a problem that our tagged builds are named `elements-elements.....` due to the way we tag, this PR also fixes that for both guix and gitian


gwillen:
utACK 7ab26e5. CI failure so far looks spurious. I would like us to merge this ASAP, since I need to bring it into the release branch for 22.

PR #1128, opened 2022-08-01
Fix elements naming in gitian builds

psgreco:
Currently, our gitian builds for releases generate tarballs called `elements-elements-xxxxx` instead of `elements-xxxxx`. This is fixed by correctly trimming the `elements-` prefix in our release tags (just as bitcoin trims the initial `v`)

### DIFF
--- a/contrib/gitian-descriptors/assign_DISTNAME
+++ b/contrib/gitian-descriptors/assign_DISTNAME
@@ -5,6 +5,7 @@
 # A helper script to be sourced into the gitian descriptors
 
 if RECENT_TAG="$(git describe --exact-match HEAD)"; then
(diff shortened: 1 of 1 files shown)


delta1 (approved):
utACK c6677f84b7adc78dc9759b78150255da679c6820 this change will remove the first instance of `elements-` in the `RECENT_TAG` var

psgreco:
> utACK [c6677f8](https://github.com/ElementsProject/elements/commit/c6677f84b7adc78dc9759b78150255da679c6820) this change will remove the first instance of `elements-` in the `RECENT_TAG` var

yes, bitcoin tags their releases with `vXXXX`, we tag them with `elements-XXXX`, this patch strips the `elements-` prefix the same way bitcoin removes their `v`

PR #1129, opened 2022-08-01
Elements 22 rc1

gwillen:
The changes shown here are mostly backports from bitcoin core.

There are several manual changes by me, near the end.

ca2d72a - Removes a new upstream assertion around change-handling in coin selection. I would really like someone familiar with the coin selection algorithm to look at this. As far as I can tell, the assertion either (1) does not apply because we have substantially changed this due to asset handling, or (2) is caught by things we already check. But I'd like that reviewed.

9fb51a2 - Generates updated manual pages.

f4820e5 - Reverts a change to the translation files. To quote `doc/translation_process.md`:
> For general PRs, you shouldn’t include any updates to the translation source files. They will be updated periodically, primarily around pre-releases, allowing time for any new phrases to be translated before public releases. This is also important in avoiding translation related merge conflicts.

If we want to change our policy around how we handle these, we would need to discuss setting up the translation software ourselves. I don't think we should do that; the existing approach will fall back to English when a translation isn't available, and we have not gotten any complaints about that, to my knowledge.


doc/man/elements-util.1
```diff
@@ -0,0 +1,156 @@
+.\" DO NOT MODIFY THIS FILE!  It was generated by help2man 1.47.13.
+.TH ELEMENTS-UTIL "1" "August 2022" "elements-util v22.0.0rc2" "User Commands"
+.SH NAME
+elements-util \- manual page for elements-util v22.0.0rc2
+.SH SYNOPSIS
+.B bitcoin-util
```
psgreco, line 6: should be elements-util here

gwillen: Fixed!

psgreco:
UTACK ff439d8d50e4c827f09c3ece4bffd57e32fc65e8

psgreco:
re-utack ada87dc6db176bfbb2f76d345ef886d692b5cfbd for version fix

apoelstra:
I'm gonna go ahead and merge this but I get consistent failures on `./test/functional/rpc_fundrawtransaction.py` which I think will need to be addressed in a followup PR.

apoelstra:
I fixed the CI stuff in https://github.com/apoelstra/elements/tree/2022-08--fix-fees

gwillen:
@apoelstra took your fix, thanks

psgreco:
ACK 7286cdc3b5592195bd82523f069760acb8e3e312 local tests look good now

safarioops:
.

PR #1131, opened 2022-08-02
Allow dumpblinding key to accept non-CT address

sanket1729:
This facilitates the following workflow:
1) Obtain an updated psbt with in_witness_utxo and in_utxo_rangeproof
2) Get the blinding key from the input utxo address obtained from input
script pubkey without revealing master blinding key
3) Rewind the proof to obtain blinding factors and implement stateless
blinding


apoelstra (approved):
ACK a14223238400209ace0fecdce615df865eecb685

PR #1134, opened 2022-08-08
Check the value assertion only on valid amounts

### LINKED ISSUE #1105: Node crashes when processing partial transaction
I made the node crash while trying to blind the following pset:

```bash
$ e1 walletprocesspsbt cHNldP8BAgQCAAAAAQMEAAAAAAEEAQEBBQEDAfsEAgAAAAABDiAqrJ9c31tcJHnWDjQWJNUfArZMJr1w1h7/S6GIMG31DwEPBAAAAAABEAT/////AAEEFgAUBFTPa7NoO3V+jzsjbmvRlQwcOWoBAwgA4fUFAAAAAAf8BHBzZXQCINMRFPznA5TB+dNUdQG0udNvQgI27GQZkVRWZDSIWs8tAAEEFgAU+a3P0cB6H1+r8pR099KXbz20PbwBAwgAAAAAAAAAAAf8BHBzZXQCICWyUQcOKcoZBDzzPM1zJOLdqwPsxK4LXnfE/A5c9slaB/wEcHNldAYhAv2YM9xdvR/Mxx+ulKcSzcom+JQpzsV/hrxKHexk7BYYB/wEcHNldAgEAAAAAAABBAABAwjoAwAAAAAAAAf8BHBzZXQCICWyUQcOKcoZBDzzPM1zJOLdqwPsxK4LXnfE/A5c9slaAA== false
```

The input of the tx is confidential and its amount is exactly 0.00001, therefore it's spent entirely by the fee out.
The first output of the tx is a non-LBTC unblinded one, therefore i tried to add a dummy LBTC blinded output to balance the tx as `{"el1qqt7esv7utk73lnx8r7hfffcjeh9zd7y5988v2luxh39pmmryastp37ddelguq7slt74l99r57lffwmeaks7mc883qj7565fdg": 0, "blinder_index": 0}` in `createpsbt`.

These are the last logs of the node:

```
2022-04-15T15:39:50Z [default wallet] AddToWallet 02338810e562a4fe99d555e616932f76dbb8301e9e264fcf6e9a3cf56ce5322d  new
2022-04-15T15:39:50Z [default wallet] AddToWallet 0ff56d3088a14bff1ed670bd264cb6021fd52416340ed679245c5bdf5c9fac2a  update
elementsd: blindpsbt.cpp:554: BlindingStatus BlindPSBT(PartiallySignedTransaction&, std::map<unsigned int, std::tuple<long int, CAsset, uint256, uint256> >, std::map<unsigned int, std::pair<CKey, CKey> >): Assertion `rangeresult' failed.
```

Seems like the blinder role fails to create a rangeproof for the dummy output. Am I doing something wrong?

sanket1729:
Fixes #1105 
This causes crash on elements wallet when dealing with transactions that
have explicit values and confidential assets. This creates a somewhat
serious DoS attack as the sender can cause the reciever's wallet to
crash by partially blinding the change output. To make matters worse,
the wallet initially accepts the transaction, but fails while spending
the output.

This is likely caused by a combination of two bugs:
1) The wallet's current behaviour stores the complete transaction of interest
in CWalletTx instead of just Outpoints. Only that the spend time do we
iterate over all outputs, try to unblind them and check which are
isMine. When calling wtx.GetOutputValueOut() or similar calls, we hit this assertion.

While the current behaviour is okay, I think the correct way is move
the IsMine == ISMINE_NO at the start of the loop. We should not do be
any checks on outputs that are not ours. This is used in multiple
places at different parts of the codebase for different RPCs.

2) When dealing with partially blinded trasactions, ComputeBlindingData
correctly sets value = -1, and the cache byte to 1. When getting the
data again with GetBlindingData for explicit value and confidential
asset, we load the precomputed data with value = -1 and assert the
loaded value be the explicit value in the transaction. This is only true
for explicit value and explicit asset.

The changed assertion checks that written value should be same as the
explicit value that was written only when the amounts are valid

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3464,7 +3464,7 @@ void CWalletTx::GetBlindingData(const unsigned int map_index, const std::vector<
         memcpy(asset_tag.begin(), &*(it + 73), 32);
         pubkey.Set(it + 105, it + 138);
 
-        if (conf_value.IsExplicit()) {
+        if (amount != -1 && conf_value.IsExplicit()) {
             assert(conf_value.GetAmount() == amount);
         }
     } else {


apoelstra (approved):
ACK 53a75ebbaec5e982713f04629fbbb04f0e28d137

PR #1130, opened 2022-08-02
QT: fix title in custom chains

psgreco:
In 0.21.x, custom chains appeared in the title in QT, now it just says regtest. This PR recovers that functionality

PR #1137, opened 2022-08-09
[backport] Backport #1121 and #1131 to elements-22.x for rc3

gwillen:
Backport #1121 and #1131 to elements-22.x for rc3.

(Because I ended up just doing this as a merge of master into the release branch, #1130 and #1134 also appear in it, which makes the history confusing -- sorry. Those are no-ops here because the commits were already cherry-picked into the branch.)


jhfrontz:
Compared to locally constructed back port -- only delta is rc version bump.

utack a17887b1718b7c1468f8b0220248847c60636dc4

jamesdorfman:
Compared each commit's diff to their diff on master and confirmed that they match exactly (aside from the rc version bump)

utACK a17887b1718b7c1468f8b0220248847c60636dc4.

PR #1140, opened 2022-08-11
Fix Icon position in dmg

psgreco:
Mac installer is showing the elements icon in a bad position 
image

This small fix puts it back where it should be 
image


philippem:
TACK with e509f565
image

apoelstra:
utACK 7fb99ff33f2e90b704dbcdce819766ed723dcbfd. Confirmed that it only touches Mac stuff.

PR #1141, opened 2022-08-12
Fix tapscript comment

roconnor-blockstream:
In elements tapscript is verison 0xc4


apoelstra (approved):
utACK 707cf20b126788fcec7a29126eebaa83be7dd6a8

PR #1034, opened 2021-09-06
A couple minor PSET cleanups

apoelstra:
.

PR #1145, opened 2022-08-18
Fixing issuance cases and half blinded cases in PSET

allenpiscitello:
A few minor cases fixed in PSET-related code. Caused either a crash or invalid transactions created in some cases.

### DIFF
--- a/src/blindpsbt.cpp
+++ b/src/blindpsbt.cpp
@@ -476,6 +476,9 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
                         }
                     }
                 }
+                else {
+                    input_asset_blinders.emplace_back();
+                }
             }
         }
     }
--- a/src/psbt.cpp
+++ b/src/psbt.cpp
@@ -123,12 +123,25 @@ CMutableTransaction PartiallySignedTransaction::GetUnsignedTx(bool force_unblind
         txin.nSequence = input.sequence.value_or(max_sequence);
         txin.assetIssuance.assetBlindingNonce = input.m_issuance_blinding_nonce;
         txin.assetIssuance.assetEntropy = input.m_issuance_asset_entropy;
-        if (input.m_issuance_value != std::nullopt && input.m_issuance_inflation_keys_amount != std::nullopt && force_unblinded) {
+        // If there is a commitment we should set the value to the commitment unless we are forcing unblinded.
+        // If we are forcing unblinded but there is no value, we just use the commitment.
+        if (input.m_issuance_value != std::nullopt && (input.m_issuance_value_commitment.IsNull() || force_unblinded)) {
             txin.assetIssuance.nAmount.SetToAmount(*input.m_issuance_value);
-            txin.assetIssuance.nInflationKeys.SetToAmount(*input.m_issuance_inflation_keys_amount);
-        } else {
+        }
+        else if(!input.m_issuance_value_commitment.IsNull()) {
             txin.assetIssuance.nAmount = input.m_issuance_value_commitment;
+        }
+        else {
+            txin.assetIssuance.nAmount.SetNull();
+        }
+        if (input.m_issuance_inflation_keys_amount != std::nullopt && (input.m_issuance_inflation_keys_commitment.IsNull() || force_unblinded)) {
+            txin.assetIssuance.nInflationKeys.SetToAmount(*input.m_issuance_value);
+        }
+        else if(!input.m_issuance_inflation_keys_commitment.IsNull()) {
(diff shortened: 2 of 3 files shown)


src/blindpsbt.cpp
```diff
@@ -476,6 +476,9 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
                         }
                     }
                 }
+                else
+                    input_asset_blinders.emplace_back();
```
allenpiscitello, line 480: In the case of unblinded issuances, we need to add a null value here for a valid surjection proof to exist.

src/psbt.cpp
```diff
@@ -123,13 +123,26 @@ CMutableTransaction PartiallySignedTransaction::GetUnsignedTx(bool force_unblind
         txin.nSequence = input.sequence.value_or(max_sequence);
         txin.assetIssuance.assetBlindingNonce = input.m_issuance_blinding_nonce;
         txin.assetIssuance.assetEntropy = input.m_issuance_asset_entropy;
-        if (input.m_issuance_value != std::nullopt && input.m_issuance_inflation_keys_amount != std::nullopt && force_unblinded) {
+        // If there is a commitment we should set the value to the commitment unless we are forcing unblinded.
```
allenpiscitello, line 126: This code handled blinded vs. unblinded issuances badly. It always tried to blind the issuances, even if they were unblinded issuance assets.

Since there are no fields in the PSET spec to define if an issuance should be blinded or not, we will use the presence of the issuance blinding commitment to tell us what to do.

src/wallet/wallet.cpp
```diff
@@ -2048,11 +2048,15 @@ TransactionError CWallet::SignPSBT(PartiallySignedTransaction& psbtx, bool& comp
             txin.assetIssuance.nAmount = input.m_issuance_value_commitment;
         } else if (input.m_issuance_value) {
             txin.assetIssuance.nAmount.SetToAmount(*input.m_issuance_value);
+        } else {
```
allenpiscitello, line 2051: These variables were never cleared, so they retained the value of the previous loop iteration, which resulted in problems when there were multiple inputs beyond the first issuance, as well as problems when reissuing assets (where there are no inflation keys created).

src/psbt.cpp
```diff
@@ -531,12 +544,9 @@ bool PSBTOutput::Merge(const PSBTOutput& output)
 CTxOut PSBTOutput::GetTxOut() const
 {
```
allenpiscitello, line 545: This would fail on situations where an asset was blinded but a value was not. I simplified the code to ensure that a value or commitment was checked in the assert (note: this will still crash if you don't provide either, maybe we should't assert?). Then it will use whatever combination of what is given.

apoelstra (approved):
utACK 8e1153a5a6cb1d2d2f9cab4c645fe2d39c2a4868

(I cannot run tests because of #1034)

src/blindpsbt.cpp
psgreco: can we add the missing `{}` here?

src/psbt.cpp
allenpiscitello: This looks similar that it may solve the problem. The use case I was trying to solve was an unblinded re-issuance.

psgreco: Just tested #1119 which provides a testcase for #1103 and it is not fixed by this

psgreco:
Tests pass after the assert fix, @apoelstra can you re-ack, even after I merge this?

apoelstra:
re-utACK 5954b10a1aa5e55ecd3ea0b8a9c0e01ab6cdcda4

src/blindpsbt.cpp
```diff
@@ -476,6 +476,9 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
                         }
                     }
                 }
+                else {
```
jgriffiths, line 479: Prefer to have this as the if, with the existing code in the else. As it is its difficult to see what if this else relates to when reviewing. i.e.

```
                if (!blind_issuance) {
                    input_asset_blinders.emplace_back();
                } else {
```

jgriffiths: Also all of your `else` statements (here and elsewhere) are on new lines which is inconsistent with the rest of the code. Please use
```
} else {
```
Instead of
```
}
else {
```

Here and in the other changes.

tiero: for code formatting/style wouldn't be better to rely on a GitHub action that enforces shared rules automatically?

jgriffiths: That would make the code harder to rebase over upstream core changes, unfortunately.

Issue #1144, opened 2022-08-16
Elements-qt: the arrows to increase/decrease the "Amount" don't work in the Send page


andreabonel:
Tested on elements 0.21.0.2 and v22.0.0rc2
![Screenshot from 2022-08-16 19-44-30](https://user-images.githubusercontent.com/94554885/184998158-8908f727-bb90-498a-8a23-eb07d8241249.png)

PR #1146, opened 2022-08-18
Elements-qt: Fix arrows to increase/decrease Amount

andreabonel:
Fixes: #1144

The problem was that Elements uses different defaults for the "step" between coins and assets, but these defaults were ignored after calculation.

This PR just uses the right variable for the step calculation.


apoelstra (approved):
utACK 38e9490de58198af8dccfe062395ee7a52607897 confirmed that this only touches qt stuff

PR #1148, opened 2022-08-19
docs: describe elements transaction serialization format

jamesdorfman:
**Note**: When reviewing this markdown file, please open it in a markdown viewer. Otherwise, the tables will be illegible. Github will show you the rendered markdown if you open the "Files changed" tab, click the three dots on the top right of the file, and select "View file".

This PR adds a new markdown document which contains a detailed description of the elements transaction serialization format.

The document includes a detailed descriptions of all fields and examples where complex raw transactions are deserialized, step-by-step.

Intended audience:
- Hardware Wallet developers implementing liquid apps
- Developers creating Liquid integrations
- Developers writing their own serialized Liquid transactions from other languages
- Bitcoin users wanting to see what's new in an Elements transactions
- People creating and using PSETs (PSETs use serialized transactions)
- People wanting to know how Liquid and Elements work

### DIFF
--- a/doc/elements-tx-format.md
+++ b/doc/elements-tx-format.md
@@ -1,0 +1,570 @@
+# Elements Transaction Format
+
+This document describes the format used to serialize elements transactions. Once a transaction has been converted into this raw, serialized form, it can be broadcast across the network.
+
+This document requires familiarity with the architecture of bitcoin and elements, as well the various parts that are expected in an elements transaction (inputs, outputs, assets, pegin data, etc.). For more information on those, please refer to the [elements code tutorial](https://elementsproject.org/elements-code-tutorial/overview).
+
+### Data Types
+
+*Note:* The encoding column in the following tables corresponds to the way that the fields are serialized. In bitcoin, numbers and hashes are serialized in little endian, and this same notion carries over to elements.
+
+*Note:* Values are listed in multiples of 100,000,000 (this corresponds directly to how bitcoin encodes value, where 1 BTC is divisible into 100,000,000 satoshis). For example, a `ConfidentialAmount` representing 7 inflation keys would be encoded as 700,000,000.
+
+#### Transaction
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Version | Yes | 4 bytes | `UnsignedInteger` | Little endian | Protocol version (should always be 2). |
+| Flags | Yes | 1 byte | | | Whether the transaction contains a witness. Either 0 or 1. |
+| Num Inputs | Yes | Varies | `VarInt` | | Number of inputs to the transaction. |
+| Inputs | Yes | Varies | `Vector<TxIn>` | | |
+| Num Inputs | Yes | Varies | `VarInt` | | Number of outputs from the transaction. |
+| Outputs | Yes | Varies | `Vector<TxOut>` | | |
+| Locktime | Yes | 4 bytes | `UnsignedInteger` | Little endian | See [BIP 113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki). |
+| Witness | Only if flags is 1 | Varies | `Witness` | | See [BIP 141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki). Note that elements witnesses contain more data than bitcoin witnesses. This extra data is described further below. |
+
+Serialized format: `[Version][Flags][Num Inputs][Inputs][Num Outputs][Outputs][Locktime][Witness]`
+
+This is the overarching structure of a serialized transaction. The rest of this document contains further details on specific parts, as well as examples.
+
+#### TxIn
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Output Hash (TXID) | Yes | 32 bytes | `UnsignedInteger` | Little endian | Hashes in bitcoin are encoded as integers in little endian. Be mindful – this means that they will look backwards when serialized. |
+| Output Index | Yes | 4 bytes | `UnsignedInteger` | Little endian | **Output is a coinbase**: `0xffffffff`<br><br>The two most significant bits are reserved for flags.<br><br>**Output is a pegin:** second most significant bit is 1.<br><br>**Output is an asset issuance:** most significant bit is 1. |
+| ScriptSig Length | Yes | Varies | `VarInt` | | Set to `0x00` if the transaction is segwit and the witness contains the signature. |
+| ScriptSig | If ScriptSig Length is non-zero | Varies | `hex` | Big endian | |
+| Sequence | Yes | 4 bytes | `UnsignedInteger` | Little endian | |
+| Asset Issuance | Only if the transaction has an issuance (as indicated by the Output Index) | Varies | `AssetIssuance` |  | |
+
+Serialized format: `[Output Hash][Output Index][ScriptSig Length][ScriptSig][Sequence][Asset Issuance]`
+
+#### TxOut
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Asset | Yes | 1 or 33 bytes | `ConfidentialAsset` | |  |
+| Amount | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | |
+| Nonce | Yes | 1 or 33 bytes | `ConfidentialNonce` | | |
+| ScriptPubKey Length | Yes | Varies | `VarInt` | | | |
+| ScriptPubKey | If ScriptPubKey Length is non-zero | Varies | `hex` | Big endian | |
+
+Serialized format: `[Asset][Amount][Nonce][ScriptPubKey Length][ScriptPubKey]`
+
+#### Witness
+
+As noted above, the witness is only present for segwit transactions.
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Input Witnesses| Yes | Varies | `Vector<InputWitness>` | | There is exactly one input witness for each input in the transaction.<br><br>This number is not explicitly included in the witness — it is implied by the number of inputs. |
+| Output Witnesses| Yes | Varies | `Vector<OutputWitness>` | | There is exactly one output witness for each output in the transaction.<br><br>This number is not explicitly included in the witness — it is implied by the number of outputs. |
+
+Serialized format: `[Input Witnesses][Output Witnesses]`
+
+#### InputWitness
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Issuance Amount Range Proof | Yes | Varies | `Proof` | | `0x00` → null. |
+| Inflation Keys Range Proof | Yes | Varies | `Proof` | | `0x00` → null. |
+| Script Witness | Yes | Varies | `Vec<hex>` | | The vector represents the witness stack.<br>Can be empty (length of 0). |
+| Pegin Witness | Yes | Varies | `Vec<hex>` | | The vector represents the witness stack.<br>Can be empty (length of 0). |
+
+Serialized format:
+`[Issuance Amount Range Proof][Inflation Keys Range Proof][Script Witness][Pegin Witness]`
+
+The range proofs must be empty if their asociated amount (issuance / inflation keys) is explicit.
+
+#### OutputWitness
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Surjection Proof | Yes | Varies | `Proof` | | Will be null (`0x00`) if corresponding output is not confidential. |
+| Range Proof | Yes | Varies | `Proof` | | Will be null (`0x00`) if corresponding output is not confidential. |
+
+Serialized format: `[Surjection Proof][Range Proof]`
+
+The Range Proof must be empty if the output’s amount is explicit. 
+
+The Surjectionproof must be empty if the output’s assetId is explicit.
+
+#### AssetIssuance
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Asset Blinding Nonce | Yes | 32 bytes | `hex` | | Zero for a new asset issuance; otherwise a blinding factor for the input. |
+| Asset Entropy | Yes | 32 bytes | `hex` | | Freeform entropy field, no consensus-defined meaning, but is used as additional entropy to the asset tag calculation. |
+| Amount | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | Amount of the asset to issue. Both explicit and blinded amounts are supported. |
+| Num Inflation Keys | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | Number of inflation keys to issue. Both explicit and blinded amounts are supported. |
+
+Serialized format: `[Asset Blinding Nonce][Asset entropy][Amount][Num Inflation Keys]`
+
+#### Variable Length Integer (VarInt)
+
+This data type is derived from bitcoin and allows an integer to be encoded depending on the represented value, to save space. Variable length integers always precede an vector of a type of data that may vary in length. Longer numbers are encoded in little endian.
+
+Refer to the [bitcoin protocol documentation](https://en.bitcoin.it/wiki/Protocol_documentation#Variable_length_integer) for a full description on how this type is serialized.
+
+#### Vector\<Type\>
+
+Serialized format: `[Length (n)][Item #1][Item #2][...][Item #n]`
+
+Each `Vector` begins with a `VarInt` describing the number of items it contains.
+
+If the vector is of type `hex`, then the size / structure of each individual item is not known in advance. In this case, each item begins with a `VarInt` describing its size `s` in bytes, followed by `s` bytes which should be interpreted as the item itself.
+Otherwise, size prefixes are omitted, and each item should be interpreted in accordance with the vector's type.
+
+Refer to the examples section below for more concrete examples of serialized vectors.
+
+#### ConfidentialAsset
+
+Serialized format: `[header][value]`
+
+A header byte of `0x00` indicates a “null” value with no subsequent bytes.
+A header byte of `0x01` indicates an “explicit” value with the following 32 bytes denoting the key used to generate the assetID (little endian).
+A header byte of `0x0a` or `0x0b` indicates a blinded value encoded as a compressed elliptic curve point. With the least significant bit of the header byte denoting the least significant bit of the y-coordinate, and the remaining 32 bytes denoting the x-coordinate (big endian). The point must be a point on the curve.
+
+#### ConfidentialAmount
+
+Serialized format: `[header][value]`
+
+A header byte of `0x00` indicates a “null” value with no subsequent bytes.
+A header byte of `0x01` indicates an “explicit” value with the following 8 bytes denoting a 64-bit value (big endian).  This value must be between 0 and `MAX_MONEY` inclusive.
+A header byte of `0x08` or `0x09` indicates a blinded value encoded as a compressed elliptic curve point. With the least significant bit of the header byte denoting the least significant bit of the y-coordinate, and the remaining 32 bytes denoting the x-coordinate (big endian). The point must be a point on the curve.
+
+#### ConfidentialNonce
+Serialized format: `[header][value]`
+
+A header byte of `0x00` indicates a “null” value with no subsequent bytes.
+A header byte of `0x01` indicates an “explicit” value with the following 32 bytes denoting a value (big endian).
+A header byte of `0x02` or `0x03` indicates a compressed elliptic curve point. With the least significant bit of the header byte denoting the least significant bit of the y-coordinate, and the remaining 32 bytes denoting the x-coordinate (big endian). This point is not required to be on the curve.
+
+#### Proof
+Serialized format: `[length][value]`
+
+The length is a VarInt, and specifies the number of bytes in the subsequent value (big endian).
+A length of `0x00` signifies a "null" value with no subsequent bytes.
+
+## Examples
+
+#### Example 1
+Signed transaction on liquidtestnet, moving 2 existing tL-BTC outputs into two new outputs with sizes: 0.00055 tL-BTC, 0.01 tL-BTC.
+
+Raw hex:
+```
+02000000010290683fa6b0393b659df707b65e05cbfdf92cc2688589a0ed4931e8a61dfe64c40000000000ffffffff8d83eb1b0826f46d473003d041116927470e2ce0f7cc0c634a983d438d770ac80000000000ffffffff0201499a818545f6bae39fc03b637f2a4e1e64e590cac1bc3a6f6d71aa4443654c1401000000000000d6d8028272c6d7d23b5d8836daac705b4a312b0ebc03f87290da2ed511a833096f0d8e1600144447cb259b2a9857922c5a19cd6449b227dd6c3501499a818545f6bae39fc03b637f2a4e1e64e590cac1bc3a6f6d71aa4443654c140100000000000f42400372fdd5c6e805a50d73ab15ec41cfaadcbe408ecc7a5867621918f1070f84ec9516001424ae71d4804ca7dd1fa66486a87af9dff1663c8400000000000002463043021f3a851011d5b7b52c9761e8a0bd558f2d43129bd1602ffc30945c0a9eaeeaa4022078763819f41ed3084d88fbd3ccf00f8107db76a431a3c085191e3aceb1cb12b0012103207312a1d3e7c2aa5c5212e18f2fe0095e8d11e9ca78f4acc7f6e40dbd21d2ed0000000247304402203e13aa1c792fc14ff06f2c2269e7fff35dc7710dbd73bca3738d21888f031a72022008f9eeb1ea00a8da358321a23e9cd14f6163f609ba391205d5eadcf47661288f012103daa4c9684c369d4420b9f20e4116362b6a3786e9a12a4e1927d1a25c4ffd931b0000000000
+```
+
+Deserialization:
+```
+02000000 ............................. Version
+
+01 ................................... Flags (0x01 → witness data is present)
+
+02 ................................... Num Inputs
+|
+|                                      Input #1
+| 90683fa6b0393b659df707b65e05cbfd
+| f92cc2688589a0ed4931e8a61dfe64c4 ... Outpoint TXID
+| 00000000 ........................... Outpoint index
+|
+| 00 ................................. ScriptSig length
+| | .................................. ScriptSig (empty: segwit transaction)
+|
+| ffffffff ........................... Sequence number: UINT32_MAX
+|
+|                                      Input #2
+| 8d83eb1b0826f46d473003d041116927
+| 470e2ce0f7cc0c634a983d438d770ac8 ... Outpoint TXID
+| 00000000 ........................... Outpoint index 
+|
+| 00 ................................. ScriptSig length
+| | .................................. ScriptSig (empty)
+|
+| ffffffff ........................... Sequence number: UINT32_MAX
+ 
+02 ................................... Num Outputs
+|
+|                                      Output #1
+| 01 ................................. Asset Header (0x01 → explicit)
+| 499a818545f6bae39fc03b637f2a4e1e
+| 64e590cac1bc3a6f6d71aa4443654c14 ... Key used to generate assetID
+|
+| 01 ................................. Amount header (0x01 → explicit)
+| 000000000000d6d8 ................... Amount: 0xd6d8 tL-Satoshis = 0.00055 tL-BTC
+|
+| 02 ................................. Nonce header (0x02 → compressed point)
+| 8272c6d7d23b5d8836daac705b4a312b
+| 0ebc03f87290da2ed511a833096f0d8e ... Nonce x-coordinate (big-endian)
+|
+| 16 ................................. ScriptPubKey length (0x16 = 22 bytes)
+| | 00144447cb259b2a9857922c5a19cd
+| | 6449b227dd6c35 ................... ScriptPubKey
+|
+|                                      Output #2
+| 01 ................................. Asset header (0x01 → explicit)
+| 499a818545f6bae39fc03b637f2a4e1e
+| 64e590cac1bc3a6f6d71aa4443654c14 ... Key used to generate assetID
+|
+| 01 ................................. Amount header (0x01 → explicit)
+| 00000000000f4240 ................... Amount: 0xf4240 tL-Satoshis = 0.01 tL-BTC
+|
+| 03 ................................. Nonce header (0x03 → compressed point)
+| 72fdd5c6e805a50d73ab15ec41cfaadc
+| be408ecc7a5867621918f1070f84ec95 ... Nonce x-coordinate (big-endian)
+| 
+| 16 ................................. ScriptPubKey length (0x16 = 22 bytes)
+| | 001424ae71d4804ca7dd1fa66486a8
+| | 7af9dff1663c84 ................... ScriptPubKey
+
+00000000 ............................. Locktime
+
+...................................... Input witnesses (1 per input)
+|
+|                                      Input #1 witness
+| 00 ................................. Issuance amount range proof (null)
+| 00 ................................. Inflation keys range proof (null)
+| 02 ................................. Script witness stack length
+| | 46 ............................... Stack item #1 length (0x46 = 70 bytes)
+| | | 3043021f3a851011d5b7b52c9761
+| | | e8a0bd558f2d43129bd1602ffc30
+| | | 945c0a9eaeeaa4022078763819f4
+| | | 1ed3084d88fbd3ccf00f8107db76
+| | | a431a3c085191e3aceb1cb12b001 ... Stack item #1
+| | 21 ............................... Stack item #2 length (0x21 = 33 bytes)
+| | | 03207312a1d3e7c2aa5c5212e18f
+| | | 2fe0095e8d11e9ca78f4acc7f6e4
+| | | 0dbd21d2ed ..................... Stack item #2
+| 00 ................................. Pegin witness (0x00 → null)
+|
+|                                      Input #2 witness
+| 00 ................................. IssuanceAmountRangeproof (null)
+| 00 ................................. InflationKeysRangeproof (null)
+| 02 ................................. Script witness stack length
+| | 47 ............................... Stack item #1 length (0x47 = 71 bytes)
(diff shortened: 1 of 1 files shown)


delta1 (commented):
Great work, I have just done a first pass with mostly grammatical suggestions, and will have a second pass when I can and go over the examples closely.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
+# Elements Transaction Format
+
+This document describes the format used to serialize elements transactions. Once a transaction has been converted into this raw, serialized form, it can be broadcast across the network.
+
+This document requires familiarity with the architecture of bitcoin and elements, as well the various parts that are expected in an elements transaction (inputs, outputs, assets, pegin data, etc.). For more information on those, please refer to the [elements code tutorial](https://elementsproject.org/elements-code-tutorial/overview).
```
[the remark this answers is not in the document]
jamesdorfman: Thanks! I will add all these, except for the "peg-in" one.

It appears that the existing elements docs use "pegin", not "peg-in".

> claimpegin "bitcoinTx" "txoutproof" ( "claim_script" )
> Claim coins from the main chain by creating a pegin transaction with the necessary metadata after the corresponding Bitcoin transaction.

https://elementsproject.org/en/doc/0.18.1.6/rpc/wallet/claimpegin/

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+
+*Note:* Values are listed in multiples of 100,000,000 (this corresponds directly to how bitcoin encodes value, where 1 BTC is divisible into 100,000,000 satoshis). For example, a `ConfidentialAmount` representing 7 inflation keys would be encoded as 700,000,000.
+
+#### Transaction
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Version | Yes | 4 bytes | `UnsignedInteger` | Little endian | Protocol version (should always be 2). |
+| Flags | Yes | 1 byte | | | Whether the transaction contains a witness. Either 0 or 1. |
+| Num Inputs | Yes | Varies | `VarInt` | | Number of inputs to the transaction. |
```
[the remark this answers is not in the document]
jamesdorfman: I sort of agree with you... however, I think that noting this might create more confusion.

For one, VarInts also have their own specific encoding (little-endian).
Additionally, we can't state any other concrete type (such as `uint32`) because we don't know the integer's size in advance.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+| Asset Issuance | Only if the transaction has an issuance (as indicated by the Output Index) | Varies | `AssetIssuance` |  | |
+
+Serialized format: `[Output Hash][Output Index][ScriptSig Length][ScriptSig][Sequence][Asset Issuance]`
+
+#### TxOut
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Asset | Yes | 1 or 33 bytes | `ConfidentialAsset` | |  |
+| Amount | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | |
```
roconnor-blockstream, line 48: Null asset and null amounts are illegal in outputs.  They cannot be 1 byte.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+The Range Proof must be empty if the output’s amount is explicit. 
+
+The Surjectionproof must be empty if the output’s assetId is explicit.
+
+#### AssetIssuance
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Asset Blinding Nonce | Yes | 32 bytes | `hex` | | Zero for a new asset issuance; otherwise a blinding factor for the input. |
+| Asset Entropy | Yes | 32 bytes | `hex` | | Freeform entropy field, no consensus-defined meaning, but is used as additional entropy to the asset tag calculation. |
```
roconnor-blockstream, line 98: This is only freeform for new issuances.  For reissuances it is required to be the asset's entropy value.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+| Amount | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | Amount of the asset to issue. Both explicit and blinded amounts are supported. |
+| Num Inflation Keys | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | Number of inflation keys to issue. Both explicit and blinded amounts are supported. |
```
roconnor-blockstream, line 100: Note: Inflation Keys cannot be reissued.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+Serialized format: `[Version][Flags][Num Inputs][Inputs][Num Outputs][Outputs][Locktime][Witness]`
+
+This is the overarching structure of a serialized transaction. The rest of this document contains further details on specific parts, as well as examples.
+
+#### TxIn
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Output Hash (TXID) | Yes | 32 bytes | `UnsignedInteger` | Little endian | Hashes in bitcoin are encoded as integers in little endian. Be mindful – this means that they will look backwards when serialized. |
+| Output Index | Yes | 4 bytes | `UnsignedInteger` | Little endian | **Output is a coinbase**: `0xffffffff`<br><br>The two most significant bits are reserved for flags.<br><br>**Output is a pegin:** second most significant bit is 1.<br><br>**Output is an asset issuance:** most significant bit is 1. |
```
roconnor-blockstream, line 35: *Input* is a pegin; *Input* has an asset issuance.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+| Version | Yes | 4 bytes | `int32_t` | Little-endian | Transaction version number. Currently version 2 (see [BIP 68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki#specification)). | |
+| Flags | Yes | 1 byte | `unsigned char` | | Whether the transaction contains a witness. Either 0 (no witness) or 1. |
+| Num Inputs | Yes | Varies | `VarInt` | | Number of inputs to the transaction. |
+| Inputs | Yes | Varies | `Vector<TxIn>` | | |
+| Num Inputs | Yes | Varies | `VarInt` | | Number of outputs from the transaction. |
+| Outputs | Yes | Varies | `Vector<TxOut>` | | |
+| Locktime | Yes | 4 bytes | `uint32_t` | Little-endian | See [BIP 113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki). |
+| Witness | Only if flags is 1 | Varies | `Witness` | | See [BIP 141](https://github.com/bitcoin/bips/blob/master/bip-0141.mediawiki). Note that Elements witnesses contain more data than Bitcoin witnesses. This extra data is described further below. |
+
+Serialized format: `[Version][Flags][Num Inputs][Inputs][Num Outputs][Outputs][Locktime][Witness]`
```
philippem, line 26: Do we need to have the serialized form listed here? The table is sequential and conveys the form.  Duplicating it makes it redundant and the reader may wonder if it differs .

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+
+This is the overarching structure of a serialized transaction. The rest of this document contains further details on specific parts, as well as examples.
+
+#### TxIn
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Output Hash (TXID) | Yes | 32 bytes | `uint256` | Little-endian | Bytes of the output hash are encoded in little-endian. **NB**: this means that they appear backwards when serialized. |
```
philippem, line 34: Can probably omit the NB. We can assume familiarity with little-endian and big-endian serializations.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+*Note:* The encoding column in the following tables corresponds to the way that the fields are serialized. In Bitcoin, numbers and hashes are serialized in little-endian, and this same notion carries over to Elements.
+
+*Note:* Values are listed in multiples of 100,000,000 (this corresponds directly to how Bitcoin encodes value, where 1 BTC is divisible into 100,000,000 Satoshis). For example, a `ConfidentialAmount` representing 7 inflation keys would be encoded as 700,000,000.
+
+#### Transaction
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Version | Yes | 4 bytes | `int32_t` | Little-endian | Transaction version number. Currently version 2 (see [BIP 68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki#specification)). | |
+| Flags | Yes | 1 byte | `unsigned char` | | Whether the transaction contains a witness. Either 0 (no witness) or 1. |
```
philippem, line 18: We should state if other values are accepted or ignored. (In this case, they are rejected)

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Input Witnesses| Yes | Varies | `Vector<InputWitness>` | | There is exactly one input witness for each input in the transaction.<br><br>This number is not explicitly included in the witness — it is implied by the number of inputs. |
+| Output Witnesses| Yes | Varies | `Vector<OutputWitness>` | | There is exactly one output witness for each output in the transaction.<br><br>This number is not explicitly included in the witness — it is implied by the number of outputs. |
+
+Serialized format: `[Input Witnesses][Output Witnesses]`
+
+#### InputWitness
+
```
philippem, line 67: Might be helpful to add a one sentence description of this field, in case someone is arriving here without the previous context.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+| | | 1ed3084d88fbd3ccf00f8107db76
+| | | a431a3c085191e3aceb1cb12b001 ... Stack item #1
+| | 21 ............................... Stack item #2 length (0x21 = 33 bytes)
+| | | 03207312a1d3e7c2aa5c5212e18f
+| | | 2fe0095e8d11e9ca78f4acc7f6e4
+| | | 0dbd21d2ed ..................... Stack item #2
+| 00 ................................. Pegin witness (0x00 → null)
+|
+|                                      Input #2 witness
+| 00 ................................. IssuanceAmountRangeproof (null)
```
philippem, line 243: the use of camel case here is inconsistent with input 1

doc/elements-tx-format.md
roconnor-blockstream: FWIW, I think the situation here is more complex than this documentation presents.  The serialization order of the hashes *is* the correct order of the bytes produced by SHA256.  It is the elements client that takes this correct serialization order, reinterprets it as a 256 bit integer in little endian format and then displays that integer in hex, which has the effect of displaying AssetIDs backwards.

I just don't know a concise way of explaining all this.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+Serialized format: `[Version][Flags][Num Inputs][Inputs][Num Outputs][Outputs][Locktime][Witness]`
+
+This is the overarching structure of a serialized transaction. The rest of this document contains further details on specific parts, as well as examples.
+
+#### TxIn
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Output Hash (TXID) | Yes | 32 bytes | `uint256` | Little-endian | Bytes of the output hash are encoded in little-endian. **NB**: this means that they appear backwards when serialized. |
+| Output Index | Yes | 4 bytes | `uint32_t` | Little-endian | **Input is a coinbase**: `0xffffffff`<br><br>The two most significant bits are reserved for flags.<br><br>**Input is a pegin:** second most significant bit is 1.<br><br>**Input has an asset issuance:** most significant bit is 1. |
```
philippem, line 35: what happens if both bits are 1?

doc/elements-tx-format.md
```diff
@@ -0,0 +1,570 @@
...
+Each `Vector` begins with a `VarInt` describing the number of items it contains.
+
+If the vector is of type `hex`, then the size / structure of each individual item is not known in advance. In this case, each item begins with a `VarInt` describing its size `s` in bytes, followed by `s` bytes which should be interpreted as the item itself.
+Otherwise, size prefixes are omitted, and each item should be interpreted in accordance with the vector's type.
+
+Refer to the examples section below for more concrete examples of serialized vectors.
+
+#### ConfidentialAsset
+
+Serialized format: `[header][value]`
```
philippem, line 123: might be helpful to keep with the table description for these types, for consistency and clarity (even though there are only 2 fields).

doc/elements-tx-format.md
jamesdorfman: I see. Thank you for the clarification.

I was actually under the (incorrect) impression that the elements client displayed the correct order of bytes produced by SHA256, and the hashes were just serialized in reverse. 

Your explanation makes much more sense.

I will make this distinction more clear.

doc/elements-tx-format.md
jamesdorfman: I think it's self-evident that an asset issuance and a pegin are mutually exclusive, from a logical perspective. 

I'd prefer not to enumerate every single thing that could be invalid, because that would make the document much longer and more confusing to read.

jamesdorfman: I think it's self-evident that an asset issuance and a pegin are mutually exclusive, from a logical perspective.

I'd prefer not to enumerate every single thing that could be invalid, because that would make the document much longer and more confusing to read.

I could add a link to the actual transaction validation code [here](https://github.com/ElementsProject/elements/blob/a49c01693b27148cdf188c85b89b83fe0b598ba1/src/consensus/tx_check.cpp#L10), but I'm not sure how helpful that will be.

doc/elements-tx-format.md
apoelstra: I think we should drop any reference to endianness and just say that the bytes are reversed. You can add a footnote explaining that historically this has something to do with endianness and type conversions but I don't think it would make things any clearer.

doc/elements-tx-format.md
roconnor-blockstream: As far as I am aware you can pegin and issue an asset at the same time. I've never tried it though.

doc/elements-tx-format.md
jamesdorfman: Done. I've stated that the bytes are reversed, and I've added a footnote explaining the historical reasons for this.

doc/elements-tx-format.md
jamesdorfman: Fixed.

doc/elements-tx-format.md
jamesdorfman: I've added a note here explaining this.

doc/elements-tx-format.md
jamesdorfman: Added.

doc/elements-tx-format.md
jamesdorfman: I've added notes explaining this.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+This document describes the format used to serialize Elements transactions. Once a transaction has been converted into this raw, serialized form, it can be broadcast across the network.
+
+This document assumes some familiarity with Bitcoin and Elements (UTXOs, [Script](https://en.bitcoin.it/wiki/Script), assets, pegins, etc.). For more information on those, please refer to the [Bitcoin Wiki](https://en.bitcoin.it/wiki/Main_Page) and the [Elements Code Tutorial](https://elementsproject.org/elements-code-tutorial/overview).
+
+### Data Types
+
+*Notes*:
+* Fields in the following table are listed in the same order in which they are serialized.
+
+* The *Encoding* column in the following tables indicates how the fields are serialized. All fields in Elements which are the result of a hash (TXID, asset ID, etc.) have their bytes serialized in reverse order. For example: the hash `0x112345...` would be serialized as `...543211`.[^1]
```
roconnor-blockstream, line 12: I don't mean to keep harping on this, and if you want to leave it as it, that is acceptable, but I think this still still all very confusing/misleading.

A hash, as in the output of the SHA-256 function, is **not** being serialized in reverse order.  Instead it is the Element's CLI and GUI interfaces that are displaying these hash values in reverse byte-order.

I get the problem you are having: your main point here is that the order serialized and the order displayed by the Elements UI are reversed from one and other, so who is to say which one is backwards and which one is forward?

But for a developer who is making transactions and computing hashes by hand, there is a clear canonical order: the order produced by ever SHA-256 library out there.  And that canonical order is the order that appears in the serialization format.

apoelstra: Agreed -- nothing in the binary serialization is reversed. I think it is confusing to the point of being incorrect to say otherwise.

jamesdorfman: I understand that the SHA-256 algorithm's output corresponds directly to the serialized hash. However, is this what we want to call the "canonical order"?

If someone sends me a payment and I ask them "what is the TXID", should I be expecting them to send me the reverse of what the Elements client will show (i.e. the aforementioned canonical order)? I would personally be expecting to see the same TXID that Elements would show.

I am fine with either way, I just want to confirm that we are all on the same page. Thanks! :) 

(And for what it's worth, I briefly touched on this subtlety in the attached footnote:
> [^1]: This is for historical reasons. In Bitcoin, numbers and hashes are interpreted as being encoded in little-endian, and this same policy has been carried over to Elements. So, when Elements encounters a hash (TXID, Asset ID, etc.) it interprets it as a little-endian integer and parses the bytes backwards.
)

apoelstra: To be clear, in my earlier comments suggesting "reversed" language rather than "endianness" language, I missed the context, which is that the binary encoding doesn't actually have any hash reversals in it so there is no need to mention this stuff at all.

apoelstra: >I understand that the SHA-256 algorithm's output corresponds directly to the serialized hash. However, is this what we want to call the "canonical order"?

Yes.

>If someone sends me a payment and I ask them "what is the TXID", should I be expecting them to send me the reverse of what the Elements client will show (i.e. the aforementioned canonical order)? I would personally be expecting to see the same TXID that Elements would show.

I would be expecting them to send a hex-encoded byte-reversed version of the txid. It is easy to tell that it is byte-reversed because it is in hex.

jamesdorfman: I agree with you that this is not directly relevant here.

However, a serialized transaction contains outpoint TXIDs, which are reversed when serialized / deserialized. So, if a developer is trying to create or parse a raw transaction, this might be useful for them to understand.

Nevertheless, it is certainly possible to entirely omit any mentions of this, and that might be the best move.

apoelstra: Yeah, I think it's reasonable to add a footnote saying that hex encodings of these values are usually reversed, and will need to be re-reversed to match the serialized data. And also mention that this is exactly the same as the situation in Bitcoin.

roconnor-blockstream: The canonical order of a TXID is the reverse of the canonical order of a SHA-256 hash.  I totally get why this is a mess to explain.

apoelstra: I don't like this framing -- the implication is then that the serialized order of a txid (in a transaction input) is reversed.

roconnor-blockstream: The serialization order is reversed for a TXID.  Certainly when I ask for a TXID what I get is a hex string that is the reverse of what is serialized in the transaction.

roconnor-blockstream: I think my suggestion is not to call this big or little endian.  There are two orders: hash order and txid/assetid order.  These two orders are reversed from each other.  The format being serialized is hash order, which is the reverse of the txid and assetid orders.

apoelstra: I can live with that, although I still find it confusing because it's inconsistent with the way that I think about these objects.

I consider "reversed hex" to be the encoding of these types and still consider the natural order of their bytes to be the ones that come out of the sha2 engine, even for TXIDs/blockhashes/etc (which I agree are distinct types from sha2 hashes, if only because they are encoded differently for user display :)).

jamesdorfman: Update: thank you for your feedback. I have omitted all mentions of this byte-reversal from the body of the document. Instead, in the encoding column for the *Output Hash (TXID)* field, I have linked to a footnote which reads:

> The hex encodings of hashes by the Elements client (TXID, asset ID) are reversed, and so they will need to be re-reversed to match the serialized data. This is the same situation as in Bitcoin. For example, the hash `0x1123...7899` would be displayed by the Bitcoin and Elements clients as `9987...3211`. This is primarily for historical reasons: the Bitcoin client has always interpreted and displayed hashes as little-endian integers and parsed their bytes in reverse order.


In other words, I have explained the situation without identifying which order is the "canonical order'. I have simply stated which order is the hash order, and which order is used for display by the Elements client. The user can interpret it how they wish, so long as they understand that the underlying hash corresponds to the serialized order, and the bytes are reversed for display by the client.

Please let me know if this is still confusing.

Thank you!

doc/elements-tx-format.md
```diff
@@ -49,13 +49,15 @@ Refer to the examples section below for more concrete examples of serialized vec
...
 | Output Index | Yes | 4 bytes | `uint32_t` | Little-endian | **Input is a coinbase**: `0xffffffff`<br><br>The two most significant bits are reserved for flags.<br><br>**Input is a pegin:** second most significant bit is 1.<br><br>**Input has an asset issuance:** most significant bit is 1. |
 | ScriptSig Length | Yes | Varies | `VarInt` | | Set to `0x00` if the transaction is SegWit and the witness contains the signature. |
 | ScriptSig | If ScriptSig Length is non-zero | Varies | `hex` | Big-endian | |
 | Sequence | Yes | 4 bytes | `uint32_t` | Little-endian | |
 | Asset Issuance | Only if the transaction has an issuance (as indicated by the Output Index) | Varies | `AssetIssuance` |  | |
 
+[^1]: The hex encodings of hashes by the Elements client (TXID, asset ID) are reversed, and so they will need to be re-reversed to match the serialized data. This is the same situation as in Bitcoin. For example, the hash `0x1123...7899` would be displayed by the Bitcoin and Elements clients as `9987...3211`. This is primarily for historical reasons: the Bitcoin client has always interpreted and displayed hashes as little-endian integers and parsed their bytes in reverse order.
```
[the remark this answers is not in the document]
jamesdorfman: Thanks. Done.

It now reads:

> The hex encodings of hashes by the Elements client (TXID, asset ID) are byte-reversed, and so the bytes will need to be re-reversed to match the serialized data. This is the same situation as in Bitcoin. For example, the hash `1123...7899` would be displayed by the Bitcoin and Elements clients as `9987...3211`. This is primarily for historical reasons: the Bitcoin client has always interpreted and displayed hashes as little-endian integers and parsed their bytes in reverse order.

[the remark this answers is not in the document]
roconnor-blockstream: Slightly prefer useing `deff` instead of `7899`.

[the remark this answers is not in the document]
jamesdorfman: Done.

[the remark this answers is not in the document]
roconnor-blockstream: should read "displayed by the Bitcoin and Elements clients as `ffde...2311`".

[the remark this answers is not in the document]
jamesdorfman: Right... the bytes are being reversed, not the whole hash itself. Sorry about that. Fixed.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+Refer to the examples section below for more concrete examples of serialized vectors.
+
+#### TxIn
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Output Hash (TXID) | Yes | 32 bytes | `hex` | [^1] | |
+| Output Index | Yes | 4 bytes | `uint32_t` | Little-endian | **Input is a coinbase**: `0xffffffff`<br><br>The two most significant bits are reserved for flags.<br><br>**Input is a pegin:** second most significant bit is 1.<br><br>**Input has an asset issuance:** most significant bit is 1. |
+| ScriptSig Length | Yes | Varies | `VarInt` | | Set to `0x00` if the transaction is SegWit and the witness contains the signature. |
+| ScriptSig | If ScriptSig Length is non-zero | Varies | `hex` | Big-endian | |
```
roconnor-blockstream, line 55: I would drop the "Big-endian" for scripts and leave the field blank.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+
+The Surjection Proof must be empty if the output’s asset ID is explicit.
+
+#### AssetIssuance
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Asset Blinding Nonce | Yes | 32 bytes | `hex` | | Zero for a new asset issuance; otherwise a blinding factor for the input. |
+| Asset Entropy | Yes | 32 bytes | `hex` | | **New issuances:** Freeform entropy field, no consensus-defined meaning, but is used as additional entropy to the asset tag calculation.<br><br>**Reissuances:** Required to be the asset's entropy value (from its initial issuance). |
+| Amount | Yes | 1 or 9 or 33 bytes | `ConfidentialAmount` | | Amount of the asset to issue. Both explicit and blinded amounts are supported. |
```
roconnor-blockstream, line 112: Note: Amount and Num Inflation Keys cannot be explicitly 0.  (Instead they should be NULL values).

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+#### TxOut
+
+|Field|Required|Size|Data Type|Encoding|Notes|
+|-----|--------|-----|---------|--------|-----|
+| Asset | Yes | 33 bytes | `ConfidentialAsset` | | Cannot be null. |
+| Amount | Yes | 9 or 33 bytes | `ConfidentialAmount` | | Cannot be null. |
+| Nonce | Yes | 1 or 33 bytes | `ConfidentialNonce` | | |
+| ScriptPubKey Length | Yes | Varies | `VarInt` | | | |
+| ScriptPubKey | If ScriptPubKey Length is non-zero | Varies | `hex` | Big-endian | |
+
```
roconnor-blockstream, line 70: Do we want a section on the definition and requirements of "UNSPENDABLE" outputs?

apoelstra: I would suggest not, it seems out of scope.

roconnor-blockstream: There are restrictions on outputs with explicit 0 / rangeproofs with 0 and the UNSPENDABLE pubkey condition.

apoelstra: Right, but I consider those similar to "there are restrictions on inputs which have bad signatures" ... i.e.this is a validation concern, not a parsing one.

roconnor-blockstream: I disagree about it not being a parsing issue, but I'll accept it.

apoelstra: I have two arguments for it being a validation issue:

* In order to check this you'd have to call into secp-zkp to extract parameters from the rangeproof itself, which otherwise parsers can simply treat as a byte-vector. So if this were a parsing issue then all parsers would need secp-zkp and to "understand the crypto" when currently they don't
* Post-bulletproofs the minimum value will be implicit, so it definitely won't be a parser concern then

roconnor-blockstream: That is a pretty strong argument.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+This document assumes some familiarity with Bitcoin and Elements (UTXOs, [Script](https://en.bitcoin.it/wiki/Script), assets, pegins, etc.). For more information on those, please refer to the [Bitcoin Wiki](https://en.bitcoin.it/wiki/Main_Page) and the [Elements Code Tutorial](https://elementsproject.org/elements-code-tutorial/overview).
+
+### Data Types
+
+*Notes*:
+* Fields in the following table are listed in the same order in which they are serialized.
+
+* The *Encoding* column in the following tables indicates how the fields are serialized.
+
+* Values are defined in multiples of 100,000,000 (this corresponds directly to how Bitcoin encodes value, where 1 BTC is divisible into 100,000,000 Satoshis). For example, a `ConfidentialAmount` representing 7 units of something (`L-BTC`, inflation keys, etc.) would be encoded as 700,000,000.
```
roconnor-blockstream, line 14: @apoelstra correct me if I'm wrong, but how the element/green UI handles units is ... more complicated than this suggests.  There is something about the asset registry etc. etc.

I think we should simply note that all unit values are in satoshis, or say all values are in the smallest indivisible unit (i.e. satoshis) or something like that.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+| Value | If header byte is not `0x00` | 8 or 32 bytes | `hex` | Big-endian | |
+
+#### ConfidentialNonce
+
+|Field|Required|Size|Data Type|Encoding|Notes| 
+|-----|--------|-----|---------|--------|-----|
+| Header | Yes | 1 byte | | | A header byte of `0x00` indicates a “null” value with no subsequent bytes.<br><br>A header byte of `0x01` indicates an “explicit” value with the following 32 bytes denoting a value (big-endian).<br><br>A header byte of `0x02` or `0x03` indicates a compressed elliptic curve point. With the least significant bit of the header byte denoting the least significant bit of the y-coordinate, and the remaining 32 bytes denoting the x-coordinate (big-endian). This point is not required to be on the curve. |
```
roconnor-blockstream, line 133: @apoelstra Is there any existing wallet interpretation of an explicit nonce?  Care to comment on its endianness?

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+
+Deserialization:
+```
+02000000 ............................. Version
+
+01 ................................... Flags (0x01 → witness data is present)
+|
+|                                      Input #1
+| 3d0bfb6e61989331d67dac3b6adf067a
+| fda20f504cabb1a7364853a68a0bb6df ... Outpoint TXID
```
roconnor-blockstream, line 473: I suggest commenting that the previous TXID is `dfb6...`

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+| | .................................. ScriptSig (empty: segwit transaction)
+|
+| ffffffff ........................... Sequence number: UINT32_MAX
+
+02 ................................... Num outputs
+|
+|                                      Output #1
+| 01 ................................. Asset header (0x01 → explicit, unblinded value)
+| 6d521c38ec1ea15734ae22b7c4606441
+| 2829c0d0579f0a713d1c04ede979026f ... Key used to generate asset ID
```
roconnor-blockstream, line 486: I suggest commenting that the assetID is `6f02...` here and in the other examples.

doc/elements-tx-format.md
```diff
@@ -0,0 +1,562 @@
...
+| | | 907e511e99bf317059264f89a56c
+| | | 50c56fefa82583b8e941b865facc
+| | | 01 ............................. Stack item #1
+| | 21 ............................... Stack item #2 length (0x21 = 33 bytes)
+| | | 02e776bdb5d8fc14d24b330c0efb
+| | | f34227cbba1cf25eb0778aa45f8b
+| | | 7cb3495046 ..................... Stack item #2
+| 06 ................................. Pegin witness stack length
+| | 08 ............................... Stack item #1 length  
+| | | e8092b0000000000 ............... Stack item #1
```
roconnor-blockstream, line 531: @apoelstra Should be be describing the pegin structure here, amount, claim_script, genesis hash, peg-in transaction?  I know that Allen's tool does.

doc/elements-tx-format.md
apoelstra: No, and lol. It's hard to imagine an interpretation of "endianness" for nonces.

doc/elements-tx-format.md
apoelstra: Agreed.

Even if other assets didn't have "natural units" that differed from 10^8 I don't know what 10^8 would have to do with the encoding, so we should drop any mention of it.

doc/elements-tx-format.md
apoelstra: Yeah, I think that we should. I'm tempted to say "it's a stack of byte vectors" but in this case every field has a clear meaning and fixed format, and it's simple enough to parse.

doc/elements-tx-format.md
apoelstra: Marking this resolved to try to convince github not to hide other, active discussions.

doc/elements-tx-format.md
roconnor-blockstream: If you would take the explicit value as an secp256k1 scalar value and multiply the generator with it, that would give it an endianness.  But I wasn't sure if a wallet is expected to do anything with an explicit nonce.

apoelstra: Yeah, that would make sense. But no, in practice nobody interprets "explicit nonces" at all.

doc/elements-tx-format.md
jamesdorfman: The reason I mentioned this was because `Example 2` below contains an asset issuance with the number of inflation keys encoded as:
```
0x29b92700 = 700 000 000 
```

Does this correspond to 7 keys? A single inflation key is indivisible, correct?

apoelstra: I think that's 700 million reissuance keys.

jamesdorfman: I see. So when I parse the full issuance transaction with `$elements-cli decoderawtransaction <...>`, the result contains:
```
"issuance": {
        "assetBlindingNonce": "0000000000000000000000000000000000000000000000000000000000000000",
        "assetEntropy": "503f1db03b8ed61ace8fbadc2fcb6c200bdf1ada403d74b891e55e1a047b6596",
        "isreissuance": false,
        "token": "6e88fa86a097b9ede564f67c60e8c5d24ed1f10eeb56c3a60223961f2e38737f",
        "asset": "d42c5ba3724b7f7a045968e0c04e69f81070aca8b6e2db88cd84c65dd07b8324",
        "assetamount": 33.00000000,
        "tokenamount": 7.00000000
}
```
      
Is the `"tokenamount": 7.00000000` indicating that there are really 700,000,000 reissuance tokens?

apoelstra: lol. yes.

jamesdorfman: Ok that makes sense now. Thanks!

apoelstra: I opened https://github.com/ElementsProject/elements/issues/1152 about this.

roconnor-blockstream (approved):
Approved, though I think we might as well inline the definition of the encoding for VarInts.

philippem:
ack thank you!!

philippem (approved):
great job, thank you!

PR #1142, opened 2022-08-14
A couple backports to fix compilation

apoelstra:
I updated gcc and boost on my system so of course Elements stopped compiling. Backport the fixes from upstream when they encountered this.

PR #1150, opened 2022-08-24
Changing PSET to not blind asset issuances by default.

allenpiscitello:
PSET does not allow a user to specify if an issuance is to be blinded or not. The previous default was only allowing blinded, and it would blind the issuances if they were unblinded.

We should modify the PSET spec to allow specifying blinding or unblinded issuances, and in the meantime honor whatever is set in the PSET. If a commitment is set, use that, otherwise we should leave it unblinded.

Fixing this also uncovered some other minor bugs that were in previous fixes to PSET-related code.

### DIFF
--- a/src/blindpsbt.cpp
+++ b/src/blindpsbt.cpp
@@ -387,55 +387,53 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
 
         // Handle issuances
         if (input.m_issuance_value != std::nullopt || input.m_issuance_value_commitment.IsCommitment() || input.m_issuance_inflation_keys_amount != std::nullopt || input.m_issuance_inflation_keys_commitment.IsCommitment()) {
-            if (!input.m_issuance_value_commitment.IsCommitment() && input.m_issuance_rangeproof.size() == 0 && input.m_issuance_inflation_keys_rangeproof.size() == 0) {
-                CAsset issuance_asset;
-                CAsset reissuance_asset;
+            CAsset issuance_asset;
+            CAsset reissuance_asset;
 
-                uint256 entropy;
-                if (!input.m_issuance_blinding_nonce.IsNull()) {
-                    // Reissuance, use assetEntropy as the asset entropy
-                    entropy = input.m_issuance_asset_entropy;
-                } else {
-                    // New issuance, make new entropy
-                    GenerateAssetEntropy(entropy, input.GetOutPoint(), input.m_issuance_asset_entropy);
+            uint256 entropy;
+            if (!input.m_issuance_blinding_nonce.IsNull()) {
+                // Reissuance, use assetEntropy as the asset entropy
+                entropy = input.m_issuance_asset_entropy;
+            } else {
+                // New issuance, make new entropy
+                GenerateAssetEntropy(entropy, input.GetOutPoint(), input.m_issuance_asset_entropy);
+            }
+
+            if (input.m_issuance_value != std::nullopt || input.m_issuance_value_commitment.IsCommitment()) {
+                // Asset isn't blinded yet. Add it to the list of input assets
+                CalculateAsset(issuance_asset, entropy);
+                fixed_input_tags.emplace_back();
+                memcpy(fixed_input_tags.back().data, issuance_asset.begin(), 32);
+                ephemeral_input_tags.emplace_back();
+                if (input.m_issuance_value_commitment.IsNull()) {
+                    if (secp256k1_generator_generate(secp256k1_blind_context, &ephemeral_input_tags.back(), issuance_asset.begin()) != 1) {
+                        return BlindingStatus::INVALID_ASSET;
+                    }
                 }
-
-                if (input.m_issuance_value != std::nullopt || input.m_issuance_value_commitment.IsCommitment()) {
-                    // Asset isn't blinded yet. Add it to the list of input assets
-                    CalculateAsset(issuance_asset, entropy);
-                    fixed_input_tags.emplace_back();
-                    memcpy(fixed_input_tags.back().data, issuance_asset.begin(), 32);
-                    ephemeral_input_tags.emplace_back();
-                    if (input.m_issuance_value_commitment.IsNull()) {
-                        if (secp256k1_generator_generate(secp256k1_blind_context, &ephemeral_input_tags.back(), issuance_asset.begin()) != 1) {
-                            return BlindingStatus::INVALID_ASSET;
-                        }
-                    }
-                    else {
-                        memcpy(ephemeral_input_tags.back().data, input.m_issuance_value_commitment.vchCommitment.data(), 33);
-                    }
-                    input_asset_blinders.emplace_back();
+                else {
+                    memcpy(ephemeral_input_tags.back().data, input.m_issuance_value_commitment.vchCommitment.data(), 33);
                 }
+                input_asset_blinders.emplace_back();
+            }
 
-                bool blind_issuance = input.m_issuance_inflation_keys_commitment.IsCommitment() || input.m_issuance_value_commitment.IsCommitment();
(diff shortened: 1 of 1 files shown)


src/blindpsbt.cpp
```diff
@@ -386,7 +386,7 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
         }
 
         // Handle issuances
-        if (input.m_issuance_value) {
+        if (input.m_issuance_value != std::nullopt || input.m_issuance_value_commitment.IsCommitment() || input.m_issuance_inflation_keys_amount != std::nullopt || input.m_issuance_inflation_keys_commitment.IsCommitment()) {
```
allenpiscitello, line 389: The previous code missed cases where there was a re-issuance token but not an issuance. This is allowed, and should be accounted for.

src/wallet/wallet.cpp
```diff
@@ -1921,8 +1921,8 @@ BlindingStatus CWallet::WalletBlindPSBT(PartiallySignedTransaction& psbtx) const
             our_input_data[i] = std::make_tuple(amount, asset, asset_blinder, value_blinder);
         }
 
-        // Blind issuances on our inputs
-        if (input.m_issuance_value || input.m_issuance_inflation_keys_amount) {
```
allenpiscitello, line 1925: Leaving this logic in place even though blinding is removed, so that when the PSET spec is updated we can easily add blinding back.

src/blindpsbt.cpp
```diff
@@ -400,83 +400,40 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
...
-                if (secp256k1_generator_generate(secp256k1_blind_context, &ephemeral_input_tags.back(), issuance_asset.begin()) != 1) {
-                    return BlindingStatus::INVALID_ASSET;
+                if (input.m_issuance_value != std::nullopt || input.m_issuance_value_commitment.IsCommitment()) {
+                    // Asset isn't blinded yet. Add it to the list of input assets
+                    CalculateAsset(issuance_asset, entropy);
+                    fixed_input_tags.emplace_back();
+                    memcpy(fixed_input_tags.back().data, issuance_asset.begin(), 32);
+                    ephemeral_input_tags.emplace_back();
+                    if (input.m_issuance_value_commitment.IsNull()) {
+                        if (secp256k1_generator_generate(secp256k1_blind_context, &ephemeral_input_tags.back(), issuance_asset.begin()) != 1) {
```
apoelstra, line 410: I believe `&ephemeral_input_tags.back()` should be `ephemeral_input_tags.back().data`.

apoelstra: Ah, no, `ephemeral_input_tags.back()` returns a `secp256k1_generator`. The real bug here is that we ever touch `.data` which is supposed to be a secp-zkp implementation detail, but that's a fight for another day.

apoelstra (approved):
utACK 420de43c1ce3d897517249958dab0d11c320cd37

(cannot test because of #1142)

src/blindpsbt.cpp
```diff
@@ -386,7 +386,7 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
         }
 
         // Handle issuances
-        if (input.m_issuance_value) {
+        if (input.m_issuance_value != std::nullopt || input.m_issuance_value_commitment.IsCommitment() || input.m_issuance_inflation_keys_amount != std::nullopt || input.m_issuance_inflation_keys_commitment.IsCommitment()) {
             if (!input.m_issuance_value_commitment.IsCommitment() && input.m_issuance_rangeproof.size() == 0 && input.m_issuance_inflation_keys_rangeproof.size() == 0) {
```
sanket1729, line 390: This entire block would be skipped if there is `m_issuance_value_commitment` which seems wrong?

sanket1729: The code for dealing with issuances/re-issuances should be same. If we are not doing it, most like a sign of bug.

src/blindpsbt.cpp
```diff
@@ -400,83 +400,40 @@ BlindingStatus BlindPSBT(PartiallySignedTransaction& psbt, std::map<uint32_t, st
...
+                    }
+                    input_asset_blinders.emplace_back();
                 }
-                unsigned int iss_to_blind = 1; // Always do the first issuance blinding iteration for the issuance value
 
-                bool blind_issuance = our_issuances_to_blind.count(i) > 0;
+                bool blind_issuance = input.m_issuance_inflation_keys_commitment.IsCommitment() || input.m_issuance_value_commitment.IsCommitment();
 
-                if (input.m_issuance_blinding_nonce.IsNull() && input.m_issuance_inflation_keys_amount) {
+                if (input.m_issuance_blinding_nonce.IsNull() && (input.m_issuance_inflation_keys_amount != std::nullopt || input.m_issuance_inflation_keys_commitment.IsCommitment())) {
```
sanket1729, line 422: Where are we dealing with the case `m_issuance_blinding_nonce` is not Null? Seems like this would not work correctly in re-issuance case?

sanket1729 (commented):
Still not completely convinced that this is correct.

src/blindpsbt.cpp
allenpiscitello: I inspected the transaction it converted to a PSBT. The output for the reissuance token was that of the blinded reissuance token ID. This will cause a surjection proof failure because there is no input for the blinded reissuance token, only the unblinded one, when calling the PSBT version of this.

src/blindpsbt.cpp
allenpiscitello: In a reissuance case, there are no reissuance tokens being created, only the asset tokens. Those are handed in the section above.

allenpiscitello:
> @apoelstra , @allenpiscitello. I think the following fixups would fix this. [sanket1729@6182dda](https://github.com/sanket1729/elements/commit/6182dda8a1d3307b91df7c1a45ad0a5e16e637e6). Feel free to cherry-pick/squash as you see fit and if my reasoning is correct :P
> 
> 1. The bool for checking whether the re-issuance id calculation should be blinded is based only asset amount and not on issuance amount

I believe this is incorrect as stated above.

.
> 2. Cleanly deals with cases where only one of issuance/re-issuance is blinded. The previous logic had some confusing checks on `m_issuance_value_commitment`.

The intended behavior with this modification is to just create a transaction that would not be valid if you only specified one or the other as blinded. This transaction would be invalid if you mixed and matched it.

> 3. I think the previous case had a check on BlindingNonce being Null which could have missed the considering the confidential re-issuance outputs.
> 

I believe this is wrong based on the comment above.

> This should be easy to review if you directly compare it with validation logic in confidential_validation.cpp. Sorry if the code does not build, I do not have the machine to test building right now

sanket1729 (approved):
ACK dd2d758bc6a205348dcb1f2d4fd87c05160cae35. Thanks for putting up with the iterations. I am happy with the present logic

src/blindpsbt.cpp
jgriffiths: For complex/long-winded logic checks like this, it would be nice to have methods on the input such as:

```
bool PSBTInput::HasIssuance() const
{
    if (m_issuance_value != std::nullopt || m_issuance_value_commitment.IsCommitment()) {
        return true; // Blinded or unblinded issuance
    }
    if (m_issuance_inflation_keys_amount != std::nullopt || m_issuance_inflation_keys_commitment.IsCommitment()) {
         return true; // Re-issuance token without an issuance
    }
    return false; // No issuance or re-issuance present
}
```

This would make the underlying logic much clearer IMO.

apoelstra (approved):
utACK dd2d758bc6a205348dcb1f2d4fd87c05160cae35

PR #1155, opened 2022-09-01
rpc: change getnewblockhex to take multiple commitments

delta1:
Modifies the `getnewblockhex` json rpc call to accept an "array" of commitments instead of a single commitment. The given string parameter will be split on spaces. 

Adds a new functional test for this rpc call.

edit: no longer a breaking change

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -1277,7 +1277,7 @@ static RPCHelpMan getnewblockhex()
                 "\nGets hex representation of a proposed, unmined new block\n",
                 {
                     {"min_tx_age", RPCArg::Type::NUM, RPCArg::Default{0}, "How many seconds a transaction must have been in the mempool to be included in the block proposal. This may help with faster block convergence among functionaries using compact blocks."},
-                    {"proposed_parameters", RPCArg::Type::OBJ, RPCArg::Optional::OMITTED , "Parameters to be used in dynamic federations blocks as proposals. During a period of `-dynamic_epoch_length` blocks, 4/5 of total blocks must signal these parameters for the proposal to become activated in the next epoch.",
+                    {"proposed_parameters", RPCArg::Type::OBJ, RPCArg::Optional::OMITTED, "Parameters to be used in dynamic federations blocks as proposals. During a period of `-dynamic_epoch_length` blocks, 4/5 of total blocks must signal these parameters for the proposal to become activated in the next epoch.",
                         {
                             {"signblockscript", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "Hex-encoded block signing script to propose"},
                             {"max_block_witness", RPCArg::Type::NUM, RPCArg::Optional::NO, "Total size in witness bytes that are allowed in the dynamic federations block witness for blocksigning"},
@@ -1289,7 +1289,11 @@ static RPCHelpMan getnewblockhex()
                             },
                         },
                         "proposed_parameters"},
-                        {"commit_data", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED_NAMED_ARG, "Data in hex to be committed to in an additional coinbase output."},
+                    {"commit_data", RPCArg::Type::ARR, RPCArg::Optional::OMITTED, "Array of data in hex to be committed to in additional coinbase outputs.",
+                        {
+                            {"", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "Hex encoded string for commit data"},
+                        },
+                    },
                 },
                 RPCResult{
                     RPCResult::Type::STR_HEX, "blockhex", "the block hex",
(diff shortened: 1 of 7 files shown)


apoelstra:
Concept ACK -- heads up you can make this a non-breaking change by explicitly accepting a string and just converting it to an array internally.

delta1:
@apoelstra thanks, yeah good point is that preferable?

apoelstra:
Yes, it's much preferable to keep existing RPC scripts working.

delta1:
Okay sure will do

src/rpc/mining.cpp
```diff
@@ -1289,7 +1291,7 @@ static RPCHelpMan getnewblockhex()
                             },
                         },
                         "proposed_parameters"},
-                        {"commit_data", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED_NAMED_ARG, "Data in hex to be committed to in an additional coinbase output."},
+                    {"commit_data", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED_NAMED_ARG, "Array of additional hex data to be committed to in additional coinbase outputs. Separate each hex commitment with one or more spaces."}
```
stevenroose, line 1294: Not sure I like the implicit "multiple pushes" when you push mutliple hexes with spaces.

If I would pass in `"deadbeef deadb00f"`, I would intuitively expect the result to be a single nulldata output with multiple pushed like `OP_RETURN <deadbeef> <deadb00f>` instead of multiple OP_RETURN outputs..

Also did you check if `RPCArg::Type::STR_HEX` actually allows spaces? The doc says
```
        STR_HEX,       //!< Special type that is a STR with only hex chars
```

apoelstra: There are two issues here -- one is that if you want to specify multiple hex strings you should use an array rather than spaces (which are not used anywhere else and also are likely to be confusing when calling from sh). The other is whether we should have multiple outputs or multiple pushes in the one output.

Personally I think either is fine as long as it's documented.

delta1: > Also did you check if RPCArg::Type::STR_HEX actually allows spaces?  

The functional test exercises this, the spaces are accepted

> if you want to specify multiple hex strings you should use an array rather than spaces 

This was the breaking change previously? Or is it that you want a different separator? 

> Personally I think either is fine as long as it's documented 

I agree, I'm not fussed either way please let me know what you both prefer.

apoelstra: > This was the breaking change previously? Or is it that you want a different separator?
>

Try to parse as both and do the right thing based on the given type. This is done in a few other RPCs as well (or at least used to be).

delta1: Done

src/rpc/mining.cpp
```diff
@@ -1346,18 +1350,33 @@ static RPCHelpMan getnewblockhex()
...
+    std::set<CScript> data_commitments;
     if (!request.params[2].isNull()) {
-        std::vector<unsigned char> data_bytes = ParseHex(request.params[2].get_str());
-        data_commitment = CScript() << OP_RETURN << data_bytes;
+        UniValue commitments(UniValue::VARR);
+        try {
+            // backwards compatibility: attempt to parse as a string first
+            UniValue hex = request.params[2].get_str();
+            commitments.push_back(hex);
+        } catch (const std::runtime_error& e) {
```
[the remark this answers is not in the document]
delta1: I think I did see `is_str()` on `UniValue` let me check that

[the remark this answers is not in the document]
stevenroose: I looked in univalue.cpp and the subfiles and I couldn't find it. I also thought I remembered it existed. Maybe it's in some header file that's in another folder?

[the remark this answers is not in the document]
delta1: Good catch, thanks! Updated.

stevenroose:
utACK 110087764022671d699891592e74fe89a4adae37

src/rpc/mining.cpp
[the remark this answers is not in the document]
delta1: Yes it's in `univalue.h` 

https://github.com/ElementsProject/elements/blob/master/src/univalue/include/univalue.h#L81

src/rpc/mining.cpp
```diff
@@ -1346,18 +1350,34 @@ static RPCHelpMan getnewblockhex()
...
+            commitments = request.params[2].get_array();
+        }
+
+        for (unsigned int i = 0; i < commitments.size(); i++) {
+            std::vector<unsigned char> data_bytes = ParseHex(commitments[i].get_str());
+            CScript data_commitment = CScript() << OP_RETURN << data_bytes;
+            auto res = data_commitments.insert(data_commitment);
+            if (!res.second) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Duplicate commitment");
+            }
```
apoelstra, line 1373: In 110087764022671d699891592e74fe89a4adae37:

Also not going to hold up the PR for this, but noting in case people run into this, so they'll see that we at least thought about it. Why do we have this restriction?

delta1: I wasn’t sure it made sense to allow duplicate commitments in the same call?

.gitignore
```diff
@@ -151,3 +151,5 @@ dist/
 *.background.tiff
 
 /guix-build-*
+
+.vscode
```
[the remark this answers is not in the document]
delta1: Okay will remove

delta1:
> ./test/functional/rpc_help.py fails now

sorry will take a look

apoelstra:
```
19:11:17 -bash@camus ~/code/ElementsProject/elements/pr-review e08a86df0e5$ ./test/functional/feature_blocksign.py 
2022-09-06T19:11:19.104000Z TestFramework (INFO): Initializing test directory /tmp/bitcoin_func_test_3ghet7ze
2022-09-06T19:11:19.471000Z TestFramework (INFO): Mining and signing 101 blocks to unlock funds
2022-09-06T19:11:19.476000Z TestFramework (ERROR): Assertion failed
Traceback (most recent call last):
  File "/store/home/apoelstra/code/ElementsProject/elements/pr-review/test/functional/test_framework/test_framework.py", line 128, in main
    self.run_test()
  File "/store/home/apoelstra/code/ElementsProject/elements/pr-review/./test/functional/feature_blocksign.py", line 189, in run_test
    self.mine_blocks(101, False)
  File "/store/home/apoelstra/code/ElementsProject/elements/pr-review/./test/functional/feature_blocksign.py", line 178, in mine_blocks
    self.mine_block(transactions)
  File "/store/home/apoelstra/code/ElementsProject/elements/pr-review/./test/functional/feature_blocksign.py", line 128, in mine_block
    assert_equal(CScript(dummy_struct.vtx[0].vout[0].scriptPubKey).hex(), '6a04deadbeef')
  File "/store/home/apoelstra/code/ElementsProject/elements/pr-review/test/functional/test_framework/util.py", line 62, in assert_equal
    raise AssertionError("not(%s)" % " == ".join(str(arg) for arg in (thing1, thing2) + args))
AssertionError: not(51 == 6a04deadbeef)
2022-09-06T19:11:19.527000Z TestFramework (INFO): Stopping nodes
2022-09-06T19:11:19.583000Z TestFramework (WARNING): Not cleaning up dir /tmp/bitcoin_func_test_3ghet7ze
2022-09-06T19:11:19.583000Z TestFramework (ERROR): Test failed. Test logging available at /tmp/bitcoin_func_test_3ghet7ze/test_framework.log
2022-09-06T19:11:19.583000Z TestFramework (ERROR): 
2022-09-06T19:11:19.583000Z TestFramework (ERROR): Hint: Call /store/home/apoelstra/code/ElementsProject/elements/pr-review/test/functional/combine_logs.py '/tmp/bitcoin_func_test_3ghet7ze' to consolidate all logs
2022-09-06T19:11:19.583000Z TestFramework (ERROR): 
2022-09-06T19:11:19.583000Z TestFramework (ERROR): If this failure happened unexpectedly or intermittently, please file a bug and provide a link or upload of the combined log.
2022-09-06T19:11:19.583000Z TestFramework (ERROR): https://github.com/ElementsProject/elements/issues
2022-09-06T19:11:19.583000Z TestFramework (ERROR): 
```

delta1:
@apoelstra all fixed now

apoelstra (approved):
ACK 5d10ff9f52d5bf372325390c1cda880b3dc7b9dc

PR #1157, opened 2022-09-01
pset: Add input explicit value, assets, and proofs, and issuance blinding flag

### LINKED ISSUE #1153: add unblinded value and asset input fields to PSET
Apps running on highly constrained hardware wallets wish to display unblinded value and asset inputs to the user, when confirming transactions specified by a PSET. These same devices are unable to perform unblinding due to hardware and software constraints.

It has been requested that unblinded value and asset input fields be added to the PSET spec, so that the companion app (which runs on the host) may provide this information to the device, along with the proofs and commitments. The device would perform verification independently, before displaying the unblinded values.

**Proposed Solution**

New elements per-input fields for unblinded value and asset, e.g.

```
PSBT_ELEMENTS_IN_VALUE
PSBT_ELEMENTS_IN_ASSET
```

achow101:
Adds fields for explicit input value and assets, and their proofs. Also adds a flag to indicate whether an issuance should be blinded.

Closes #1153

### DIFF
--- a/doc/pset.mediawiki
+++ b/doc/pset.mediawiki
@@ -295,6 +295,16 @@ The currently defined elements per-input proprietary types are as follows:
 |
 | 0
 | 2
+|-
+| Blinded Issuance Flag
+| <tt>PSBT_ELEMENTS_IN_BLINDED_ISSUANCE = 0x15</tt>
+| None
+| No key data
+| <tt><1 byte boolean></tt>
+| A boolean flag. <tt>0x00</tt> indicates the issuance should not be blinded, <tt>0x01</tt> indicates it should be. If not specified, assumed to be <tt>0x01</tt>. Note that this does not indicate actual blinding status, but rather the expected blinding status prior to signing.
+|
+| 0
+| 2
 |}
 
 The currently defined elements per-output proprietary types are as follows:
--- a/doc/pset.mediawiki
+++ b/doc/pset.mediawiki
@@ -255,6 +255,46 @@ The currently defined elements per-input proprietary types are as follows:
 |
 | 0
 | 2
+|-
+| Explicit Value
+| <tt>PSBT_ELEMENTS_IN_EXPLICIT_VALUE = 0x11</tt>
+| None
+| No key data
+| <tt><64-bit little endian int value></tt>
+| The explicit value for the input being spent. If provided, <tt>PSBT_ELEMENTS_IN_VALUE_PROOF</tt> must be provided too. Must not be provided if the input's value in the UTXO is already explicit.
+|
+| 0
+| 2
+|-
+| Explicit Value Proof
+| <tt>PSBT_ELEMENTS_IN_VALUE_PROOF = 0x12</tt>
+| None
+| No key data
+| <tt><rangeproof></tt>
+| An explicit value rangeproof that proves that the value commitment in this input's UTXO matches the explicit value in <tt>PSBT_ELEMENTS_IN_EXPLICIT_VALUE</tt>. If provided, <tt>PSBT_ELEMENTS_IN_EXPLICIT_VALUE</tt> must be provided too.
+|
+| 0
+| 2
+|-
+| Explicit Asset
+| <tt>PSBT_ELEMENTS_IN_EXPLICIT_ASSET = 0x13</tt>
+| None
+| No key data
+| <tt><32 byte asset tag></tt>
+| The explicit asset for the input being spent. If provided, <tt>PSBT_ELEMENTS_IN_ASSET_PROOF</tt> must be provided too. Must not be provided if the input's asset in the UTXO is already explicit.
+|
+| 0
+| 2
+|-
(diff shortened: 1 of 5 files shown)


doc/pset.mediawiki
```diff
@@ -255,6 +255,56 @@ The currently defined elements per-input proprietary types are as follows:
...
+| None
+| No key data
+| <tt><proof></tt>
+| An asset surjection proof with this input's asset as the only asset in the input set in order to prove that the asset commitment in the UTXO matches the explicit asset in <tt>PSBT_ELEMENTS_IN_EXPLICIT_ASSET</tt>. If provided, <tt>PSBT_ELEMENTS_IN_EXPLICIT_ASSET</tt> must be provided too.
+|
+| 0
+| 2
+|-
+| Blinded Issuance Flag
+| <tt>PSBT_ELEMENTS_IN_BLINDED_ISSUANCE = 0x15</tt>
```
philippem, line 300: Should this be `PSBT_ELEMENTS_IN_ISSUANCE_NEEDS_BLINDED = 0x15` ?

doc/pset.mediawiki
```diff
@@ -255,6 +255,46 @@ The currently defined elements per-input proprietary types are as follows:
 |
 | 0
 | 2
+|-
+| Explicit Value
+| <tt>PSBT_ELEMENTS_IN_EXPLICIT_VALUE = 0x11</tt>
+| None
+| No key data
+| <tt><64-bit little endian int value></tt>
+| The explicit value for the input being spent. If provided, <tt>PSBT_ELEMENTS_IN_VALUE_PROOF</tt> must be provided too.
```
apoelstra, line 264: In d1d7af3442059fd0b7a632ac0c23dbc2eafa94e3:

What if the input value is already explicit? Should we just forbid this field? I guess so.

src/blindpsbt.cpp
```diff
@@ -220,6 +220,31 @@ BlindProofResult VerifyBlindProofs(const PSBTOutput& o) {
...
+BlindProofResult VerifyBlindProofs(const PSBTInput& i) {
+    CTxOut utxo;
+    if (!i.GetUTXO(utxo)) {
+        return BlindProofResult::OK;
+    }
+
+    if (i.m_explicit_value != std::nullopt) {
+        if (i.m_value_proof.empty()) {
+            return BlindProofResult::MISSING_VALUE_PROOF;
+        } else if (!VerifyBlindValueProof(*i.m_explicit_value, utxo.nValue, i.m_value_proof, utxo.nAsset)) {
```
apoelstra, line 232: In ceee8a6abafe368775b5c71a38d3973691f134bb:

Need to check that `utxo.nValue` is confidential, otherwise we potentially wind up passing uninitialized memory to the libsecp functions. If it's not confidential it's fine to just return an error.

doc/pset.mediawiki
apoelstra: (In 67d3a95268abf96ae463865494a70e45237df69d:)

No, I think the current name is clearer. "Needs blinded" implies that it's not currently blinded, but eventually it will be and this field should still be set.

apoelstra:
Done reviewing 902b7363e3319e2ab6b5e7a8555312f6c95f89ee:

Mostly just nits, except the `nUtxo` `IsConfidential` thing which needs to be fixed. This is a super well-organized PR, thank you!!

doc/pset.mediawiki
philippem: The declaration in psbt.h is `PSBT_ELEMENTS_IN_ISSUANCE_NEEDS_BLINDED = 0x15` at`https://github.com/ElementsProject/elements/pull/1157/files#diff-2fb519d85c6ae99cd87c94e5b7cba43ff2ddd031b2c523dd2ffb12e991c16494R80`
but the constant here is `PSBT_ELEMENTS_IN_BLINDED_ISSUANCE`.

apoelstra: Oh, in that case psbt.h needs to be updated in ec1c6dc962f3aa8ebb81ff99a4d5ec3327215eed

doc/pset.mediawiki
achow101: Same for the asset?

doc/pset.mediawiki
achow101: Made the naming consistent to `PSBT_ELEMENTS_IN_BLINDED_ISSUANCE`

doc/pset.mediawiki
achow101: Added a sentence saying it cannot be provided when the input already has an explicit value. Did the same for the asset field.

src/blindpsbt.cpp
achow101: Done. Also added the same check below for `nAsset`.

apoelstra (approved):
ACK 9ecbf38050ecb67939a111bf3e3cef43aee49720

PR #1162, opened 2022-09-06
Add more CPUs to the fuzzer test

psgreco:
Currently fuzzer test fails due to the 120 minute timeout, add some CPUs to avoid that


apoelstra (approved):
utACK c66363bdc6fbe387f6dea78c2cecb71f97245eac

PR #1165, opened 2022-09-06
pset: Tests and bug fixes for input explicit value and asset fields

achow101:
Adds some tests for the explicit value, value proof, explicit asset, and asset proof PSET input fields. Also fixes a few bugs that these caught.


apoelstra (approved):
ACK 66290e60cfe797c81e5c90eca36414cdcd27c094

PR #1167, opened 2022-09-09
Add native support for liquidtestnet

psgreco:
Currently using liquidtestnet requires a big config file, which is error-prone and uncomfortable for users, this MR adds support for just using `-chain=liquidtestnet`.

There are a few things that need to be discussed/addressed

1. Main port (currently set to 18891 because that's the one used originally, and the one that's configured in `liquidtestnet.com`)
1. RPC and extra ports, I just chose some non-overlapping ports that are "close" to the ones in the other liquid/elements networks.
1. Icon color, I just used something different than `liquidv1` and `elementsregtest`, suggestions accepted :wink: 
1. `liquid-testnet.blockstream.com` is still using port 18892, it will be changed to 18891 soon.
1. both seeds are just CNAME dns records at the moment, they should be changed to real seed nodes (no changes in code required, but important for review)


delta1:
> adds support for just using -chain=elementsregtest 

should this be `-chain=liquidtestnet`  ?

psgreco:
> > adds support for just using -chain=elementsregtest
> 
> should this be `-chain=liquidtestnet` ?

Yes, fixed

delta1:
ACK c6e074e 

Tested by compiling the codebase, started `elementsd -chain=liquidtestnet` sync, and started `elements-qt -chain=liquidtestnet` sync successfully

delta1 (approved):

wintercooled:
ACK https://github.com/ElementsProject/elements/commit/c6e074e25b48b491120a5bad2e0ba1998d8b0dc7

Compiled, installed and ran `elements-qt` with `-chain=liquidtestnet` and could sync to tip.

Note: I only tested running without `-chain=liquidtestnet` briefly - did not wait to sync live to tip.

gwillen:
utACK. My main concern is making sure there is absolutely no accidental behavior change for existing users of custom chains. (As far as I can tell there is not.)

apoelstra:
utACK c6e074e25b48b491120a5bad2e0ba1998d8b0dc7

PR #1171, opened 2022-09-19
Replace icons

andreabonel:
Replace Bitcoin icons with the corresponding ones from Elements.


delta1 (approved):
utACK f5399df08aa75c195ef5fdca324359f56ee5c6bb

PR #1172, opened 2022-09-19
wallet: don't clear out all the blinding data when dropping change

apoelstra:
The Elements 22 blinding logic has an edge case where when we drop change, leaving only a single blinded output, we recompute a bunch of blinding data to handle the potential for us to have 0 inputs and 1 output to blind. (BlindTransaction will fail in this case because it cannot make the transaction balance with only one output to mess with.)

In this recomputation, we dropped more data than we meant to, causing us to incorrectly blind an output.


jgriffiths:
ack 982ed6078c0bfcadccfbdf6f15e710eb642a06ab - passes our internal CI tests now, thanks

delta1:
ACK 62598099fb5074d7dbc40e6204c0ff2f50c356c6 

I ran the regression test locally: `test/functional/test_runner.py`

delta1 (approved):

apoelstra:
Holding off on merge because I think @psgreco would like to forward-port the existing elements-22 branch so that I can rebase on that.

Meanwhile @jgriffiths believes there is another related bug related to transactions that carry issuances. I'll track that down and open a new PR on top of this one.

apoelstra:
Rebased on the spend.cpp-relevant Bitcoin backports already in Elements 22, so we should be able to merge this into the 22 branch without too much trouble.

jgriffiths:
tested ack e5e3ec2700a1f72351e2d227da68a7cc07c7285a

With this PR our internal CI tests now pass.

Issue #1120, opened 2022-06-29
validatepegin should try bitcoind's default cookie path


wtogami:
Elements BTC peg needs to query a bitcoind full node via RPC to verify that all pegged BTC on the sidechain is fully backed. `validatepegin=0` turns off this requirement but we should not encourage this unsafe practice. It would simplify deployment for those running `bitcoind` and `elementsd` as the same user if `elementsd` looks for the default cookie path to connect to the local `bitcoind`.

### Fix
If `validatepegin=1` ...
- [ ] If any of the `mainchainrpc*` options are defined then use that. If it fails print an error and exit.
- [ ] NEW: Otherwise look at the default path `$HOME/.bitcoin/.cookie` and try if that exists.
- [ ] NEW: During runtime if the underlying bitcoind restarts the cookie changes. Elementsd should recognize this change and switch to the new cookie contents during runtime.

PR #1173, opened 2022-09-21
Fix location for default cookie path

andreabonel:
Fixes: https://github.com/ElementsProject/elements/issues/1120

### DIFF
--- a/src/rpc/request.cpp
+++ b/src/rpc/request.cpp
@@ -134,8 +134,7 @@ static fs::path GetMainchainAuthCookieFile()
     if (gArgs.GetChainName() == "liquidv1") {
         cookie_file = ".cookie";
     }
-    std::string path = gArgs.GetArg("-mainchainrpccookiefile", cookie_file);
-    return AbsPathForConfigVal(fs::path(path));
+    return fsbridge::AbsPathJoin(GetMainchainDefaultDataDir(), gArgs.GetArg("-mainchainrpccookiefile", cookie_file));
 }
 
 bool GetMainchainAuthCookie(std::string *cookie_out)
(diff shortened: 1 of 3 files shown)


src/rpc/request.cpp
```diff
@@ -134,8 +134,11 @@ static fs::path GetMainchainAuthCookieFile()
...
         cookie_file = ".cookie";
     }
-    std::string path = gArgs.GetArg("-mainchainrpccookiefile", cookie_file);
-    return AbsPathForConfigVal(fs::path(path));
+    std::string path = gArgs.GetArg("-mainchainrpccookiefile", "");
+    if (path == "") {
+        return fsbridge::AbsPathJoin(GetMainchainDefaultDataDir(), cookie_file);
+    }
+    return AbsPathForConfigVal(fs::path(cookie_file));
 }
```
gwillen, line 142: This doesn't work right if `-mainchainrpccookiefile` is specified, which is presumably why CI is failing.

Instead of this, I would suggest just replacing the line 

`cookie_file = ".cookie";`

Which sets the default, with 

`cookie_file = fsbridge::AbsPathJoin(GetMainchainDefaultDataDir(), ".cookie");`

So that the default is to use the cookie file in the right place, but the logic for handling user arguments with GetArg is unchanged.

gwillen, line 142: Deleted my previous comment, which was wrong. Let's try again.

This doesn't work right if `-mainchainrpccookiefile` is specified, which is presumably why CI is failing.

Instead, I would suggest something like this:

```
    if (gArgs.GetChainName() == "liquidv1") {
        cookie_file = ".cookie";
    }
    cookie_file = fsbridge::AbsPathJoin(GetMainchainDefaultDataDir(), cookie_file);
```

so that you are only changing how the default works, and leave the rest of the logic (handling user arguments with GetArg) alone.

gwillen:
utACK, thanks!

PR #1182, opened 2022-10-05
rpc: decodepsbt: fix check for blind_reissuance_amount_proof output

jgriffiths:
Also make sure value rangeproof generation fails if its asset commitment isn't provided.


apoelstra (approved):
utACK e2348dadf3488380a2123b4b8447a66311a736ca

PR #1183, opened 2022-10-11
backport to 0.21: change getnewblockhex to take multiple commitments

delta1:
Modifies the getnewblockhex json rpc call to accept an array of commitments instead of a single commitment.

Backwards compatibility is maintained by first attempting to parse as a string for a singular commitment.

Related PR: #1155


gwillen:
utACK backport (appears to match the existing change it's backporting.)

Issue #1201, opened 2023-01-27
getsidechaininfo rpc help incorrectly specifies the order of current_fedpegscripts array


philippem:
The help for the `getsidechaininfo` RPC says "Oldest first" in 
```
...                                       
  "fedpegscript" : "hex",                        (string) The fedpegscript from genesis block
  "current_fedpegscripts" : [                    (json array) The currently-enforced fedpegscripts in hex. Peg-ins for any entries on this list are honored by consensus and policy. Oldest first. Two total entries are possible
    "hex",                                       (string) active fedpegscript
    ...
  ],
  "current_fedpeg_programs" : [                  (json array) The currently-enforced fedpegscript scriptPubKeys in hex. Prior to a transition this may be P2SH scriptpubkey, otherwise it will be a native segwit script. Results are paired in-order with current_fedpegscripts
    "hex",                                       (string) active fedpegscript scriptPubKeys
    ...
  ],
````

but the array appears to be newest first.

The result of running this command in prod just after the DynaFed WM transition is (output elided)

```
 "fedpegscript": "745c87635b2102...5bb5368ae",
  "current_fedpeg_programs": [
    "0020333a4af67452308b9814ff59dc2fc3d0325fd7fdcb61d108e992b3956095f2fe",
    "a9149e10aa3d2f248e0e42f9bab31e858240e7ed40e487"
  ],
  "current_fedpegscripts": [
    "5b21020e0338c96a8....ca5bb53ae68",
    "745c87635b21020e0.......a5bb5368ae"

....

Note that the genesis block fedpegscript `745...` is listed second.
```

`getblockchaininfo` reports

```
{
  "chain": "liquidv1",
  "blocks": 2199553,
...

  "current_fedpeg_program": "0020333a4af67452308b9814ff59dc2fc3d0325fd7fdcb61d108e992b3956095f2fe",
  "current_fedpeg_script": "5b21020e0338c96a...
  ```

The implementation walks backwards through the blocks at https://github.com/ElementsProject/elements/blob/master/src/pegins.cpp#L476

If a program uses (and caches) an expiring fedpegscript, it may result in invalid peg-in addresses being generated.

psgreco:
Fixed by https://github.com/ElementsProject/elements/pull/1203, closing

PR #1202, opened 2023-01-27
Correct error in rpc help for getsidechaininfo -- addresses  https://github.com/ElementsProject/elements/issues/1201

philippem:
- order of current_fedpegscripts was incorrectly listed as oldest first


philippem:
addresses https://github.com/ElementsProject/elements/issues/1201

PR #1203, opened 2023-01-27
Correct help for getsidechaininfo RPC

philippem:
- described order of current_fedpegscripts array was incorrect


philippem:
Addresses https://github.com/ElementsProject/elements/issues/1201

Issue #1208, opened 2023-02-08
Elements requires a dynafed transition before p2wsh pegins are enabled


psgreco:
Currently, even when elements starts in dynafed mode, it uses p2shwsh until the first transition, then it moves to p2wsh.

We should add an option so that regtest networks are able to start in p2wsh, so people can test current peg-ins in CI environments without doing a transition.

PR #1209, opened 2023-02-08
Allow deploying regtest networks with p2wsh peg-ins

psgreco:
Fixes: https://github.com/ElementsProject/elements/issues/1208

### DIFF
--- a/src/dynafed.cpp
+++ b/src/dynafed.cpp
@@ -52,6 +52,10 @@ DynaFedParamEntry ComputeNextBlockFullCurrentParameters(const CBlockIndex* pinde
         epoch_start_height -= epoch_length;
     }
 
+    if (consensus.genesis_style == "elements" && epoch_start_height == 0 && next_height > 1 && consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime == Consensus::BIP9Deployment::ALWAYS_ACTIVE) {
+        // when starting in dynafed mode, the full parameters are stored in block 1 instead of 0
+        epoch_start_height = 1;
+    }
     // We need to put in place the previous epoch's current which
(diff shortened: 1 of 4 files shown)


src/dynafed.cpp
```diff
@@ -52,6 +52,10 @@ DynaFedParamEntry ComputeNextBlockFullCurrentParameters(const CBlockIndex* pinde
         epoch_start_height -= epoch_length;
     }
 
+    if (consensus.genesis_style == "elements" && epoch_start_height == 0 && next_height > 1 && consensus.vDeployments[Consensus::DEPLOYMENT_DYNA_FED].nStartTime == Consensus::BIP9Deployment::ALWAYS_ACTIVE) {
+        // when starting in dynafed mode, the full parameters are stored in block 1 instead of 0
+        epoch_start_height = 1;
+    }
```
gwillen, line 58: This complicated edge case makes me nervous. I am assuming I don't need to worry about it affecting liquid prod, because epoch_start_height in liquid prod can never possibly be low enough to trigger it, at this point. (Except when looking at the very start of the chain, but I assume that's been tested and works.)

It seems like there ought to be a nicer way to do this, but maybe it's too late for that.

gwillen:
I am somewhat annoyed that this adds more special cases. We already have too many of those. It would be nice if that were avoidable.

HOWEVER, it's easy to verify that the special cases won't trigger in prod, at least, so they aren't risky from that angle.

utACK.

src/dynafed.cpp
psgreco: You don't need to worry about prod because dynafed is not "ALWAYS_ACTIVE" so even if we were regenerating a block from the first epoch somehow, it would behave the same way as always

PR #1210, opened 2023-02-14
Check in Andrew's script for merging changes into Elements integration branch

gwillen:
We've been using this for quite awhile but as far as I know it was not at any point checked in.

This PR also contains minor changes of mine, but the bulk of the code is written by Andrew Poelstra. (I have attributed the main commit to him, although I'm not 100% sure it doesn't contain any changes from me.)

### DIFF
--- a/contrib/merge-prs.sh
+++ b/contrib/merge-prs.sh
@@ -1,0 +1,123 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+BASE=merged-master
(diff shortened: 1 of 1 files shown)


apoelstra:
Just checked that you didn't leave the `-j80` in :) that OOMs on my laptop and is not fun.

Looks good to me.

In future we should really use nix for this but this existing script worked fine for at least 3 versions.

Issue #1211, opened 2023-02-21
Trim Headers: headers_ahead underflows when blocks are ahead of headers (somehow)


psgreco:
```
0;31m node2 stderr net_processing.cpp:4535:34: runtime error: implicit conversion from type 'int' of value -5 (32-bit, signed) to type 'uint64_t' (aka 'unsigned long') changed the value to 18446744073709551611 (64-bit, unsigned)
    #0 0x55e1eef43b78 in (anonymous namespace)::PeerManagerImpl::SendMessages(CNode*) ./net_processing.cpp:4535:34
    #1 0x55e1eef4e7ed in non-virtual thunk to (anonymous namespace)::PeerManagerImpl::SendMessages(CNode*) ./net_processing.cpp
    #2 0x55e1eeea7358 in CConnman::ThreadMessageHandler() ./net.cpp:2268:28
```

PR #1212, opened 2023-02-21
Fix underflow when blocks are ahead of headers

psgreco:
Fixes: https://github.com/ElementsProject/elements/issues/1211


gwillen:
utACK, good catch.

PR #1118, opened 2022-06-08
Improvements to pr review script

gwillen:
Improve the merge PR review script (for handling merges between upstream bitcoin core and elements) in a bunch of ways, including:
- More types of diffs to optionally look at
- Better handling of non-merge fixup commits
- Handling of new types of Bitcoin Core merge commits

PR #1214, opened 2023-02-24
Fix linter (trailing whitespace)

psgreco:
Fix linter by removing some trailing whitespaces introduced by https://github.com/ElementsProject/elements/pull/1118


delta1 (approved):
utACK a5976a6364850c7dfd764cdce852e17ad7a92ded

Issue #1224, opened 2023-03-16
Assert crash on getblockheader on a trimmed header


psgreco:
```
elements-qt: ./chain.h:223: void CBlockIndex::assert_untrimmed() const: Assertion `!m_trimmed' failed.
Aborted (core dumped)
```

PR #1225, opened 2023-03-16
RPC: Avoid assert by keeping a flag to identify trimmed dynafed blocks

psgreco:
This MR fixes 2 crashes:
1) assert in blockheaderToJSON when checking for dynafed block on a trimmed header
2) crash in blocktoJSON/blockheaderToJSON due to loaded block being incomplete.

Fixes: https://github.com/ElementsProject/elements/issues/1224

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -200,6 +200,7 @@ protected:
     std::optional<CScriptWitness> m_signblock_witness{};
 
     bool m_trimmed{false};
+    bool m_trimmed_dynafed_block{false};
 
     friend class CBlockTreeDB;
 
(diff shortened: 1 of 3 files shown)


delta1:
utack 8f16e00

delta1 (approved):

src/chain.h
```diff
@@ -228,6 +230,13 @@ class CBlockIndex
         return proof.value();
     }
 
+    const bool dynafed_block() const {
```
roconnor-blockstream, line 233: Clang says

    'const' type qualifier on return type has no effect

delta1: yeah this warning has been annoying me, I thought it was fixed on the 22.x branch but it is not

me639276639936: Thank you for posting this, but it has no effect on my RPI4 even though I have the latest version. On startup I get the same assert error. Any tips for me?

psgreco: If you see the assertion on startup, it usually means that you started IBD without `trim_headers` and then enabled `trim_headers` without finishing IBD, so the headers chain is much longer than the blocks. Either way, let's continue that conversation in https://github.com/ElementsProject/elements/issues/1240

PR #1232, opened 2023-04-04
test: re-enable and fix script assets test 

delta1:
Fixes the `script_assets_test` by changing the `feature_taproot.py` functional test to include the genesis block hash when dumping the json tests. The asset tests were failing with the Taproot flag enabled, because the Elements Taphash calculation requires the genesis block hash. 

Requires the json file from https://github.com/ElementsProject/qa-assets/pull/1

Steps to run the test: 
- compile this branch with tests enabled 
- checkout my updated branch of the qa-assets repo in the PR above 
-  run the test `DIR_UNIT_TEST_DATA=/path/to/qa-assets/unit_test_data ./src/test/test_bitcoin -t script_tests/script_assets_test`

### DIFF
--- a/src/test/script_tests.cpp
+++ b/src/test/script_tests.cpp
@@ -1684,7 +1684,7 @@ static void AssetTest(const UniValue& test)
     size_t idx = test["index"].get_int64();
     unsigned int test_flags = ParseScriptFlags(test["flags"].get_str());
     bool fin = test.exists("final") && test["final"].get_bool();
-    // ELEMENTS FIXME: update feature_taproot.py --dumptests to actually output these
+    // ELEMENTS: feature_taproot.py --dumptests outputs this field
     uint256 hash_genesis_block = test.exists("hash_genesis_block") ? uint256S(test["hash_genesis_block"].get_str()) : uint256{};
 
     if (test.exists("success")) {
@@ -1725,6 +1725,8 @@ BOOST_AUTO_TEST_CASE(script_assets_test)
 {
     // See src/test/fuzz/script_assets_test_minimizer.cpp for information on how to generate
     // the script_assets_test.json file used by this test.
+    // ELEMENTS: qa-assets repo can be cloned here https://github.com/ElementsProject/qa-assets
+    // then set DIR_UNIT_TEST_DATA env var to /path/to/qa-assets/unit_test_data when running this test
 
     const char* dir = std::getenv("DIR_UNIT_TEST_DATA");
     BOOST_WARN_MESSAGE(dir != nullptr, "Variable DIR_UNIT_TEST_DATA unset, skipping script_assets_test");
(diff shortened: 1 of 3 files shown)


src/test/script_tests.cpp
```diff
@@ -1744,10 +1746,11 @@ BOOST_AUTO_TEST_CASE(script_assets_test)
     BOOST_CHECK(tests.isArray());
     BOOST_CHECK(tests.size() > 0);
 
-    ELEMENTS: temporarily disabled until we implement the new Taproot sighash and upload new qa-assets
+    g_con_elementsmode = true;
```
apoelstra, line 1749: In feac5cf1d57be065d568bae69e756028a207054c:

Have you tested this? I am quite certain it is not going to work because the boost test harness is multithreaded.

delta1: oh interesting, I'm certain the test itself worked, but I'll run `make -j X check` to test all and confirm

apoelstra: Yes, the test will work by itself but dozens of other tests will fail.

test/functional/feature_taproot.py
```diff
@@ -131,9 +131,10 @@
 # while ctx2 only uses the modified hashtype inside the sighash calculation.
 #
 # ELEMENTS:
-# Elements taphash calculation also depends on genesis_block_hash which is stored as
-# `genesis_hash` in the test config
-g_genesis_hash = None
+# Elements taphash calculation also depends on the uint256 value of the genesis block hash
+# which is stored as genesis_uint256 in the test config
+g_genesis_uint256 = None
```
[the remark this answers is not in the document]
delta1: It's been a while, but I think at the time I was indicating it's not the serialized version. 

> intend to reverse the hash during serialization, you should do that explicitly 

Thanks I agree, will do so

src/test/script_tests.cpp
apoelstra: Ok, I'll try to reproduce locally. If it doesn't seem to cause problems then I guess it's okay.

apoelstra: It fails for me. `*** 160 failures are detected in the test module "Bitcoin Core Test Suite"`.

apoelstra: Actually, these failures all seem to be in the assets test itself ... I need to investigate further.

delta1: maybe an obvious question: are you using the updated qa-assets from ElementsProject/qa-assets#1

apoelstra (approved):
ACK 80c7751ed4bb4a69ebfa2c3b40915372864c2619

src/test/script_tests.cpp
apoelstra: Derp, that was it. Let me merge that over there.

apoelstra:
Can you squash or reorder the two commits closest to the tip (the one that re-enables the tests and the one that fixes them)?

Then every commit would compile and pass tests in isolation.

delta1:
> Can you squash or reorder the two commits closest to the tip

Sure! Done

apoelstra:
I have queued this up to run a local CI matrix; will ACK and merge when that's done (assuming no serious failures in the actual CI here).

apoelstra (approved):
ACK 5cbd44f1c89533b11548b411abce416f49d2ad8b

Issue #1234, opened 2023-04-19
Elements QT always shows default asset as L-BTC whereas elements-cli respects the defaultpeggedassetname param


wintercooled:
If you start a new chain (standalone, not pegged) with a `defaultpeggedassetname` specified `elements-cli` respects this when you call `getwalletinfo` etc but Elements QT always shows the default asset as `L-BTC`. If you `dumpassetlabels` and then use the asset id of the default asset as an `assetdir` entry in your config Elements QT errors saying there is a duplicate entry. So it's as if it knows that the default asset id is mapped to a name but it doesn't show it in the UI. Example config that uses the `defaultpeggedassetname` value etc to show this:

```
chain=standalone
rpcuser=user1
rpcpassword=password1
daemon=1
server=1
listen=1
txindex=1
fallbackfee=0.0002
[standalone]
rpcport=18884
port=18886
anyonecanspendaremine=1
connect=localhost:18887
validatepegin=0
defaultpeggedassetname=newasset
initialfreecoins=100000000000000
initialreissuancetokens=200000000
```

PR #1236, opened 2023-04-24
Elements-qt: Show correct default asset name

andreabonel:
Fixes: https://github.com/ElementsProject/elements/issues/1234


gwillen:
utACK

Issue #1238, opened 2023-05-04
`signrawtransactionwithwallet` produces invalid Schnorr signatures


apoelstra:
If you try modifying the `wallet_taproot.py` functional test like so:

```
diff --git a/test/functional/wallet_taproot.py b/test/functional/wallet_taproot.py
index 123a6a40893..f37fc906549 100755
--- a/test/functional/wallet_taproot.py
+++ b/test/functional/wallet_taproot.py
@@ -275,6 +275,12 @@ class WalletTaprootTest(BitcoinTestFramework):
             self.nodes[0].generatetoaddress(1, self.boring.getnewaddress())
             test_balance = int(self.rpc_online.getbalance()['bitcoin'] * 100000000)
             ret_amnt = random.randrange(100000, test_balance)
+            # Start...
+            rawret = self.rpc_online.createrawtransaction([], [{self.boring.getnewaddress(): Decimal(ret_amnt) / 100000000}])
+            rawret = self.rpc_online.fundrawtransaction(rawret, {"changePosition":1, "subtractFeeFromOutputs":[0]})
+            rawret = self.rpc_online.signrawtransactionwithwallet(rawret['hex'])
+            self.rpc_online.sendrawtransaction(rawret['hex'])
+            # ...end
             res = self.rpc_online.sendtoaddress(address=self.boring.getnewaddress(), amount=Decimal(ret_amnt) / 100000000, subtractfeefromamount=True)
             self.nodes[0].generatetoaddress(1, self.boring.getnewaddress())
             assert(self.rpc_online.gettransaction(res)["confirmations"] > 0)
```

You will get a failure of the form
```
test_framework.authproxy.JSONRPCException: non-mandatory-script-verify-flag (Invalid Schnorr signature) (-26)
```
on the `sendrawtransaction` line. You can verify by printing intermediate values that the output of `signrawtransactionwithwallet` shows `complete: true` and that the transaction indeed has something Schnorr-signature-shaped in its witness.

Issue #1237, opened 2023-05-04
`signrawtransactionwithwallet` does not take enough information to sign


apoelstra:
If you try to call `signrawtransactionwithwallet` on an unconfirmed input, you need to provide the input's information, including the amount (for segwit) and asset (for Taproot). But the RPC does not take the asset. So you cannot sign Taproot outputs with this RPC.

PR #1239, opened 2023-05-04
testmempoolaccept: feed genesis hash into sighash computation

apoelstra:
I cannot test this because of #1238 and #1237 as well as the lack of functionality in the PSBT RPCs, inability to inform the wallet about a UTXO without broadcasting the transaction it comes from, etc etc.


delta1 (approved):
utACK ccffeba3b281aa067882bc1816587082bfc4704a

PR #1246, opened 2023-07-14
fix: getnewaddress - don't blind bech32 addresses

### LINKED ISSUE #1245: bug: getnewaddress bech32 returns a blech32 address
it should correctly return an unblinded bech32 address 

reported by @roconnor-blockstream

delta1:
fixes #1245 - `getnewaddress` will now return an unblinded bech32 address even when `-blindedaddresses=1` in config

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -294,7 +294,9 @@ static RPCHelpMan getnewaddress()
         label = LabelFromValue(request.params[0]);
 
     OutputType output_type = pwallet->m_default_address_type;
-    bool force_blind = false;
+    // default blinding to the blindedaddresses setting
+    bool add_blinding_key = gArgs.GetBoolArg("-blindedaddresses", g_con_elementsmode);
+
     if (!request.params[1].isNull()) {
         if (!ParseOutputType(request.params[1].get_str(), output_type)) {
             throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, strprintf("Unknown address type '%s'", request.params[1].get_str()));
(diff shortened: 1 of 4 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -294,23 +294,27 @@ static RPCHelpMan getnewaddress()
...
         }
         if (output_type == OutputType::BECH32M && pwallet->GetLegacyScriptPubKeyMan()) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Legacy wallets cannot provide bech32m addresses");
         }
-        // Special case for "blech32" when `-blindedaddresses=0` in the config.
         if (request.params[1].get_str() == "blech32") {
-            force_blind = true;
+            // always blind for "blech32" even if `-blindedaddresses=0` in the config.
+            add_blinding_key = true;
+        } else if (output_type == OutputType::BECH32) {
```
roconnor-blockstream, line 310: It seems to me that this condition ought to be `request.params[1].get_str() == "bech32"`  to match the branch above.

A careful analysis of the logic has convinced me that what you have written here is logically the same as my proposal. (If `output_type` is `OutputType::BECH32`, then `request.params[1].get_str()` is either `"bech32"` or `"blech32"`, and if we are in this else clause then `request.params[1].get_str() != "blech32"`, and hence `request.params[1].get_str() == "bech32"`.)  But still, it seems unusual to consider `request.params[1].get_str()` in one branch but not the other.

apoelstra:
059a4a621cc226ff339c57c9d62211562c616de2 looks fine to me.

Locally the following tests fail
```
elements-059a4a> feature_confidential_transactions.py               | ✖ Failed  | 1 s
elements-059a4a> feature_sighash_rangeproof.py                      | ✖ Failed  | 2 s
elements-059a4a> rpc_signrawtransaction.py --legacy-wallet          | ✖ Failed  | 6 s
```

These look like real failures; they happen consistently.

delta1:
Thanks @apoelstra - fixed

apoelstra (approved):
ACK 49c3ea309ec061729c675469a17331d53e392ba5

roconnor-blockstream:
Now I'm a bit concerned that this change will  break existing workflows of people by unblinding their addresses on thm.

apoelstra:
It may, but I think the old behavior was probably wrong. Not sure how best to handle this.

Issue #1241, opened 2023-06-13
Crashes after being off for a few days


me639276639936:
RPI4 with pruned node and trim headers. 

Fails sync when machine has been off a few days.

PR #1249, opened 2023-07-18
test: add trim_headers functional test

### LINKED ISSUE #1242: create a trim_headers functional test
we really need a functional test that exercises the `trim_headers` mode (#1190) 

this would be useful to make sure that it continues to work even after merging change from upstream or any other fixes/features

delta1:
fixes #1242 

This test is based on `feature_blocksign.py`. It runs 1 block-signing nodes and 2 syncing nodes with `trim_headers=1`, 1 of which also has `-prune=550`.  

This reproduces the crash in #1241

### DIFF
--- a/test/functional/feature_trim_headers.py
+++ b/test/functional/feature_trim_headers.py
@@ -1,0 +1,268 @@
+#!/usr/bin/env python3
+
+import codecs
+from decimal import Decimal
+from time import sleep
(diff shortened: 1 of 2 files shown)


delta1:
Updated the test and description

delta1:
rebased on master

delta1:
was included in #1270

PR #1254, opened 2023-08-01
Improve error reporting from VerifyAmounts

stevenroose:
When running this, I get a segmentation fault, so there is something wrong with the references or something.. borrowck, plz help me..


delta1:
What's your input? You're getting a segfault when running SignPSBT with this patch?

apoelstra:
I think if we're going to be touching core consensus code like this we should do it properly, with an error enum.

PR #1256, opened 2023-08-08
Small fixes in preparation for trim_headers version 2

psgreco:
Preparing for the new version of trim headers, both in 22 and 23, there are a few commits that don't create conflicts and that it would be easier to have before branching.


delta1 (approved):
ACK 88c3b6f8cfad987718d636a3ee90c991369ccd12

PR #1258, opened 2023-08-16
KnapsackSolver fix

goatpig:
- KnapsackSolver applies non_policy_effective_value incorrectly to policy output amount. This leads to a negative expected fee in lots of inputs with few outputs scenarios (e.g. utxo consolidations). Fixed that and added a comment.

- Without this fix, an assert on policy change amount vs fee paid will trigger (spend.cpp:1449). I consider asserts too aggressive a check for these kinds of condition (this will crash elements unless an explicit handler for SIGSEVs is registered). I took to liberty of replacing the assert with a log event and graceful failure (returning false).

- No new tests was added to cover for this issue,  functional tests pass.

### DIFF
--- a/src/wallet/coinselection.cpp
+++ b/src/wallet/coinselection.cpp
@@ -299,7 +299,13 @@ bool KnapsackSolver(const CAmountMap& mapTargetValue, std::vector<OutputGroup>& 
     }
 
     // Perform the standard Knapsack solver for the policy asset
-    CAmount policy_target = non_policy_effective_value + mapTargetValue.at(::policyAsset);
+    /*
+    NOTE:
+    CInputCoin::effective_value is a negative value so non_policy_effective_value
+    will be negative. Therefor, it should be substracted from the policy output
+    amount, not added.
+    */
+    CAmount policy_target = mapTargetValue.at(::policyAsset) - non_policy_effective_value;
     if (policy_target > 0) {
         inner_groups.clear();
         inner_coinsret.clear();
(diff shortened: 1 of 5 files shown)


delta1:
Hi @goatpig thanks for the PR. I want to look into adding a unit test to cover this specifically. 

Can you tell me more about the transaction that triggered the assert? Does it have many outputs, with multiple asset types?

goatpig:
@delta1
Hi,

The issue triggered while creating a transaction that aimed to consolidate some 20+ USDT 2-of-3 inputs into a single output.

To reproduce you need to redeem enough non-LBTC inputs for the total estimated fee cost of these inputs to overwhelm the base cost fed to KnapsackSolver. The base cost is that of the raw tx before inputs (see CWallet::AttemptSelection)

For reference, in our case with a single explicit output, the coin selection algo set the fee output to roughly 250 sats. A single one of our 2-of-3 USDT input had its cost set to -31 (set in CInputCoin::effective_value).

delta1:
I've added a functional test that fails without this fix, thanks @goatpig

src/wallet/coinselection.cpp
```diff
@@ -299,7 +299,13 @@ bool KnapsackSolver(const CAmountMap& mapTargetValue, std::vector<OutputGroup>&
     }
 
     // Perform the standard Knapsack solver for the policy asset
-    CAmount policy_target = non_policy_effective_value + mapTargetValue.at(::policyAsset);
+    /*
+    NOTE:
+    CInputCoin::effective_value is a negative value so non_policy_effective_value
+    will be negative. Therefor, it should be substracted from the policy output
+    amount, not added.
```
delta1, line 306:
```suggestion
    CInputCoin::effective_value is negative for non-policy assets, so the sum non_policy_effective_value is negative. Therefore, it is subtracted in order to increase policy_target by the fees required.
```

delta1 (approved):
ACK cf0f56107be24c0272df08c449444d47b92a9c15 

thanks @goatpig

goatpig:
Thanks, deleting the branch.

Issue #1263, opened 2023-09-06
Segfault on getpeginaddress


evd0kim:
When validatepegin=0 is provided in the config, typing in getpeginaddress leads to a crash.

```
2023-09-06T10:24:19Z init message: Loading wallet...
2023-09-06T10:24:19Z [default wallet] Wallet File Version = 10500
2023-09-06T10:24:19Z [default wallet] Keys: 0 plaintext, 0 encrypted, 0 w/ metadata, 0 total. Unknown wallet records: 0
2023-09-06T10:24:19Z [default wallet] keypool added 2000 keys (1000 internal), size=2000 (1000 internal)
2023-09-06T10:24:19Z [default wallet] LegacyScriptPubKeyMan::NewKeyPool rewrote keypool
2023-09-06T10:24:20Z [default wallet] Wallet completed loading in             910ms
2023-09-06T10:24:20Z [default wallet] setKeyPool.size() = 2000
2023-09-06T10:24:20Z [default wallet] mapWallet.size() = 0
2023-09-06T10:24:20Z [default wallet] m_address_book.size() = 0
2023-09-06T10:24:22Z Flushing wallet.dat
2023-09-06T10:24:22Z Flushed wallet.dat 20ms
2023-09-06T10:24:22Z Received a POST request for / from 127.0.0.1:37374
2023-09-06T10:24:22Z ThreadRPCServer method=getpeginaddress user=lquser1 peeraddr=127.0.0.1:37374
2023-09-06T10:24:22Z [default wallet] keypool reserve 1
2023-09-06T10:24:22Z [default wallet] keypool keep 1
Segmentation fault (core dumped)
```

`-chain=main`.

delta1:
> -chain=main 

are you running this elementsd with `-chain=main` and `validatepegin=0` ? 

afaik `chain=main` is the Bitcoin blockchain, and `getpeginaddress` is trying to find dynafed headers which are not on that chain. 

can you please tell me more about your setup/requirements?

evd0kim:
> are you running this elementsd with -chain=main and validatepegin=0 
Yes. But just as an experiment.

Because I can not run it with validatepegin=1 and it seems there is another bug.

> can you please tell me more about your setup/requirements?

I have no requirements yet I am just investigating documentation and trying to do a setup for "sidechain" with "strong federation" running along with mainnet Core on remote host.

delta1:
> I can not run it with validatepegin=1 and it seems there is another bug 

any more info? 

[...]

PR #1276, opened 2023-10-18
fix: segfault on getpeginaddress rpc when not in elements mode

delta1:
Fixes https://github.com/ElementsProject/elements/issues/1263

When not running in Elements mode, GetValidFedpegScripts returns an empty vector. Subsequently `calculate_contract` is called with `fedpegscripts.front().second` which is a null reference. This reference is passed into `MatchLiquidWatchman` which tries to dereference the null reference and causes a segfault.

This fix throws an exception in `getpeginaddress` if fedpegscripts is empty, returning an error to the RPC caller. It also adds a functional regression test.


psgreco:
UTACK 5326676ff3eb9b9a47a77c48567a442171ecc001

PR #1281, opened 2023-11-02
Fix CI on simplicity branch

uncomputable:
Fix the CI that currently fails on the simplicity branch.


delta1 (approved):
ACK 09fcaec

test/lint/lint-files.py
```diff
@@ -19,7 +19,7 @@
 ALLOWED_FILENAME_REGEXP = "^[a-zA-Z0-9/_.@][a-zA-Z0-9/_.@-]*$"
 ALLOWED_SOURCE_FILENAME_REGEXP = "^[a-z0-9_./-]+$"
 ALLOWED_SOURCE_FILENAME_EXCEPTION_REGEXP = (
-    "^src/(secp256k1/|univalue/|test/fuzz/FuzzedDataProvider.h)"
+    "^src/(secp256k1/|univalue/|test/fuzz/FuzzedDataProvider.h|simplicity)"
```
roconnor-blockstream, line 22: Should have a slash after `simplicity` like it does for `secp256k1`?

uncomputable: I read `/|` as an escape sequence, like `\(`. Not an expert here.

PR #1286, opened 2023-11-10
Script: Move SCRIPT_ERR_COUNT

### LINKED ISSUE #1285: Bad enumeration of script errors
https://github.com/ElementsProject/elements/blob/20c5630f8c13c3df37c8a34c96fdbeceb8af1e27/src/script/script_error.h#L87

has incorrectly inserted new Elements errors into this enumeration.  All the new elements errors need to go **before** `SCRIPT_ERR_ERROR_COUNT`, which isn't actually a real error.  Then `SCRIPT_ERR_LAST` needs to be reset to `SCRIPT_ERR_ERROR_COUNT`, as it originally was.

AFAICT the only consequences of this error are for the fuzzer.

I'm not sure that having both `SCRIPT_ERR_LAST` and `SCRIPT_ERR_ERROR_COUNT` makes sense, but that is how it is in Bitcoin at the moment.

cc @uncomputable

uncomputable:
Fix #1285


delta1 (approved):
ACK f4243128eeaad039cf0e70347935acf6c61789c7

Issue #1199, opened 2022-12-22
BlindTransaction() and VerifyAmounts() need detailed error return codes


philippem:
The functions `BlindTransaction()` and `VerifyAmounts()` each have several failure paths, but only use -1 and false to indicate to the caller that an operation failed. This makes it difficult to report helpful error messages to the user.

The error paths could return an enum indicating the reason for failure, and the callers could then surface error strings to the user.

`BlindTransaction()`: https://github.com/ElementsProject/elements/blob/master/src/blind.cpp#L286
`VerifyAmounts()`: https://github.com/ElementsProject/elements/blob/master/src/confidential_validation.cpp#L134

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

CrypticBantu:
Is there any work currently being done on this? if not, please assign.

delta1:
Go ahead @CrypticBantu ! Thank you

philippem:
Here's an example of an error message resulting from too many inputs:


```
bash-5.1# elements-cli -rpcwallet=oldbdb -named sendtoaddress address=lq1qq2dt7w63n8xk6qlsmg9cmywpenhdmrlw6qkdp3q3hkngkppek0pa27eqan7s74g32nad3lmg6hd90vfdqpx7jc0meqvvk778e amount=0.00000528 assetlabel=78557eb89ea8439dc1a519f4eb0267c86b261068648a0f84a5c6b55ca39b66f1
error code: -6
error message:
Unable to blind the transaction properly. This should not happen.

```

PR #1288, opened 2023-11-14
refactor: change BlindTransaction to return more detailed error information

delta1:
re: #1199 

looking for concept/approach feedback on this change


delta1:
a31e125b99cd801e4b24b789e5528d0c18412121 updates the error message to specify if the issue is because of trying to blind too many inputs

PR #1293, opened 2023-11-22
ci: fix macos failing functional tests 

delta1:
The macOS 13 CI job in elements-23.x defaults to descriptor wallets, so `example_elements_code_tutorial.py` fails with "insufficient funds". 

Fixed by running the test with legacy wallet.


delta1:
added same fix for regression test 1263

delta1:
fixed in #1392

PR #1317, opened 2024-02-09
feat: discounted fees for confidential transactions

delta1:
This is a mempool policy only change to enable discounted fees for Confidential Transactions.

Two new configuration options are added to toggle accepting and creating these discounted CTs. `accept` defaults to true for liquidv1 only, false for other chains. `create` defaults to false on all chains for now. 

At a feerate of 1 sat/vb for a 2 input, 3 output transaction:
- explicit transaction has a fee of 326 sats
- confidential transaction has a fee of 410 sats (without discount: 2575 sats)  

This change is optional and backwards compatible for nodes and wallets. 

This PR also alters block template transaction ordering, such that transactions are ordered by their feerate according to `discountvsize`. This means that discounted CT won't be selected last due to their lower *real* feerate according to vsize.


src/primitives/transaction.cpp
```diff
@@ -162,3 +162,12 @@ std::string CTransaction::ToString() const
         str += "    " + tx_out.ToString() + "\n";
     return str;
 }
+
+bool CTransaction::IsConfidential() const
+{
+    for (const CTxOut& out : vout) {
+        if (out.IsFee()) continue;
+        if (out.nValue.IsExplicit() || out.nAsset.IsExplicit()) return false;
```
tiero, line 170: is this the right behavior: a transaction can have both confidential and unconfidential outputs, does this mean that to obtain the discount would suffice to have a single output being blinded then?

tiero:
Please can you clearly add to the description if this change impacts wallets and other Elements implementation?

psgreco:
> Please can you clearly add to the description if this change impacts wallets and other Elements implementation?

The idea for this change is that it will be optional, and more beneficial to users, so if wallets keep doing the same thing they were doing, it will still be valid, they will just be overpaying a bit based on the new standard.

src/primitives/transaction.cpp
delta1: the way it currently works here is that all outputs in the transaction must be blinded (except fee output)

delta1:
yes this is optional and backwards compatible, and I still need to write an ELIP describing it for other implementations

delta1:
Force pushed with the change to calculate the discount per confidential output

delta1:
Force pushed to include commit which changes the block template ordering

src/policy/discount.h
```diff
@@ -0,0 +1,45 @@
...
+    for (size_t i = 0; i < tx.vout.size(); ++i) {
+        const CTxOut& output = tx.vout[i];
+        if (output.IsFee()) continue;
+        if (output.nAsset.IsCommitment() && output.nValue.IsCommitment()) {
+            // subtract the weight of the output witness
+            weight -= ::GetSerializeSize(tx.witness.vtxoutwit[i], PROTOCOL_VERSION);
+            // subtract the weight difference of amount commitment (33) vs explicit amount (9)
+            weight -= (33 - 9);
+            // subtract the weight difference of the confidential nonce (33)
+            weight -= 33;
```
roconnor-blockstream, line 34: A non-confidential nonce is 1 byte.

roconnor-blockstream (commented):
My review is incomplete, but this is my first pass.

src/policy/discount.h
apoelstra: Also, a confidential asset is 33 bytes while a non-confidential asset is 32.

I think this code would be clearer if we added a method to `CConfidentialCommitment` which computed `vchCommitment.size() - ExplicitSize`, and then called that here (without gating on `IsCommitment`) for all three output values.

src/txmempool.cpp
```diff
@@ -965,7 +966,14 @@ void CTxMemPool::queryHashes(std::vector<uint256>& vtxid) const
 }
 
 static TxMempoolInfo GetInfo(CTxMemPool::indexed_transaction_set::const_iterator it) {
-    return TxMempoolInfo{it->GetSharedTx(), it->GetTime(), it->GetFee(), it->GetTxSize(), it->GetModifiedFee() - it->GetFee()};
+    // ELEMENTS: include the discounted vsize of the tx
+    size_t discountvsize = it->GetTxSize();
+    CTransaction tx = it->GetTx();
```
apoelstra, line 983: In 8bafd0c9fa448c3790ed2992f02b8e1245e6c56a:

This variable assignment should be moved inside the `if` block.

src/validation.cpp
```diff
@@ -914,7 +914,12 @@ bool MemPoolAccept::PreChecks(ATMPArgs& args, Workspace& ws)
 
     // No transactions are allowed below minRelayTxFee except from disconnected
     // blocks
-    if (!bypass_limits && !CheckFeeRate(ws.m_vsize, ws.m_modified_fees, state)) return false;
+    bool fee_check = CheckFeeRate(ws.m_vsize, ws.m_modified_fees, state);
```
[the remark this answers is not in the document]
apoelstra: This is true, and it may be clearer to make this an `if`/`else`, but the poisoning of `state` doesn't matter the way that it's used in this codebase. `state` is only accessed when a function returns a falsey value (which are all of the form `state.Invalid` which re-poisons `state` overwriting any previous incorrect value). So either everything else succeeds after this line, and `state` is never accessed, or something else false, and `state` is correctly set to `Invalid`.

But great catch! Let's fix this to prevent any future bugs.

This business of using pseudo-globals to emulate monadic error handling is peak C++..

src/policy/discount.h
apoelstra: (In 8bafd0c9fa448c3790ed2992f02b8e1245e6c56a)

test/functional/feature_discount_ct.py
```diff
@@ -0,0 +1,151 @@
...
+        info = node0.getaddressinfo(addr)
+        txid = node0.sendtoaddress(info['unconfidential'], 1.0, "", "", False, None, None, None, None, None, None, feerate)
+        tx = node0.gettransaction(txid, True, True)
+        decoded = tx['decoded']
+        vin = decoded['vin']
+        vout = decoded['vout']
+        assert_equal(len(vin), 2)
+        assert_equal(len(vout), 3)
+        assert_equal(tx['fee']['bitcoin'], Decimal('-0.00000326'))
+        assert_equal(decoded['vsize'], 326)
```
apoelstra, line 76: In 8bafd0c9fa448c3790ed2992f02b8e1245e6c56a:

Everywhere you assert the value of `vsize` can you also assert the value of `discountvsize`?

apoelstra: Also in the ordering test in 41abd40b5cdbc910cabac009e4e27e670b2e502a

src/txmempool.h
```diff
@@ -103,20 +103,21 @@ class CTxMemPoolEntry
...
+    LockPoints lockPoints;          //!< Track the height and time at which tx was final
 
     // Information about descendants of this transaction that are in the
     // mempool; if we remove this transaction we must remove all of these
     // descendants as well.
     uint64_t nCountWithDescendants{1}; //!< number of descendant transactions
-    uint64_t nSizeWithDescendants;   //!< ... and size
-    CAmount nModFeesWithDescendants; //!< ... and total fees (all including us)
+    uint64_t nSizeWithDescendants;     //!< ... and size
+    CAmount nModFeesWithDescendants;   //!< ... and total fees (all including us)
```
apoelstra, line 113: In 41abd40b5cdbc910cabac009e4e27e670b2e502a

More formatting changes here that would be better to avoid.

apoelstra: In fact the majority of changes to this file in this commit are just formatting.

apoelstra:
Done reviewing 41abd40b5cdbc910cabac009e4e27e670b2e502a

src/node/miner.cpp
```diff
@@ -383,7 +384,8 @@ void BlockAssembler::addPackageTxs(int& nPackagesSelected, int& nDescendantsUpda
     // and modifying them for their already included ancestors
     UpdatePackagesForAdded(inBlock, mapModifiedTx);
 
-    CTxMemPool::indexed_transaction_set::index<ancestor_score>::type::iterator mi = m_mempool.mapTx.get<ancestor_score>().begin();
+    // ELEMENTS: we use confidential_score instead of ancestor_score
+    CTxMemPool::indexed_transaction_set::index<confidential_score>::type::iterator mi = m_mempool.mapTx.get<confidential_score>().begin();
```
roconnor-blockstream, line 388: This might be working as intended, but just to be clear, what we are doing here is *unconditionally* altering the block production function to sort by discounted weight, on every chain, even those chains with `acceptdiscount = false`!

src/policy/discount.h
roconnor-blockstream: > Also, a confidential asset is 33 bytes while a non-confidential asset is 32.

False.  Both are 33 bytes.

src/txmempool.h
delta1: yeah missed this auto-format, will revert

src/txmempool.cpp
```diff
@@ -965,7 +978,14 @@ void CTxMemPool::queryHashes(std::vector<uint256>& vtxid) const
...
+    // discountvsize only differs from vsize if we accept discounted CTs
+    if (Params().GetAcceptDiscountCT()) {
+        discountvsize = GetDiscountVirtualTransactionSize(tx);
```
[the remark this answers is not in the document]
delta1: Good catch, thanks. Will fix

src/net_processing.cpp
```diff
@@ -4964,7 +4964,9 @@ bool PeerManagerImpl::SendMessages(CNode* pto)
                         auto txid = txinfo.tx->GetHash();
                         auto wtxid = txinfo.tx->GetWitnessHash();
                         // Peer told you to not send transactions at that feerate? Don't bother sending it.
-                        if (txinfo.fee < filterrate.GetFee(txinfo.vsize)) {
+                        // ELEMENTS: use the discounted vsize here so that discounted CTs are relayed.
```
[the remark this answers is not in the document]
delta1: No, nodes that don't accept log the message 

> [net_processing.cpp:3542] [ProcessMessage] b90...617 from peer=0 was not accepted: min relay fee not met, 409 < 2575  

And then call `MaybePunishNodeForTx`, which does not ban for `TX_MEMPOOL_POLICY` 

https://github.com/ElementsProject/elements/blob/master/src/net_processing.cpp#L1448

src/validation.cpp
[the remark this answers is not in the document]
delta1: fixed in e6b0945f0d3c4c1228d27fcdd8cca205f95f0f89

src/policy/discount.h
delta1: > vchCommitment.size() - ExplicitSize 

this doesn't work for Nonce which has explicit size defined as 33? 

changed in d0c18edda0274dab7d8582f521e9bd86ccd85925, where I've added an `IsConfidential` method to abstract this

src/txmempool.cpp
delta1: unnecessary after 6c67c5e842ef4c9db22095b5bfcdccb2add12540

test/functional/feature_discount_ct.py
delta1: done

src/policy/discount.h
```diff
@@ -0,0 +1,45 @@
...
+    int64_t size_bytes = ::GetSerializeSize(tx, PROTOCOL_VERSION | SERIALIZE_TRANSACTION_NO_WITNESS) * (WITNESS_SCALE_FACTOR - 1) + ::GetSerializeSize(tx, PROTOCOL_VERSION);
+    int64_t sigop_bytes = nSigOpCost * bytes_per_sig_op;
+
+    int64_t weight = std::max(size_bytes, sigop_bytes);
+
+    // for each confidential output
+    for (size_t i = 0; i < tx.vout.size(); ++i) {
+        const CTxOut& output = tx.vout[i];
+        if (output.IsFee()) continue;
+        if (output.nAsset.IsCommitment() && output.nValue.IsCommitment()) {
```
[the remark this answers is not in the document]
delta1: done in d0c18edda0274dab7d8582f521e9bd86ccd85925

src/txmempool.cpp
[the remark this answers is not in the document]
delta1: done

src/txmempool.cpp
```diff
@@ -965,7 +978,14 @@ void CTxMemPool::queryHashes(std::vector<uint256>& vtxid) const
 }
 
 static TxMempoolInfo GetInfo(CTxMemPool::indexed_transaction_set::const_iterator it) {
-    return TxMempoolInfo{it->GetSharedTx(), it->GetTime(), it->GetFee(), it->GetTxSize(), it->GetModifiedFee() - it->GetFee()};
+    // ELEMENTS: include the discounted vsize of the tx
+    size_t discountvsize = it->GetTxSize();
+    CTransaction tx = it->GetTx();
+    // discountvsize only differs from vsize if we accept discounted CTs
+    if (Params().GetAcceptDiscountCT()) {
```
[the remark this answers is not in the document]
delta1: Thank you for catching this, updated

src/node/miner.cpp
delta1: `CTxMemPoolEntry::GetDiscountTxSize()` fixed to be conditional, and added into the ordering test to check that nodes that don't accept discounted CT have their ordering as per ancestor package fee rate

src/validation.cpp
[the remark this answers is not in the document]
roconnor-blockstream: You applied to the `return state.Valid();` to the wrong function.

Furthermore, IMHO, we want to endevour to minimize the diff between upstream bitcoin and elements in order to make merging as easy as we can reasonably get away with.  Granted, I'm not in charge of elements development, but adding a whole new method to `class ValidationState` doesn't strike me as minimizing the diff, especially when all we need to do is use an if statement or a conditional operator.

e.g. just do the following:

```
   if (!bypass_limits && !CheckFeeRate(Params().GetAcceptDiscountCT() ? GetDiscountVirtualTransactionSize(tx) : ws.m_vsize, ws.m_modified_fees, state)) return false;
```
or, if you think that line is too long (which it probably is) then

```
   int64_t fee_rate_tx_size = Params().GetAcceptDiscountCT() ? GetDiscountVirtualTransactionSize(tx) : ws.m_vsize
   if (!bypass_limits && !CheckFeeRate(fee_rate_tx_size, ws.m_modified_fees, state)) return false;
```

This keeps the original (contrived) logic that Bitcoin uses for error handling in place.

src/txmempool.cpp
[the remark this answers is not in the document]
roconnor-blockstream: Now that `GetDiscountTxSize` checks `Params().GetAcceptDiscountCT()`, you no longer need to do that check here.

src/txmempool.cpp
```diff
@@ -464,6 +481,7 @@ void CTxMemPoolEntry::UpdateAncestorState(int64_t modifySize, CAmount modifyFee,
     assert(int64_t(nCountWithAncestors) > 0);
     nSigOpCostWithAncestors += modifySigOps;
     assert(int(nSigOpCostWithAncestors) >= 0);
+    discountSizeWithAncestors += discountSize;
```
roconnor-blockstream, line 484: `assert(int64_t(discountSizeWithAncestors) > 0);`

src/policy/discount.h
```diff
@@ -0,0 +1,46 @@
...
+{
+    int64_t size_bytes = ::GetSerializeSize(tx, PROTOCOL_VERSION | SERIALIZE_TRANSACTION_NO_WITNESS) * (WITNESS_SCALE_FACTOR - 1) + ::GetSerializeSize(tx, PROTOCOL_VERSION);
+    int64_t sigop_bytes = nSigOpCost * bytes_per_sig_op;
+
+    int64_t weight = std::max(size_bytes, sigop_bytes);
+
+    // for each confidential output
+    for (size_t i = 0; i < tx.vout.size(); ++i) {
+        const CTxOut& output = tx.vout[i];
+        if (output.IsFee()) {
```
roconnor-blockstream, line 27: The test for `IsFee` isn't needed, and probably should just be removed unless you have a particular reason for wanting it anyways.

src/policy/discount.h
```diff
@@ -0,0 +1,46 @@
...
+            continue;
+        } else if (output.IsConfidential()) {
+            // subtract the weight of the output witness
+            weight -= ::GetSerializeSize(tx.witness.vtxoutwit[i], PROTOCOL_VERSION);
```
roconnor-blockstream, line 31: I'm afraid you are going to need to wrap this in an `if (i < tx.witness.vtxoutwit.size())`.

src/validation.cpp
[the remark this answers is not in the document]
delta1: Thanks, reverted that commit and updated in 70b9c024b158a45e7aeade64033382183d803bf5

src/txmempool.cpp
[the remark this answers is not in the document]
delta1: thanks done in 063a73ac2714767d79b524d05076fa30ce9b6674

src/txmempool.cpp
delta1: done

src/policy/discount.h
delta1: done

src/policy/discount.h
delta1: Great catch, thanks. Done.

src/wallet/spend.cpp
```diff
@@ -174,6 +174,11 @@ TxSize CalculateMaximumSignedTxSize(const CTransaction &tx, const CWallet *walle
     CTransaction ctx(txNew);
     int64_t vsize = GetVirtualTransactionSize(ctx);
     int64_t weight = GetTransactionWeight(ctx);
+    // ELEMENTS: use discounted vsize for CTs if enabled
+    if (Params().GetCreateDiscountCT()) {
```
roconnor-blockstream, line 178: I think it doesn't make any sense to CreateDiscountCT while not AcceptDiscountCT.

I think you will need to add GetAcceptDiscountCT to this `if` statement or otherwise make the configuration of CreateDiscountCT without AcceptDiscountCT illegal.

delta1: Thanks, agreed. I think it's best that if "create" is true then "accept" is also set to true. 

Done in 7fc9651

src/core_write.cpp
```diff
@@ -236,6 +237,12 @@ void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry,
     entry.pushKV("version", static_cast<int64_t>(static_cast<uint32_t>(tx.nVersion)));
     entry.pushKV("size", (int)::GetSerializeSize(tx, PROTOCOL_VERSION));
     entry.pushKV("vsize", (GetTransactionWeight(tx) + WITNESS_SCALE_FACTOR - 1) / WITNESS_SCALE_FACTOR);
+    // ELEMENTS: add discountvsize
+    if (Params().GetAcceptDiscountCT()) {
+        entry.pushKV("discountvsize", GetDiscountVirtualTransactionSize(tx));
+    } else {
+        entry.pushKV("discountvsize", (GetTransactionWeight(tx) + WITNESS_SCALE_FACTOR - 1) / WITNESS_SCALE_FACTOR);
+    }
```
apoelstra, line 245: In 8bafd0c9fa448c3790ed2992f02b8e1245e6c56a:

I think it'd be better to just not have the `discountvsize` field at all, if the node does not accept discount CTs. This is easier to machine-detect and makes the field have only one meaning.

src/policy/discount.h
apoelstra: In 8bafd0c9fa448c3790ed2992f02b8e1245e6c56a:

Still not done (it may be done in a later commit, I am not there yet)

src/primitives/transaction.h
```diff
@@ -345,6 +345,12 @@ class CTxOut
     {
         return a.scriptPubKey < b.scriptPubKey;
     }
+
+    // ELEMENTS: a "confidential" output has commitments instead of explicit values
+    bool IsConfidential() const
+    {
+        return nValue.IsCommitment() && nAsset.IsCommitment() && nNonce.IsCommitment();
```
apoelstra, line 352: In d0c18edda0274dab7d8582f521e9bd86ccd85925:

This is okay, but I really think it'd make more conceptual sense to treat the individual fields individually, instead of taking away the CT discount unless every field is present and confidential.

src/consensus/validation.h
```diff
@@ -135,12 +135,6 @@ class ValidationState
 
         return m_reject_reason;
     }
-    bool Valid()
-    {
-        m_result = Result{};
-        m_mode = ModeState::M_VALID;
-        return true;
-    }
```
apoelstra: In 6f92462634b7856579b2014cc7cbc13c51657304:

In future can you edit/replace commits rather than changing the same code in multiple commits? It makes reviewing much easier.

src/policy/discount.h
apoelstra: Done in e24bc12bf572204f0b82eb360c78ce05e181dc33

src/policy/discount.h
apoelstra: Done in e24bc12bf572204f0b82eb360c78ce05e181dc33

src/chainparams.cpp
```diff
@@ -897,6 +897,7 @@ class CCustomParams : public CRegTestParams {
         consensus.start_p2wsh_script = args.GetIntArg("-con_start_p2wsh_script", consensus.start_p2wsh_script);
         accept_discount_ct = args.GetBoolArg("-acceptdiscountct", false);
         create_discount_ct = args.GetBoolArg("-creatediscountct", false);
+        if (create_discount_ct) accept_discount_ct = true;
```
apoelstra, line 900: In 7fc96512ff71c5fea9fb167403c64ac42af1070b:

This is really just a nit, but I think it'd be clearer if you changed `create_discount_ct` to be `accept_discount_ct || args.GetBoolArg("-creatediscountct", false)` or even just `args.GetBoolArg("-creatediscountct", accept_discount_ct);`.

The current code modifies variables twice when conceptually they should be immutable, which makes it easy for readers to miss the second assignment. Especially when it's hidden at the end of an `if` line.

apoelstra:
Done reviewing 7fc96512ff71c5fea9fb167403c64ac42af1070b. Just a couple minor things: would be good to change the RPC to not output `discountvsize`  when acceptdiscounttxes is off, and would be good to change the discount computation to individually treat the output fields.

The rest are nits.

Fine if you just add commits, since that's what's been happening with this PR so far, but in general I would prefer if you were rebasing.

src/core_write.cpp
delta1: done in 1ec3239366b2febe82b62aff23d0069897a5ee8c

src/primitives/transaction.h
delta1: done in a7e9d931ec9065b5da3c271660dc8f91b6c1bd29

src/consensus/validation.h
delta1: sorry about that, I thought it was better to add commits instead of force pushing 

will rebase in future

src/chainparams.cpp
delta1: done in 523c98d69ff416ec8e61a878b477b9f2c37171f8

apoelstra:
utACK 523c98d69ff416ec8e61a878b477b9f2c37171f8

src/policy/discount.h
```diff
@@ -24,15 +24,21 @@ static inline int64_t GetDiscountVirtualTransactionSize(const CTransaction& tx,
...
     for (size_t i = 0; i < tx.vout.size(); ++i) {
         const CTxOut& output = tx.vout[i];
-        if (output.IsConfidential()) {
-            // subtract the weight of the output witness
-            if (i < tx.witness.vtxoutwit.size()) {
-                weight -= ::GetSerializeSize(tx.witness.vtxoutwit[i], PROTOCOL_VERSION);
+        if (i < tx.witness.vtxoutwit.size()) {
+            if (output.nValue.IsCommitment()) {
+                // subtract the weight of the output value range proof
+                weight -= ::GetSerializeSize(tx.witness.vtxoutwit[i].vchRangeproof, PROTOCOL_VERSION);
```
roconnor-blockstream, line 30: Move the `if (i < tx.witness.vtxoutwit.size())` inside and only apply it to the `::GetSerializeSize(tx.witness.vtxoutwit[i].vchRangeproof, PROTOCOL_VERSION);` statement.

roconnor-blockstream: Scratch that.

You should go back and discount the entire `::GetSerializeSize(tx.witness.vtxoutwit[i], PROTOCOL_VERSION);` whenever `i < tx.witness.vtxoutwit.size()` independent of whether or not any assets or values are blinded.

At least that is what I think.  @apoelstra thoughts?

apoelstra: Yeah, it also occurred to me that it would be simpler to separate things out this way. So then each individual discount is applied exactly if it makes sense.

Though I *think* that in practice the CT rules mean that the existing logic is correct.

roconnor-blockstream: One issue with the current code is that if you have a transaction with only a confidential nonce, but it has no witness data, then that confidential nonce doesn't get a discount.  ... not that such transactions exist in practice, but still.

delta1: sorry - I did mean to have the nonce check outside of the witness check since it's unrelated. pushed a fixup commit for that. 

the rest of this thread is not clear to me. as it stands now, the range proof for value is discounted if the amount is blinded, and the surjection proof for asset is discounted if the asset is blinded. 

please specify any change I should make now?

apoelstra: >the rest of this thread is not clear to me. as it stands now, the range proof for value is discounted if the amount is blinded, and the surjection proof for asset is discounted if the asset is blinded.

This is fine, but you don't need to make it conditional. Just always discount the surjectionproof and rangeproof (as long as the vtxoutwit is big enough). Yes, it is true that these fields must be empty unless the value or asset respectively are confidential. But we don't need to rely on that fact here, so let's avoid relying on it.

delta1: Interestingly, discounting the witness unconditionally affects explicit transactions by reporting their `discountvsize` as 2 bytes smaller than their `vsize` (324 vs 326 in the test case tx). Presumably that's a serialized `0` for each of the two proofs.

apoelstra: Ah, yes. These fields have a serialized length byte even if they're unpopulated.

This means your code should have a `- 1` because you *shouldn't* be discounting it because it's present even in non-CT transactions. We found a bug :).

apoelstra:
In 3a4543986a74ab69fc464c054fc2871e2ae6fa11:

You still have the nAsset and nValue discounts conditioned on the witness being present. Why?

apoelstra:
So, I guess there are three cases for the witness field:

* It is literally not present (length 0)
* It is present but unpopulated (length 2, just two length-prefix bytes for the range and surjection proof)
* It is present and populated (length > 2).

What conditions separate the first and second cases?

delta1:
> In [3a45439](https://github.com/ElementsProject/elements/commit/3a4543986a74ab69fc464c054fc2871e2ae6fa11):
> 
> You still have the nAsset and nValue discounts conditioned on the witness being present. Why?

oversight, fixed 

> So, I guess there are three cases for the witness field:
> 
>     It is literally not present (length 0)
>     It is present but unpopulated (length 2, just two length-prefix bytes for the range and surjection proof)
>     It is present and populated (length > 2).
> 
>  What conditions separate the first and second cases? 

From what I understand case 1 does not occur, since even explicit transactions have the single `0` byte for each proof

src/policy/discount.h
```diff
@@ -25,20 +25,16 @@ static inline int64_t GetDiscountVirtualTransactionSize(const CTransaction& tx,
...
-            if (output.nAsset.IsCommitment()) {
-                // subtract the weight of the output asset surjection proof
-                weight -= ::GetSerializeSize(tx.witness.vtxoutwit[i].vchSurjectionproof, PROTOCOL_VERSION);
-            }
-            if (output.nNonce.IsCommitment()) {
-                // subtract the weight difference of nonce commitment (33) vs no nonce (1)
-                weight -= 32;
-            }
+            // subtract the weight of the output witness, except the 2 bytes used to serialize the empty proofs
+            weight -= (::GetSerializeSize(tx.witness.vtxoutwit[i], PROTOCOL_VERSION) - 2);
```
roconnor-blockstream, line 29: Should also condition on `2 <= ::GetSerializeSize(tx.witness.vtxoutwit[i], PROTOCOL_VERSION`.

apoelstra: We can just assert this IMO.

apoelstra: I guess we are supposed to use `CHECK` or something rather than assert.

delta1: added assert in a67b08fc51afa45374c60024262c9e09b1a21d99 

looks like the `CHECK` macro is specific to secp256k1 testing

apoelstra: I meant `CHECK_NONFATAL`. But actually I think here we really do want to `assert` because if we're wrong about this somehow it could be catastrophic.

apoelstra:
Your assertion is `>= 0` but it should be `>= 2`.

delta1:
> Your assertion is `>= 0` but it should be `>= 2`.

the 2 is subtracted in the line above

apoelstra:
>the 2 is subtracted in the line above

:grimacing: `GetSerializeSize` returns a `size_t`. I see now that you're implicitly casting it to a `in64_t`, which I suspect that the linter is failing on and is causing one of the red Xs in CI. This really makes me nervous. I don't like signed numbers, I don't like conversions from unsigned to (possibly differently sized) signed numbers in C++, and I don't like "sizes" that might be negative.

roconnor-blockstream:
> > the 2 is subtracted in the line above
> 
> 😬 `GetSerializeSize` returns a `size_t`. I see now that you're implicitly casting it to a `in64_t`, which I suspect that the linter is failing on and is causing one of the red Xs in CI. This really makes me nervous. I don't like signed numbers, I don't like conversions from unsigned to (possibly differently sized) signed numbers in C++, and I don't like "sizes" that might be negative.

When I was reviewing this, I noted that in upstream Bitcoin Core this all gets fixed up.  So I figured eventually a rebase will fix this stuff.

That said, we should probably pass CI.

delta1:
@apoelstra thanks for your feedback, updated. Initially had it as a size_t but the compiler was giving me a warning "comparison of unsigned expression in '>= 0' is always true". Changed now do the assert >= 2 before doing the subtraction.  

@roconnor-blockstream there are a couple of CI failures fixed by #1330

apoelstra:
Great, thanks! 4fb3706e6c6afc631a447920e6ec89c78ae1f310 looks good to me.

> @apoelstra thanks for your feedback, updated. Initially had it as a size_t but the compiler was giving me a warning "comparison of unsigned expression in '>= 0' is always true". Changed now do the assert >= 2 before doing the subtraction.

Yes, but your initial change was little help to a human reviewer who expects a size to be a `size_t` :) And it replaced a "always true comparison" with a "silent conversion between integer types of different signedness". Which for some reason, gcc does not warn about.

delta1:
> "silent conversion between integer types of different signedness" 

gotta love C++ :)

delta1:
@apoelstra can I squash this into 1 commit and rebase on master?

apoelstra:
@delta1 yeah, that'd be good. Though I would prefer if you could squash this into four or five commits. For example

* one that introduces the new config args, but don't use them to do anything
* one that introduces the new discount functions but doesn't use it
* one that hooks it up to the mempool logic
* one that hooks it up to the network logic
* one that hooks it up to the wallet logic

apoelstra:
utACK 83a7fe5442d9af612183b20d988d367a74ce11de

I can't test this locally because, even on master, even with `-j1`, two functional tests are failing in my Nix environment (`interface_bitcoin_cli --descriptors` and `mempool_persist`). But I confirmed that everything appears to build on the tip of this PR, and the code changes look good (and the total diff (nearly) matches the previous 20-commit thing).

roconnor-blockstream:
I have nothing to add.

delta1:
pushed to update liquidv1 default for accept_discount_ct to false 

```
git range-diff 7ff33d45df3a07b421cd7bacc63e9f34fef72fcb..3ebc35359f1304ff9f3a2a5286a154287794e23f 7ff33d45df3a07b421cd7bacc63e9f34fef72fcb..83a7fe5442d9af612183b20d988d367a74ce11de
1:  c1661b251a ! 1:  c2692d9907 discount: introduce config args
    @@ src/chainparams.cpp: public:
      
              multi_data_permitted = true;
     +        create_discount_ct = args.GetBoolArg("-creatediscountct", false);
    -+        accept_discount_ct = args.GetBoolArg("-acceptdiscountct", false);
    ++        accept_discount_ct = args.GetBoolArg("-acceptdiscountct", true);
      
              parentGenesisBlockHash = uint256S("000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f");
              const bool parent_genesis_is_null = parentGenesisBlockHash == uint256();
2:  9cbd3e0f05 = 2:  283d245bd9 discount: add policy/discount.h
3:  b7b2288a74 = 3:  76d42f5056 discount: implement mempool logic
4:  c85f0d4fda = 4:  146b1acc71 discount: implement net processing
5:  4807f3ef1f = 5:  c7dae295ba discount: change miner tx ordering
6:  52962c16c9 = 6:  b30b79d653 discount: allow wallet to create discount txs
7:  3ebc35359f = 7:  83a7fe5442 discount: add functional tests
```

PR #1326, opened 2024-04-16
simplicity: enable coverage in configure.ac

delta1:
- enables coverage for the simplicity build when elements is built with `--enable-lcov` 
- filters `src/simplicity/secp256k1` from results (since `src/secp256k1` is also filtered by upstream)  

example output: 

qa-assets from https://github.com/uncomputable/asset-gen thanks @uncomputable 

configured with: `./configure --enable-lcov --with-incompatible-bdb --with-seccomp=no` 

built with: `DIR_UNIT_TEST_DATA=/path/to/qa-assets/unit_test_data make -j15 cov` 

uploaded results here: https://simplicity-coverage.netlify.app/ 

image

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -208,6 +208,7 @@ LCOV_FILTER_PATTERN = \
 	-p "src/univalue" \
 	-p "src/crypto/ctaes" \
 	-p "src/secp256k1" \
+	-p "src/simplicity/secp256k1" \
 	-p "depends"
 
 DIR_FUZZ_SEED_CORPUS ?= qa-assets/fuzz_seed_corpus
--- a/configure.ac
+++ b/configure.ac
@@ -810,7 +810,7 @@ if test x$use_lcov = xyes; then
 
   AX_CHECK_LINK_FLAG([[--coverage]], [LDFLAGS="$LDFLAGS --coverage"],
     [AC_MSG_ERROR("lcov testing requested but --coverage linker flag does not work")])
-  AX_CHECK_COMPILE_FLAG([--coverage],[CXXFLAGS="$CXXFLAGS --coverage"],
+  AX_CHECK_COMPILE_FLAG([--coverage],[CXXFLAGS="$CXXFLAGS --coverage";CFLAGS="$CFLAGS --coverage"],
     [AC_MSG_ERROR("lcov testing requested but --coverage flag does not work")])
   CXXFLAGS="$CXXFLAGS -Og"
 fi


uncomputable:
What is the command to just run the QA asset test with coverage?

`DIR_UNIT_TEST_DATA=/path/to/qa-assets/unit_test_data make -j15 cov` runs the entire test suite on my end, which takes a lot longer.

delta1:
> What is the command to just run the QA asset test with coverage? 

I've been unable to do this successfully. 

What you can do is at least skip the functional tests by running: 

`DIR_UNIT_TEST_DATA=/path/to/qa-assets/unit_test_data make -j15 test_bitcoin.coverage/.dirstamp`

image

Makefile.am
```diff
@@ -208,6 +208,7 @@ LCOV_FILTER_PATTERN = \
 	-p "src/univalue" \
 	-p "src/crypto/ctaes" \
 	-p "src/secp256k1" \
+	-p "src/simplicity/secp256k1" \
```
roconnor-blockstream, line 211: I'm confused why this isn't just `"src/simplicity"`.   I'm also confused as to why you got coverage output for all the simplicity files, even tough it wasn't included in this filter.

roconnor-blockstream: Oh the filter "removes" these directories.  Got it.

roconnor-blockstream:
I have managed the following (hacky) set of commands to only run the QA asset test coverage.
This presumes the project was configured with `--enable-lcov`.

1. `make`
1. `DIR_UNIT_TEST_DATA=~/asset-gen/ ./src/test/test_bitcoin --log_level=warning --run_test=script_tests`
1. `lcov --gcov-tool $(pwd)/cov_tool_wrapper.sh -c -d src/ -t test_bitcoin -o test_bitcoin.info`
1. `make test_bitcoin.coverage/.dirstamp`

The results are in `test_bitcoin.coverage`.

That said, this process actually fails on my NixOS system due to some sort of NixOS specific issue (due to the non-standard location of boost I think) that I haven't entirely figured out yet, but is fairly easy to work around.

uncomputable:
@roconnor-blockstream To fix boost on NixOS, you can set an environment variable inside nix and use it for the configuration script:

```nix
configureFlags = [
    "--with-boost-libdir=${boost.out}/lib"
    "--disable-bench"
  ] +++ ...
```

```bash
./configure $configureFlags
```

roconnor-blockstream:
@uncomputable Unfortunately that isn't the issue.

The issue is that `clang --coverage` seems to get confused about the location of boost `.hpp` files when generating `.gcno` files and says they are located in `src/include/boost` (or rather `/home/roconnor/elementsd/src/include/boost`).  I strongly suspect this is because Nix passes the location of the boost header via `-isystem` rather than by `-I`.  Anyhow, the result is that building the coverage report complains loudly about missing boost header files before finally in the last step `genhtml` decides to fail because it cannot find these files.

PR #1332, opened 2024-05-06
simplicity: add fuzz target

delta1:
adds a new fuzz target for `elements_simplicity_execSimplicity`  

compile with instructions in `fuzzing.md` 

run with: `FUZZ=simplicity src/test/fuzz/fuzz`


delta1:
as this fuzz target is now, quickly found these 2 crashes and a leak 
[fuzz.zip](https://github.com/ElementsProject/elements/files/15221086/fuzz.zip)

apoelstra:
Do you know how to use these inputs?

apoelstra:
Unfortunately I can't test this because I cannot figure out how to get a `nix-shell` in which the fuzz tests are built in the same way as `nix-build` does. Somehow stepping through the `unpackPhase`/`autoreconfPhase`/`configurePhase`/`buildPhase` in a `nix-shell --pure` results in linker errors.

delta1:
> Do you know how to use these inputs?

The input can be passed as an arg 

```
> FUZZ=simplicity src/test/fuzz/fuzz ./crash-631f08c5ca034c7a8cb316a45ec1aaf855360cc9
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3819231386
INFO: Loaded 1 modules   (247138 inline 8-bit counters): 247138 [0x5c6c425dbdf3, 0x5c6c42618355), 
INFO: Loaded 1 PC tables (247138 PCs): 247138 [0x5c6c42618358,0x5c6c429dd978), 
src/test/fuzz/fuzz: Running 1 inputs 1 time(s) each.
Running: ./crash-631f08c5ca034c7a8cb316a45ec1aaf855360cc9
AddressSanitizer:DEADLYSIGNAL
=================================================================
==2179002==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x5c6c415b43b9 bp 0x000000000000 sp 0x7ffd9e2fec50 T0)
==2179002==The signal is caused by a READ memory access.
==2179002==Hint: address points to the zero page.
    #0 0x5c6c415b43b9 in ReadBE32 /home/byron/code/elements/src/simplicity/primitive/elements/../../sha256.h:38:8
    #1 0x5c6c415b43b9 in sha256_toMidstate /home/byron/code/elements/src/simplicity/primitive/elements/../../sha256.h:100:17
    #2 0x5c6c415b43b9 in elements_simplicity_mallocTapEnv /home/byron/code/elements/src/simplicity/primitive/elements/env.c:553:3
    #3 0x5c6c405a5e92 in simplicity_fuzz_target(Span<unsigned char const>) ./test/fuzz/simplicity.cpp:42:31
```

> Unfortunately I can't test this because I cannot figure out how to get a nix-shell in which the fuzz tests are built in the same way as nix-build does. Somehow stepping through the unpackPhase/autoreconfPhase/configurePhase/buildPhase in a nix-shell --pure results in linker errors.
 
I tried to get a shell.nix working with clangstdenv but couldn't get it to compile

delta1:
The other crash is in the same place as above, I'll look into it. 

Here's the leak: 

```
> FUZZ=simplicity src/test/fuzz/fuzz ./leak-581de5ff9fb360f3d8fc849bfbcc9d58f49d59d6 
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 52330119
INFO: Loaded 1 modules   (247138 inline 8-bit counters): 247138 [0x5cade94cbdf3, 0x5cade9508355), 
INFO: Loaded 1 PC tables (247138 PCs): 247138 [0x5cade9508358,0x5cade98cd978), 
src/test/fuzz/fuzz: Running 1 inputs 1 time(s) each.
Running: ./leak-581de5ff9fb360f3d8fc849bfbcc9d58f49d59d6
Executed ./leak-581de5ff9fb360f3d8fc849bfbcc9d58f49d59d6 in 0 ms
***
*** NOTE: fuzzing was not performed, you have only
***       executed the target code on a fixed set of inputs.
***

=================================================================
==2181494==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 176 byte(s) in 1 object(s) allocated from:
    #0 0x5cade71a748e in malloc (/home/byron/code/elements/src/test/fuzz/fuzz+0x10dc48e) (BuildId: 7babd80b5ee14c62698a09fa7526057de4e74e8a)
    #1 0x5cade84a42cb in elements_simplicity_mallocTapEnv /home/byron/code/elements/src/simplicity/primitive/elements/env.c:528:22

SUMMARY: AddressSanitizer: 176 byte(s) leaked in 1 allocation(s).
```

src/test/fuzz/simplicity.cpp
```diff
@@ -0,0 +1,53 @@
...
+            const uint_fast32_t ix = fuzzed_data_provider.ConsumeIntegral<uint_fast32_t>();
+
+            const std::vector<unsigned char> control = ConsumeRandomLengthByteVector(fuzzed_data_provider);
+            const std::vector<unsigned char> script_bytes = ConsumeRandomLengthByteVector(fuzzed_data_provider);
+
+            rawTapEnv simplicityRawTap;
+            simplicityRawTap.controlBlock = control.data();
+            simplicityRawTap.pathLen = (control.size() - TAPROOT_CONTROL_BASE_SIZE) / TAPROOT_CONTROL_NODE_SIZE;
+            simplicityRawTap.scriptCMR = script_bytes.data();
+            tapEnv* taproot = elements_simplicity_mallocTapEnv(&simplicityRawTap);
```
delta1, line 42: `control` and `script_bytes` are random lengths here, `elements_simplicity_mallocTapEnv` calls `sha256_toMidstate` with `scriptCMR` (script_bytes) as the hash which which has a precondition that this must have a size of 32, which I see in interpreter.cpp is checked 

https://github.com/ElementsProject/elements/blob/simplicity/src/simplicity/sha256.h#L97

roconnor-blockstream: I can update the documentation for `elements_simplicity_mallocTapEnv`.

Note: the documentation currently says that the control block must be of size `33+pathLen*32`.

roconnor-blockstream: I suppose technically the control block need only be _at least_ `33+pathLen*32` because for C, if arrays are longer than expected, the extra data is just ignored, which you are doing here.

delta1:
Updated to include the expected invariants for control block and script_bytes. 

Haven't found any crashes now, and submitted some seed data in https://github.com/ElementsProject/qa-assets/pull/2 

Run with: `FUZZ=simplicity src/test/fuzz/fuzz /path/to/qa-assets/fuzz_seed_corpus/simplicity`

src/test/fuzz/simplicity.cpp
delta1: Thanks @roconnor-blockstream, updated

src/test/fuzz/simplicity.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    simplicityRawTap.pathLen = (control.size() - TAPROOT_CONTROL_BASE_SIZE) / TAPROOT_CONTROL_NODE_SIZE;
+                    simplicityRawTap.scriptCMR = script_bytes.data();
+                    tapEnv* taproot = elements_simplicity_mallocTapEnv(&simplicityRawTap);
+
+                    const int64_t budget = fuzzed_data_provider.ConsumeIntegral<int64_t>();
+                    const std::vector<unsigned char> amr = fuzzed_data_provider.ConsumeBytes<unsigned char>(32);
+                    const std::vector<unsigned char> program = ConsumeRandomLengthByteVector(fuzzed_data_provider);
+
+                    simplicity_err error;
+                    elements_simplicity_execSimplicity(&error, imr.data(), tx, ix, taproot, genesisBlockHash.data(), budget, amr.data(), program.data(), program.size());
```
roconnor-blockstream, line 58: We will probably need to do something about this IMR field.  I don't think the fuzzer is going to figure out the hash of the program randomly by itself.  I'll look into it.

roconnor-blockstream: Sorry, actually it is the `scriptCMR` that is the problem.  This `imr` is actually an output field (and thus you don't need to fill it with random data).

roconnor-blockstream: One idea is to exclude this check when compiling with FUZZER defined mode.

https://github.com/BlockstreamResearch/simplicity/blob/195c5152683330cc34fa4b6d39008a1f6cf09ecb/C/primitive/elements/exec.c#L71-L75

Though the idea of disabling important checks doesn't thrill me.  There might be some safer variations of this idea.

roconnor-blockstream:
One thing we should verify is that the `PRODUCTION` flag is not enable during fuzzing to ensure all assertions are being run:

https://github.com/BlockstreamResearch/simplicity/blob/master/C/simplicity_assert.h#L12

If there is some sort of `FUZZ` define, we can add a check/warning to the simplicity code that `FUZZ` and `PRODUCTION` are not both set.

roconnor-blockstream:
@delta1 I don't know why, but

    PRINT_ALL_FUZZ_TARGETS_AND_ABORT= src/test/fuzz/fuzz

doesn't list "simplicity" for some reason.

delta1:
@roconnor-blockstream I saw the same issue, made a change in fuzz.cpp and recompiled, and now it seems to appear? Maybe `ccache` related, are you using that too?  

Here's my output with some lines removed for clarity 

```
> PRINT_ALL_FUZZ_TARGETS_AND_ABORT= src/test/fuzz/fuzz
addition_overflow
...
signet
simplicity
snapshotmetadata_deserialize
...
witness_program
fuzz: test/fuzz/fuzz.cpp:66: auto initialize()::(anonymous class)::operator()() const: Assertion `"!should_abort" && check' failed.
zsh: IOT instruction (core dumped)  PRINT_ALL_FUZZ_TARGETS_AND_ABORT= src/test/fuzz/fuzz
```  

The change I made before recompiling was: 

```diff
diff --git a/src/test/fuzz/fuzz.cpp b/src/test/fuzz/fuzz.cpp
index a33297e0ed..800bcd0354 100644
--- a/src/test/fuzz/fuzz.cpp
+++ b/src/test/fuzz/fuzz.cpp
@@ -27,6 +27,7 @@ std::map<std::string_view, std::tuple<TypeTestOneInput, TypeInitialize, TypeHidd
 
 void FuzzFrameworkRegisterTarget(std::string_view name, TypeTestOneInput target, TypeInitialize init, TypeHidden hidden)
 {
+    std::cout << "register: " << name << std::endl;
     const auto it_ins = FuzzTargets().try_emplace(name, std::move(target), std::move(init), hidden);
     Assert(it_ins.second);
 }
``` 

Which I then confirmed printed `register: simplicity` and then saw simplicity in the target list...

roconnor-blockstream:
Oops I was building #1326 instead of this PR.  My bad.

src/test/fuzz/simplicity.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+FUZZ_TARGET(simplicity)
+{
+    FuzzedDataProvider fuzzed_data_provider(buffer.data(), buffer.size());
+    {
+        const std::optional<CMutableTransaction> mtx_precomputed = ConsumeDeserializable<CMutableTransaction>(fuzzed_data_provider);
+        if (mtx_precomputed) {
+            const CTransaction tx_precomputed{*mtx_precomputed};
+            const PrecomputedTransactionData precomputed_transaction_data{tx_precomputed};
+            const transaction* tx = precomputed_transaction_data.m_simplicity_tx_data;
```
roconnor-blockstream, line 27: I've finally got the coverage analysis of fuzzing working, and I will share how to do that with you and everyone else shortly.

However, in the meantime, the results I have suggest to me that `tx` is always `NULL`, and thus the fuzzer doesn't even make it past the first line of `elements_simplicity_execSimplicity`

delta1: oh interesting. maybe if we provide a well formed transaction to the fuzzer via the corpus it will have a better starting point

roconnor-blockstream: You can probably go ahead and check the return value of  `elements_simplicity_execSimplicity` and assert that it is true.  `elements_simplicity_execSimplicity` only returns `false` during catastrophic failure, such as `malloc` failing, or, as in this case, if you have called it with (detectable) incorrect parameters.

The implementation actually does this assertion: https://github.com/ElementsProject/elements/blob/2423e8559ab4c25287c1ecd5add02119f7cc11c6/src/script/interpreter.cpp#L3115-L3117

delta1:
@roconnor-blockstream rebased

PR #1333, opened 2024-05-22
wallet: allow mintxfee=0

delta1:
Allow `-mintxfee=0`, so the elements wallet can create transactions with 0 value fee outputs. 

Elements can already accept 0 value fees with `-minrelaytxfee=0` and `-blockmintxfee=0`, but before this change the wallet would not create a transaction with 0 feerate.  

Note: this check for 0 was inherited from upstream and subsequently removed in https://github.com/bitcoin/bitcoin/commit/f11eb1fe279c4a92e1bfc2c139e8838c73459d12

### DIFF
--- a/test/functional/wallet_send_zero_fee.py
+++ b/test/functional/wallet_send_zero_fee.py
@@ -4,7 +4,6 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 from decimal import Decimal
 
-from test_framework.blocktools import COINBASE_MATURITY
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (
     assert_equal,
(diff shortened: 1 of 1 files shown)


tomt1664 (commented):
`HasValidFee` is a consensus check (in `CheckTxInputs`) which was introduced with CA. 

I'm not sure why the `fee == 0` was originally added, as a transaction without any fee output (i.e. also zero fee) is consensus valid. (the `fee == 0` check in only applied to individual fee outputs, not the total fees for the tx). 

I've verified that removing this `fee == 0` check will cause a hard-fork if a zero fee output tx is confirmed in a block. 

Suggestion: we add a `-con_` config parameter (e.g. `-con_allowanyfee`) that removes this optionally.

delta1:
> why the fee == 0 was originally added 

I think that part was inherited from upstream initially 

>  add a -con_ config parameter (e.g. -con_allowanyfee) that removes this optionally 

Good idea, thanks

PR #1342, opened 2024-07-17
Ensure that arm-macos environments have sufficient QT thread stack

jhfrontz:
Issuing assets via the elements-qt console interface on MacOS arm platforms causes the process to crash.  This appears to be because the secondary threads on MacOS default to a 512k stack (per https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Multithreading/CreatingThreads/CreatingThreads.html via https://stackoverflow.com/a/23609290 ) -- this in comparison to linux threads that apparently have 8M of stack.

This change forces the QT-related threads to have an 8M stack.


delta1 (approved):
utACK 9853992

grubles:
Tested ACK 9853992

PR #1346, opened 2024-07-24
discount: fix init doc for acceptdiscountct and override for creatediscountct

delta1:
The default for acceptdiscountct in liquidv1 was changed to false in #1317 without the corresponding doc being updated. 

Also fixes that creatediscountct should override the acceptdiscountct setting.


psgreco:
UTACK fdf3d1eab13346924f12f2a93ea5e5b00f36304d

PR #1355, opened 2024-08-14
fix: size estimation for pegin transactions

delta1:
Fixes the size estimation for pegin transactions, the witness scale factor was only being applied to the pubkey and not the rest of the witness data. 

The overestimation is ~56 bytes 

```
>>> (1+1+72+1+33/4)
83.25
>>> (1+1+72+1+33)/4
27.0
>>> 83.25 - 27.0
56.25
```


apoelstra (approved):
utACK c1bc7caff59f519bacada2e38dafe16ea34e4849

PR #1356, opened 2024-08-15
fix: use discountvsize in miner if discount enabled

delta1:
related to #1317 

Discount CT at the nominal `blockmintxfee` have a lower real feerate and are not included by the miner in the block template.

If `acceptdiscountct` is enabled, use the discountvsize of the package instead of vsize.


psgreco:
UTACK b0d6cde2d03c9dde8a0ab3715ba637fa83eb2195

PR #1362, opened 2024-09-09
fix: discountct package mining

delta1:
adds a reproducing test and fixes discount CT transaction package mining

re: #1317


jamesdorfman:
UTACK f107021869266f1f4481be05c42c2c45c0b612b1.

PR #1366, opened 2024-09-20
discount: fix weight calculation

delta1:
fixes the discount calculation by multiplying the amount and nonce discounts by the witness scaling factor (4) as they form part of the base transaction. thanks to @jgriffiths !

also adds `discountweight` next to `discountvsize` in TxToUniv response if node accepts Discount CT


apoelstra (approved):
utACK 09c5068ab19390fc6553172c645b72ee3d6fecd8

apoelstra (approved):
utACK 09c5068ab19390fc6553172c645b72ee3d6fecd8

jgriffiths:
tAck 09c5068ab19390fc6553172c645b72ee3d6fecd8

PR #1368, opened 2024-10-08
simplicity: define liquid testnet activation parameters

apoelstra:
For Liquid Testnet this just sets Simplicity to "always active", meaning that it will be enabled as soon as the blocksigner is updated. (And conversely, by standardness rules Simplicity programs will not be mined until they are active.)

Also cleans up some other parts of the activation logic and adds a functional test that should cover the way we do a real deployment.


psgreco:
UTACK 626e3aeafbeae4f065b1100c644d21b732f14234

Issue #1373, opened 2024-11-14
Change Liquid Bitcoin ticker from L-BTC to LBTC


jtrivinop:
The Liquid Federation has been discussing about changing the ticker of Liquid Bitcoin from L-BTC to LBTC. They voted and agreed to make this change. Many repos in the Liquid ecosystem for wallets, explorers, websites and other apps and services should start making this change, including Elements Qt

https://github.com/psgreco/elements/commit/7398cfae87e3f6dc71d7cef7fb5c36ad3748e685

PR #1374, opened 2024-11-17
Use LBTC instead of L-BTC in QT

psgreco:
Fixes: https://github.com/ElementsProject/elements/issues/1373


delta1 (approved):
utACK 7398cfae87e3f6dc71d7cef7fb5c36ad3748e685

PR #1376, opened 2024-12-16
Wallet: avoid scanning after the last known block

psgreco:
Related to https://github.com/bitcoin/bitcoin/issues/31474 upstream.

Fixes a crash in QT when loading an old wallet with many coinbase txs

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1689,6 +1689,11 @@ CWallet::ScanResult CWallet::ScanForWalletTransactions(const uint256& start_bloc
         if (max_height && block_height >= *max_height) {
             break;
         }
+        if (block_height >= WITH_LOCK(cs_wallet, return GetLastBlockHeight())) {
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -1689,6 +1689,11 @@ CWallet::ScanResult CWallet::ScanForWalletTransactions(const uint256& start_bloc
         if (max_height && block_height >= *max_height) {
             break;
         }
+        if (block_height > GetLastBlockHeight()) {
```
delta1, line 1692: ASAN CI: "wallet/wallet.cpp:1692:28: error: calling function 'GetLastBlockHeight' requires holding mutex 'cs_wallet' exclusively [-Werror,-Wthread-safety-analysis]"

delta1: added the lock for this, waiting on CI

delta1 (approved):
utACK a8c19d47097cd1ecddd35c5c9218de5fa54f4192 

new macos CI failure is unrelated to this PR

delta1:
reACK f138566

Issue #1233, opened 2023-04-17
Elements Connections can "Stall" on large bidirectional message exchange (eg in IBD).


lolhill:
**Summary of the conditions for the problem to occur and the outcome:**

1. A "large" message must be being sent by both peers to each other contemporaneously (this condition can occur in some circumstances in IBD). 
1. Where this occurs, both peers will "pause" receiving any socket buffer traffic until the full message is sent.
1. That message must be large enough to crash the TCP recive window on both sides (See Further Comments below). 
1. Receiving would only be resumed when the message is sent - This condition never occurs because both sides crash the recv window before the full message can be sent (either way).
1. The connection at this point is inactive and effectively "hung" until the inactivity timeout kicks in and the connection is tinned.
1. It's entirely possible that after the connections is tinned, further/new connections have exactly the same issue (Further Comments Below).
1. A quick review of the Bitcoin Core 22.1.1 code base suggests that, conceptually, the problem exists there. If the relevant conditions hit, the bug should hit there,  believe.

**Further Comments:**

1. "large" in this case isn't necessarily huge - depending on the network latency, TCP Windows Scaling and Windows values in play, a few hundred kilobytes should suffice.
1. Although the message is approximately 2Mb in Elements testing, only a few hundred kb are actually sent before the Windows are closed (see the attached tcpdump). TCP Window sizes are sub-100kb.
1. The application/message command line maxsendbuffer/maxrecvbuffer settings (and the related fPauseSend and fPauseRecv flags) are irrelevant to this issue - They can be set large, on both sides, and the problem will still occur. Setting them is not a workaround.

**A bit of extra detail/markers:**

1. A `netstat -anp` will reveal large, "stuck" Recv and Send Queues for the connection.
1. A tcpdump will show TCP Zero Windows on both sides of the connection - ie confirming that nothing is being pulled off the receive buffers (See attached example).
1. Testing simply requires hitting the conditions described above. Factors that may make the issue more likely are:

	- Increasing the network latency of the connection (eg using netem).
	- Reducing the TCP Window Scale Factor (or disabling TCP Window Scaling).

**Expected behavior**
The viable TCP Connection is maintained (data is read from the socket receive buffer on at least one side of the connection to relieve the stall)

**Actual behavior**
The connection is "stalled" until the inactivity timeout is reached, when the connection is closed (but on re-establishment with the same client, it's entirely likely the issue may recur - See to reproduce below).

**To reproduce**
[...]

PR #1382, opened 2025-01-03
Continue receiving data for the current messages, even when we select…

psgreco:
https://github.com/ElementsProject/elements/issues/1233 was fixed upstream, but it's still happening in 23.x. This fix intends to be much more "surgical" since it's only selecting recv for incomplete messages. Upstream's fix always selects recv now.

Needs to be backported to 23.2.x too


delta1 (approved):
utACK cd306bf8d3a0e5d2a7ab78020c417346a071acdc

PR #1384, opened 2025-01-20
simplicity: add fuzz target

apoelstra:
This PR adds a new fuzz target. I will post a separate PR to the qa-assets repo to add an initial seed corpus later today, once I am done combining and minimizing the results from my various local runs.

This fuzz target takes its seeds in a simple and well-defined format: a four-byte LE budget, then a transaction, Simplicity program and witness, each prefixed by a four-byte LE length. The fuzz target extracts any additional randomness it needs from the txid of the first input of the transaction, since this data is not interpreted in any other way we therefore won't confuse the fuzzer.

The reason for this design, rather than a more typical "just query the fuzzer when you need stuff", is to make it possible to fairly easily generate test vectors from sources other than this fuzz test. (For example, I have an alternate target which uses Rust code to generate well-formed Simplicity programs, which quickly gets high coverage at the expense of being an unmaintainable mess.)

### DIFF
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -312,6 +312,7 @@ test_fuzz_fuzz_SOURCES = \
  test/fuzz/secp256k1_ecdsa_signature_parse_der_lax.cpp \
  test/fuzz/signature_checker.cpp \
  test/fuzz/signet.cpp \
+ test/fuzz/simplicity.cpp \
  test/fuzz/socks5.cpp \
  test/fuzz/span.cpp \
  test/fuzz/spanparsing.cpp \
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -315,6 +315,7 @@ test_fuzz_fuzz_SOURCES = \
  test/fuzz/signet.cpp \
  test/fuzz/simplicity_compute_amr.c \
  test/fuzz/simplicity.cpp \
+ test/fuzz/simplicity_tx.cpp \
  test/fuzz/socks5.cpp \
  test/fuzz/span.cpp \
  test/fuzz/spanparsing.cpp \
--- a/src/Makefile.test.include
+++ b/src/Makefile.test.include
@@ -219,6 +219,7 @@ endif
 if ENABLE_FUZZ_BINARY
 test_fuzz_fuzz_CPPFLAGS = $(AM_CPPFLAGS) $(BITCOIN_INCLUDES)
 test_fuzz_fuzz_CXXFLAGS = $(AM_CXXFLAGS) $(PIE_FLAGS)
+test_fuzz_fuzz_CFLAGS = $(AM_CFLAGS) $(PIE_FLAGS)
 test_fuzz_fuzz_LDADD = $(FUZZ_SUITE_LD_COMMON)
 test_fuzz_fuzz_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS) $(PTHREAD_FLAGS) $(RUNTIME_LDFLAGS)
 test_fuzz_fuzz_SOURCES = \
@@ -312,6 +313,7 @@ test_fuzz_fuzz_SOURCES = \
  test/fuzz/secp256k1_ecdsa_signature_parse_der_lax.cpp \
  test/fuzz/signature_checker.cpp \
  test/fuzz/signet.cpp \
+ test/fuzz/simplicity_compute_amr.c \
  test/fuzz/simplicity.cpp \
  test/fuzz/socks5.cpp \
  test/fuzz/span.cpp \
--- a/src/test/fuzz/simplicity.cpp
+++ b/src/test/fuzz/simplicity.cpp
@@ -18,15 +18,15 @@ extern "C" {
 #include <string>
 #include <vector>
 
-uint256 GENESIS_HASH;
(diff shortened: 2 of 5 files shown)


src/Makefile.test.include
```diff
@@ -219,6 +219,7 @@ endif
 if ENABLE_FUZZ_BINARY
 test_fuzz_fuzz_CPPFLAGS = $(AM_CPPFLAGS) $(BITCOIN_INCLUDES)
 test_fuzz_fuzz_CXXFLAGS = $(AM_CXXFLAGS) $(PIE_FLAGS)
+test_fuzz_fuzz_CFLAGS = $(PIE_FLAGS)
```
roconnor-blockstream, line 222: Presumably we should do some cargo culting by adding

    test_fuzz_fuzz_CFLAGS = $(AM_CFLAGS) $(PIE_FLAGS)

apoelstra: Ah, yeah, good idea. I will copy this on the next force-push.

apoelstra:
I believe the remaining CI failures are all "not my fault" . The CentOS and macOS ones don't work at all due to missing OS images. The TSan one seems te be exhibiting some sort of kernel bug https://stackoverflow.com/questions/77850769/fatal-threadsanitizer-unexpected-memory-mapping-when-running-on-linux-kernels and the Win64 one is complaining about missing link symbols, but in bitcoin-qt which I didn't touch.

roconnor-blockstream:
I recommend you rebase on `master`.

apoelstra:
This is based on `master`.

delta1:
Yes master has the same CI failures

delta1 (approved):
utACK 9437bee648e18cd4a7ee00ee49ef44b9134570d3

src/test/fuzz/simplicity.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+uint256 GENESIS_HASH;
+
+CConfidentialAsset INPUT_ASSET_UNCONF{};
+CConfidentialAsset INPUT_ASSET_CONF{};
+CConfidentialValue INPUT_VALUE_UNCONF{};
+CConfidentialValue INPUT_VALUE_CONF{};
+CScript TAPROOT_SCRIPT_PUB_KEY{};
+std::vector<unsigned char> TAPROOT_CONTROL{};
+std::vector<unsigned char> TAPROOT_ANNEX{99, 0x50};
```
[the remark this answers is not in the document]
apoelstra: Fixed.

apoelstra:
Rebased on #1390. This now adds two fuzztests (which are both needed in tandem to get high coverage) -- I will PR to qa-assets to add seeds for the new one.

src/test/fuzz/simplicity_tx.cpp
```diff
@@ -0,0 +1,236 @@
...
+                        uint256 merkle_root = ComputeTaprootMerkleRoot(control, tapleaf_hash);
+                        auto ret = internal.CreateTapTweak(&merkle_root);
+                        if (ret.has_value()) {
+                            expect_simplicity = true;
+                            //assert(0); // useful for searching for a nontrivial fuzz target
+                            // Just drop the parity; it needs to match the one in the control block,
+                            // but we want to test that logic, so we allow them not to match.
+                            const XOnlyPubKey output_key = ret->first;
+                            if (current[top - 2].size() == 32) {
+                                // FIXME remove this check when we stop using Rust
```
apoelstra, line 179: In 09de430f956c32af6d010bd52b7a11e9b9e5e51e -- should remove this check

src/test/fuzz/simplicity_tx.cpp
```diff
@@ -0,0 +1,236 @@
...
+                const auto& program = current[top - 3];
+
+                if (control.size() >= TAPROOT_CONTROL_BASE_SIZE && (control[0] & 0xfe) == 0xbe) {
+                    // The fuzzer won't be able to produce a valid CMR on its own, so we compute it
+                    // and jam it into the witness stack. But we do require the fuzzer give us a
+                    // place to put it, so we don't have to resize the stack (and so that actual
+                    // valid transactions will work with this code).
+                    // Compute CMR and do some sanity checks on it (and the program)
+                    std::vector<unsigned char> cmr(32, 0);
+                    assert(cmr.size() == 32); // fuck C++
```
apoelstra, line 163: In 09de430f956c32af6d010bd52b7a11e9b9e5e51e: lol should probably remove this too.

apoelstra:
Tried an alternate solution.

delta1 (approved):
ACK 67942f5ac65f07c1d122b73528e97a5cc7b3da90

tested the new simplicity fuzz target for a few minutes with elementsproject/qa-assets#5 corpus

apoelstra:
Will keep adding suppressions and rebasing -- FYI these are happening because of https://github.com/ElementsProject/qa-assets/pull/5 which adds significantly more coverage to our CI. (I didn't realize that we were, in fact, running these seeds there.)

apoelstra:
Figured out how to reproduce these locally. Gonna back off this and open a new PR since there are quite a few.

apoelstra:
Rebased on master. Should be good to go now!

delta1 (approved):
ACK bd852baca64c206d3fbe6e4d32d8ab2d7167c1dc

PR #1385, opened 2025-01-23
assets: allow issuance and transaction amounts of more than MAX_MONEY for non-policy assets

delta1:
allows issuing/reissuing/sending/receiving of more than 21 million of non-policy assets 

adds a functional test


psgreco:
ACK e7f32aac75dadedf6e49068fede108ac6c8f4672

PR #1392, opened 2025-02-06
Fix some functional tests when bdb is not available

psgreco:
Several tests are failing in the mac native builds because they only have descriptor wallets, peg-ins/peg-outs/initialfreecoins are not available yet for descriptor wallets so we need to disable them.

Also the 21M test fails in descriptor wallets but for different reasons, the wallets have different names


delta1 (approved):
utACK e6aa6e773e464eb4e9ebca741dda5c3fc7299bfa

PR #1393, opened 2025-02-07
CI: Fix Win64 native build, TSAN and CentOS 8 test

psgreco:
This MR fixes 3 things, but there are some still missing.
Fixed:
1.  Win64 native build
2. CentOS 8 test (moved to Rocky Linux 8)
3. TSAN test (moved to clang-18 in an updated image, test is still flaky due to parallelism, works well with `-j1`)

Still missing:
1. macOS native build, will fix after the move to arm64
2. Stabilize TSAN, either find the issue or move to `-j1`)
3. Windows native tests are still failing

PR #1394, opened 2025-02-09
Prepare 23.2.6 final

psgreco:
Includes the version bump and all the fixes from master up to https://github.com/ElementsProject/elements/commit/29cf7c21b1fa82da68319869d510f3bb36f0a22d but with simplicity reverted, to keep 23.2.x without it (only test fixes)


jsarenik:
What about four failing CI tasks?

Screenshot_20250209-090629

delta1:
@jsarenik those are unrelated to this PR and mostly fixed by #1393 

utACK 320b49d

delta1 (approved):

psgreco:
@jsarenik a bit more detail on what Byron explained

TSAN fails due to a change in the runners, clang-13 doesn't like `vm.mmap_rnd_bits=32` that's set in the runners, works locally without issues. #1393 will upgrade clang to 18, which has the bug fixed.
CentOS 8 fails due to the release being pulled from the repos, so the updates fail. Moving it to Rocky 8 will keep the same environment and make it work again.

The other 2 failing tests are related to native builds, which are not used for releases, they are both for testing debug environments. Still we're working on making them work again, but nothing that affects the release.

PR #1397, opened 2025-02-10
[23.3.x] Fix macos native test

psgreco:


psgreco:
Fuzzer failures are unrelated, will be fixed as a part of #1399 #1401

PR #1399, opened 2025-02-11
various bugfixes for new fuzz test vectors

apoelstra:
Mostly straightforward stuff in here, except that we needed to backport https://github.com/bitcoin/bitcoin/pull/23418, which in turn required https://github.com/bitcoin/bitcoin/pull/24625.

This does not fix the PSET RPC-related crash but it does fix various ubsan lints (and a couple actual crashes) for other parts of the codebase.


delta1 (approved):
ACK 363c5101c235d700a92f8963b44118b6edc813e6 

ran functional tests locally

Issue #1398, opened 2025-02-10
elements-23.2.6 build error with gcc 15.0.1


grubles:
```
CXX      elementsd-bitcoind.o                                                                
In file included from ./chainparams.h:9,                                                       
                 from bitcoind.cpp:10:                                                         
./chainparamsbase.h:34:5: error: ‘uint16_t’ does not name a type                                                                                                                              
   34 |     uint16_t RPCPort() const { return m_rpc_port; }                                                                                                                                   
      |     ^~~~~~~~                                                                                                                                                                          
./chainparamsbase.h:10:1: note: ‘uint16_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
    9 | #include <string>                                                                                                                                                                     
  +++ |+#include <cstdint>                                                                                                                                                                    
   10 |                                                                                        
./chainparamsbase.h:35:5: error: ‘uint16_t’ does not name a type                                                                                                                              
   35 |     uint16_t OnionServiceTargetPort() const { return m_onion_service_target_port; }                                                                                                   
      |     ^~~~~~~~                                                                                                                                                                          
./chainparamsbase.h:35:5: note: ‘uint16_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
./chainparamsbase.h:39:51: error: ‘uint16_t’ has not been declared
   39 |     CBaseChainParams(const std::string& data_dir, uint16_t rpc_port, uint16_t mainchain_rpc_port, uint16_t onion_service_target_port)
      |                                                   ^~~~~~~~
./chainparamsbase.h:39:51: note: ‘uint16_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’                                                         ./chainparamsbase.h:39:70: error: ‘uint16_t’ has not been declared
   39 |     CBaseChainParams(const std::string& data_dir, uint16_t rpc_port, uint16_t mainchain_rpc_port, uint16_t onion_service_target_port)                                                       |                                                                      ^~~~~~~~
./chainparamsbase.h:39:70: note: ‘uint16_t’ is defined in header ‘<cstdint>’; this is probably fixable by adding ‘#include <cstdint>’
./chainparamsbase.h:39:99: error: ‘uint16_t’ has not been declared
   39 |     CBaseChainParams(const std::string& data_dir, uint16_t rpc_port, uint16_t mainchain_rpc_port, uint16_t onion_service_target_port)
[...]

PR #1400, opened 2025-02-11
Fix build in master with newer environments

psgreco:
Cherry pick some commits from 23.x to fix master build in newer environments, also fixes the build with gcc-15

Fixes: https://github.com/ElementsProject/elements/issues/1398


apoelstra:
ACK 04b114c148c7f86fdf1efefd6083a7f0ab081516

Tested with my NixOS setup.

PR #1401, opened 2025-02-11
more fuzzer fixes (continuation of #1399)

apoelstra:
Some combination of rebasing and the fact that MarcoFalke is clearly grinding his commit hashes to all start with `fa` led to these being dropped in #1399.


delta1 (approved):
utACK 56eec70de45a1437d360b47cfa316a05dd715929

PR #1413, opened 2025-02-13
lint: fix shellcheck, spelling, and locale

delta1:
Should fix lint ci job that I broke when merging #1210


psgreco:
UTACK 2f1b86e84a6087be5aa4bd0f61d3c2f44a60abd5

PR #1414, opened 2025-02-13
fuzz: minor updates to the Simplicity fuzz tests

apoelstra:
There was a bit of dead code laying around, and also we weren't hitting the `ERR_CMR` error return of the Simplicity interpreter. The former is easy and the latter I was able to hit with a little bit of trickery, without duplicating any work or requiring any new test seeds.

### DIFF
--- a/src/test/fuzz/simplicity_tx.cpp
+++ b/src/test/fuzz/simplicity_tx.cpp
@@ -2,7 +2,6 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
-#include <cstdio>
 #include <span.h>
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
LGTM.  This line now seems to be covered by the fuzzer.

https://github.com/ElementsProject/elements/blob/0397d22941289bf0a2d00418c9179eb8001e95ef/src/script/interpreter.cpp#L3149

roconnor-blockstream (approved):
LGTM.  This line now seems to be covered by the fuzzer.

https://github.com/ElementsProject/elements/blob/0397d22941289bf0a2d00418c9179eb8001e95ef/src/script/interpreter.cpp#L3149

Issue #1406, opened 2025-02-11
23.3.0rc3: `feature_discount_ct.py` test failure: AssertionError: not(10299 == 10300)


whitslack:
```
./configure \
	--prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu \
	--mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share \
	--sysconfdir=/etc --localstatedir=/var/lib --datarootdir=/usr/share \
	--disable-dependency-tracking --disable-silent-rules --disable-static \
	--docdir=/usr/share/doc/elements-23.3.0_rc3 \
	--htmldir=/usr/share/doc/elements-23.3.0_rc3/html --with-sysroot=/ \
	--libdir=/usr/lib64 --disable-static --enable-wallet --with-sqlite \
	--with-bdb --disable-usdt --without-miniupnpc --without-natpmp \
	--enable-tests --disable-bench --disable-fuzz --disable-fuzz-binary \
	--without-qrencode --disable-ccache --enable-asm --disable-zmq \
	--enable-man --enable-external-signer --with-utils --enable-util-cli \
	--enable-util-tx --enable-util-wallet --disable-util-util \
	--without-seccomp --without-libs --with-daemon --without-gui \
	--without-qtdbus --with-system-libsecp256k1
```

```
2025-02-11T20:44:06.128000Z TestFramework (DEBUG): PRNG seed is: 8755724000128018475
2025-02-11T20:44:06.128000Z TestFramework (DEBUG): Setting up network thread
2025-02-11T20:44:06.129000Z TestFramework (INFO): Initializing test directory /var/tmp/portage/net-p2p/elements-23.3.0_rc3/temp/test_runner_₿_🏃_20250211_200814/feature_discount_ct_167
2025-02-11T20:44:06.140000Z TestFramework.node0 (DEBUG): bitcoind started, waiting for RPC to come up
2025-02-11T20:44:06.159000Z TestFramework.node1 (DEBUG): bitcoind started, waiting for RPC to come up
2025-02-11T20:44:06.190000Z TestFramework.node2 (DEBUG): bitcoind started, waiting for RPC to come up
2025-02-11T20:44:06.983000Z TestFramework.node0 (DEBUG): RPC successfully started
2025-02-11T20:44:06.986000Z TestFramework.node1 (DEBUG): RPC successfully started
[...]

PR #1421, opened 2025-02-20
Fix tests and build warnings

psgreco:
In feature_discounct, the created tx can be of 2 different weights, we need to make sure we don't fail erroneously.

Newer gcc versions complain about mixing int and enum, so we need to do the 0 from the enum.

TSAN revealed an unlocked access to pindexBestHeader, so it's moved inside the lock

Fixes an issue in windows builds accessing the cookie file with special characters in the path.

Also addresses some issues seen in Win 64 native functional tests, doesn't fully fix the issues, but it's a step in the right direction.

Fixes: https://github.com/ElementsProject/elements/issues/1406 https://github.com/ElementsProject/elements/issues/1422


delta1 (approved):
utACK b92a3ceda298737aa1ec217c26b67280ad2944bc

PR #1424, opened 2025-03-03
Avoid double free in simplicity pointer when copying PrecomputedTransactionData

psgreco:
Fixes the following tests
```
feature_block_v4.py                                 | ✓ Passed  | 3 s
feature_fedpeg.py --legacy-wallet                   | ✓ Passed  | 78 s
feature_fedpeg.py --post_transition --legacy-wallet | ✓ Passed  | 74 s
feature_fedpeg.py --pre_transition --legacy-wallet  | ✓ Passed  | 80 s
feature_nulldummy.py --descriptors                  | ✓ Passed  | 4 s
feature_nulldummy.py --legacy-wallet                | ✓ Passed  | 5 s
feature_sighash_rangeproof.py --legacy-wallet       | ✓ Passed  | 143 s
rpc_fundrawtransaction.py --descriptors             | ✓ Passed  | 971 s
rpc_fundrawtransaction.py --legacy-wallet           | ✓ Passed  | 985 s
rpc_psbt.py --descriptors                           | ✓ Passed  | 216 s
rpc_psbt.py --legacy-wallet                         | ✓ Passed  | 239 s
wallet_bumpfee.py --descriptors                     | ✓ Passed  | 160 s
wallet_bumpfee.py --legacy-wallet                   | ✓ Passed  | 106 s
wallet_keypool.py --descriptors                     | ✓ Passed  | 17 s
wallet_keypool.py --legacy-wallet                   | ✓ Passed  | 10 s
wallet_multisig_descriptor_psbt.py                  | ✓ Passed  | 19 s
wallet_send.py --descriptors                        | ✓ Passed  | 37 s
wallet_send.py --legacy-wallet                      | ✓ Passed  | 36 s
wallet_taproot.py                                   | ✓ Passed  | 150 s
wallet_timelock.py                                  | ✓ Passed  | 2 s
wallet_watchonly.py --usecli --legacy-wallet        | ✓ Passed  | 6 s
```

There are still 2 failing tests, `feature_fee_estimation` fails in all arches and the problem is the test itself, needs to be fixed for elements. `feature_trim_headers` needs some backports from 23.x, but it is working in the release branches

PR #1425, opened 2025-03-03
Avoid double free in simplicity pointer when copying PrecomputedTrans…

roconnor-blockstream:
This is an alternative to #1424.

Arguably it would be better to make a class whose constructor calls `simplicity_elements_mallocTransaction`, but that might be over-engineered. (It would be nice to just inherit from `std::unique_ptr`, but apparently it doesn't have a virtual destructor so that is considered a no-no).


psgreco:
ACK 3a80782963e3d7bd5b5b6acdd4236bdc75ddadbb

psgreco (approved):

apoelstra:
utACK 3a80782963e3d7bd5b5b6acdd4236bdc75ddadbb

This looks great! Way cleaner.

PR #1427, opened 2025-03-04
simplicity: activate on first week of unanimous signalling after block 3333333

apoelstra:
This patch simply starts signalling for Simplicity in any blocksigner that has it enabled. Once we have one week (10080 blocks) of unanimous signalling, then it activates.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -1291,7 +1291,7 @@ public:
 
         // Simplicity
         consensus.vDeployments[Consensus::DEPLOYMENT_SIMPLICITY].bit = 21;
-        consensus.vDeployments[Consensus::DEPLOYMENT_SIMPLICITY].nStartTime = Consensus::BIP9Deployment::NEVER_ACTIVE;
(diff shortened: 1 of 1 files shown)


delta1 (approved):
ACK fa02ff7759ba8cbfd3ff9bfd398e7b2bb6826c08 

confirmed the timestamp/date, and that it's for LiquidV1 params 

image

apoelstra:
Rebased. (My diff is identical to Pablo's rebase, but mine has a gpg signature from me on it.)

delta1:
Just remembered that we interpret nStartTime as a block height, not a timestamp 

https://github.com/ElementsProject/elements/blob/246a764f23e597f9ae0cb136ac19a0597bff9430/src/consensus/params.h#L51

delta1:
We’ve just passed [block 3333333](https://liquid.network/liquid/block/23ed13f8ba88f8e055404047d0928489ae53d63917cc92eeb933c2a9c7f32f9a) yesterday which would make a neat/memorable start height

apoelstra:
Oof, good catch!

delta1 (approved):
reACK 308a3932395ebbda15fcfa5e6b8750ff82865ab7

PR #1428, opened 2025-03-04
wallet: Show full pak entry in getwalletpakinfo

psgreco:


src/wallet/rpc/backup.cpp
```diff
@@ -1987,6 +2001,28 @@ RPCHelpMan getwalletpakinfo()
...
+        if (secp256k1_ret != 1) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "bitcoin_descriptor could not be parsed.");
+        }
+
+
+        // Negate the pubkey
+        secp256k1_ret = secp256k1_ec_pubkey_negate(secp256k1_ctx, &masterpub_secp);
+
+        std::vector<unsigned char> negatedpubkeybytes;
+        negatedpubkeybytes.resize(33);
```
delta1, line 2017: I think I'd prefer this as a 33 byte array and skip the vec resize, but that's minor

psgreco:
Since this almost a direct copy of how it is done in `initpegoutwallet`, we're going to keep it this way and them change both places at the same time

delta1 (approved):
utACK c878b2587bd25e8f5b9120db4a26c78a7d991fae

PR #1433, opened 2025-03-10
feat: reduce default dust relay fee

kilrau:
This PR reduces `DUST_RELAY_TX_FEE` to `100`, matching [`DEFAULT_MIN_RELAY_TX_FEE`](https://github.com/ElementsProject/elements/blob/master/src/validation.h#L59), allowing elements wallet to create transactions with amounts as low as 17 sats, see https://github.com/ElementsProject/ELIPs/pull/24#issuecomment-2709819278.

PR #1436, opened 2025-03-19
test: add dust relay output value test

delta1:
re #1433

adds a functional test showing dust threshold values for blinded outputs using Elements RPCs


psgreco:
UTACK 8cde3cd30879369fdcb7b416b9450356936a9989

PR #1437, opened 2025-03-24
docs: update security.md

delta1:
Updates the security information and PGP key


jsarenik:
The change itself looks good to me but I have a note.

Who is `security@blockstream.com`? How does it sign a message? Do more people have access to its private key?

psgreco:
> The change itself looks good to me but I have a note.
> 
> Who is `security@blockstream.com`? How does it sign a message? Do more people have access to its private key?

It's an alias for several people inside the company, related to the security team

psgreco:
ACK 2ec75af8d320aa2919a79c412305d307c15ef04d

PR #1438, opened 2025-03-24
fix: simplicity activation when dynafed is active

delta1:
When DynaFed is active, trying to activate Simplicity with version bit 24 causes [`ContextualCheckDynaFedHeader`](https://github.com/ElementsProject/elements/blob/b63f7ab9b5e710fcd4e87f92c1981ea5f9c71bd1/src/validation.cpp#L3709) to fail with `CreateNewBlock: TestBlockValidity failed: invalid-dyna-fed, dynamic block header has unknown HF extension bits set`.


apoelstra:
utack

jsarenik:
Thanks! I will watch out for this error in my elements node running v23.3.0rc5

apoelstra:
utack changing the bit to 21

PR #1439, opened 2025-03-26
Add missing proof checks and tests

tomt1664:
Add checks that surjection proof and range proof are present when tx is confidential, and if not return error. 

Additional functional tests to verify correct behaviour.


delta1 (approved):
utACK 51f81f32863be31d1e5baf4ef2317322c84afeaf

PR #1444, opened 2025-04-02
Fix: Issueasset tx with blind parameter false is not accepted after t…

### LINKED ISSUE #1443: Issueasset tx with blind parameter false is not accepted after the change for more than 21M
Discovered this issue while merging recent RC changes on Sequentia, concretely 23.3.0 RC2 (#1338) . Our issueasset is by default unblinded and `wallet_elements_21million.py` was failing. While debugging the issue I've found that `MoneyRange()` check in confidential_validation.cpp wasn't adapted to do it only for policy asset but this check is done only in case that amount is explicit, so the transaction was rejected on broadcast.

Mixa84:
Fixes #1443

The fix is done by passing the consensus params to `CheckTxInputs(), VerifyAmounts() and VerifyIssuanceAmount()` in order to have the pegged asset in the last one. I thought to pass the whole params if anything else will be needed later, but it could also be to pass pegged_asset only to `VerifyAmounts() and VerifyIssuanceAmount`.

Adapted the test also to check for unblinded issuance and that it is working correctly.

Appreciate the review and feel free to suggest any different approach if this one is not suitable.

@delta1 @psgreco


delta1:
It is a bit less invasive to just get the params where we need them as in #1445

PR #1445, opened 2025-04-03
fix: unblinded re/issuance for non-policy asset greater than 21 million

delta1:
- test: add test for unblinded re/issuance greater than 21 million
- fix: unblinded re/issuance for greater than 21 million

### DIFF
--- a/src/confidential_validation.cpp
+++ b/src/confidential_validation.cpp
@@ -1,4 +1,5 @@
 
+#include <chainparams.h>
 #include <confidential_validation.h>
 #include <issuance.h>
 #include <pegins.h>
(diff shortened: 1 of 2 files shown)


delta1:
closes #1443

tomt1664:
utACK d3502ce66465c2eb1e4ba8c22d8d669ab9739d7a

apoelstra:
utACK d3502ce66465c2eb1e4ba8c22d8d669ab9739d7a -- though this is a hardforking change and we need to be careful about where/how we deploy it.

tomt1664:
Updated to enable this with a consensus parameter (`-con_allow_any_issuance`) to make the feature available to other applications without affecting existing chains.

PR #1446, opened 2025-04-12
Add liquid.network DNS seed

wiz:
We operate a cluster of about ~50 liquid and liquid testnet nodes globally for the liquid.network explorer. The liquid.network hostname is a round-robin DNS for our BGP anycast cluster of nodes from Ashburn, Frankfurt, Tokyo, Singapore, and Honolulu. You can see the full list at https://liquid.network/monitoring

Since we now accept incoming connections for Liquid P2P on 7042/tcp and for Liquid Testnet P2P on 18891/tcp, this PR adds liquid.network as a DNS seed for both networks, which should hopefully improve the P2P connectivity.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -1058,6 +1058,7 @@ public:
         if (!args.IsArgSet("-seednode")) {
             vSeeds.emplace_back("seed.liquid-testnet.blockstream.com");
             vSeeds.emplace_back("seed.liquidtestnet.com");
+            vSeeds.emplace_back("liquid.network.");
         }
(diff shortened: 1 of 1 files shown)


psgreco:
ACK 5d82808b617db647cb995f1f88fc61b69a5a6b80, both nodes are responding.

delta1 (approved):
ACK https://github.com/ElementsProject/elements/commit/5d82808b617db647cb995f1f88fc61b69a5a6b80

delta1:
Thanks @wiz!

PR #1450, opened 2025-04-21
syscall: whitelist network access for msghand thread

apoelstra:
Unlike in Bitcoin, our message handling thread needs to open sockets in order to call out to the mainchain RPC interface. We recently brought in a seccomp syscall whitelist from upstream, which is enabled on my local CI box (though apparently not on Github CI) and it is failing on this.


delta1:
Thanks @apoelstra, confirmed I can now run the funcitonal tests with seccomp enabled. 

Tested ACK 12152bf

Issue #1460, opened 2025-06-18
Use 1/0 instead of true/false in help to avoid confusion


psgreco:
Currently there are places in the command line/options help where the options are mentioned as true/false, but they are actually 1/0.

PR #1461, opened 2025-06-18
Use 1/0 instead of bool in args help, so it matches real parsing

psgreco:
Fixes: https://github.com/ElementsProject/elements/issues/1460


delta1 (approved):
ACK ea75778bf3ab0df3f8ba352983bbd483c7303595

Issue #1410, opened 2025-02-13
fix win64 native ci job


delta1:
#1393 and #1408 fixed many issues with the win64 native ci job, but some additional fixes are required 

not a huge priority since windows release builds are cross compiled from guix, but creating this to track the issue

PR #1298, opened 2023-12-01
test: fix dbcrash and fee_estimation extended tests

delta1:
It seems these tests have never been run for Elements, since the extended tests are not run in the CI process. 

This PR fixes both of them, however there may be a further issue with `feature_dbcrash.py` running for an extremely long time which we are still in the process of testing. 

#1297 
#1296 

cc @whitslack

PR #1468, opened 2025-06-24
CI: Win64, split native test in 2 parts

psgreco:
Fixes: https://github.com/ElementsProject/elements/issues/1410

After this PR there are only 2 failing tests, which fail on all the architectures and are being handled in https://github.com/ElementsProject/elements/pull/1298


tomt1664 (approved):

delta1 (approved):
ACK a356e99de654983e76123f23a27769e97216fadc

PR #1476, opened 2025-07-18
Fee MoneyRange checks only apply to policyAsset

tomt1664:
Apply `HasValidFee` and fee `MoneyRange` checks only to the `policyAsset`. This will enable issued asset fee amounts (in individual outputs, total in transaction and total in a block) greater than MAX_MONEY

This is a consensus change and technically a hardfork, but can be added safely to a release and deployed as any issued asset fee paying transactions are currently prevented from being confirmed by policy rules. 

~Extension of https://github.com/ElementsProject/elements/pull/1333~

~HasValidFee (called from CheckInputs) is a consensus check for each tx that~
~1) any fee output does not exceed MAX_MONEY~
~2) any fee output is not 0 value~
~3) the total fees for a tx do not exceed MAX_MONEY~

~Another check in `CheckInputs` checks that the total fees for a block do not exceed MAX_MONEY~

~To enable both zero fee outputs and for total block and tx fee outputs (and total fee outputs) to exceed MAX_MONEY (fee outputs can be issued assets in addition to pegged asset, which total supply can exceed MAX_MONEY) and new chain config option is added: `-con_allow_any_fee`~

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -898,8 +898,6 @@ protected:
         std::vector<unsigned char> man_bytes = ParseHex(args.GetArg("-con_mandatorycoinbase", ""));
         consensus.mandatory_coinbase_destination = CScript(man_bytes.begin(), man_bytes.end()); // Blank script allows any coinbase destination
 
-        consensus.allow_any_fee = args.GetBoolArg("-con_allow_any_fee", consensus.allow_any_fee);
-
         // Custom chains connect coinbase outputs to db by default
         consensus.connect_genesis_outputs = args.GetIntArg("-con_connect_genesis_outputs", true);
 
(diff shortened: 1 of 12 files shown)


jsarenik:
I would just like to note here that a zero fee raw tx worked for me fine on the unofficial liquidtestnet4 already some time ago. Not tested with any other assets back then.

See https://kitchen.anyone.eu.org/lt4/zero.txt - the tx is present in current lt4 blockchain. Confidential. No fee (Liquid "fee" structure empty).

tomt1664:
> I would just like to note here that a zero fee raw tx worked for me fine on the unofficial liquidtestnet4 already some time ago. Not tested with any other assets back then.
> 
> See https://kitchen.anyone.eu.org/lt4/zero.txt - the tx is present in current lt4 blockchain. Confidential. No fee (Liquid "fee" structure empty).

A transaction can have a zero fee if there is no fee output (e.g. this lt4 tx) - that is consensus valid. But it cannot include an explicit fee output that has zero value (gets checked here: https://github.com/ElementsProject/elements/blob/b379bb24654aa99d8d7237601edc90369b873e77/src/confidential_validation.cpp#L33).

tomt1664:
I've updated https://github.com/ElementsProject/elements/pull/1333 so the wallet does not add a fee output if the fee is zero - this is valid in current consensus rules. 

The PR - enabling any fee value - is relevant to using issued assets as fees. I'll leave it open in draft - it is unlikely to be useful before issuance beyond `MAX_MONEY` is also enabled.

PR #1477, opened 2025-07-23
Unconditionally accept Simplicity spends in the mempool

roconnor-blockstream:
Taproot and other soft-forks historically in Bitcoin have not gated mempool access.


roconnor-blockstream:
@apoelstra It would be nice to test your hal-simplicity workflow against this PR.

roconnor-blockstream:
This PR is optional, but after reviewing https://github.com/bitcoin/bitcoin/pull/19953 I see that historical Bitcoin activation of similar soft-forks have no mempool logic, and simply allow the new transaction type unconditionally in the next version of the software.  Instead they relying on users not to use create such transaction until after activation (or risk having their funds stolen).

To than end this PR slightly simplifies our codebase and bring somewhat more in line with Bitcoin's.

This is a policy change and not a consensus rule change.

delta1 (approved):
utACK ddda11cb6a1909e68817c7e8e7db8c529e74902d 

Simplicity will already be active on liquidv1 by the time this is released.

apoelstra:
utACK ddda11cb6a1909e68817c7e8e7db8c529e74902d

PR #1478, opened 2025-07-24
chore: remove redundant words

jishudashu:
remove redundant words


delta1:
Thank you but these should be fixed up stream in bitcoin, secp256k1-zkp, and simplicity

PR #1499, opened 2025-10-22
ELIP 202: Implementation of optional sidechain peg-in subsidy and minimum peg-in amount

delta1:
Implementation for [ELIP 202](https://github.com/ElementsProject/ELIPs/blob/main/elip-0202.mediawiki)

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -977,7 +977,7 @@ protected:
         pegin_subsidy = ParsePeginSubsidy(args);
         pegin_minimum = ParsePeginMinimum(args);
         if (pegin_subsidy.threshold < pegin_minimum.amount) {
-            throw std::runtime_error(strprintf("Pegin subsidy threshold (%s) must be greater than or equal to pegin minimum amount (%s)", FormatMoney(pegin_subsidy.threshold), FormatMoney(pegin_minimum.amount)));
+            throw std::runtime_error(strprintf("Peg-in subsidy threshold (%s) must be greater than or equal to peg-in minimum amount (%s)", FormatMoney(pegin_subsidy.threshold), FormatMoney(pegin_minimum.amount)));
         }
 
         // Calculate pegged Bitcoin asset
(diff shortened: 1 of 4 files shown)


src/validation.cpp
```diff
@@ -663,6 +663,135 @@ class MemPoolAccept
...
+    bool CheckPeginSubsidyAndMinimum(TxValidationState& state, const CTransaction& tx, const std::vector<unsigned int>& pegin_indices)
+    {
+        // pegin_indices was calculated directly from the tx in prechecks, assert this invariant
+        assert(tx.witness.vtxinwit.size() >= pegin_indices.size());
+
+        if (pegin_indices.size() > 0) {
+            // calculate the burned subsidy value from the tx
+            CAmount subsidy = 0;
+            for (const CTxOut& txout : tx.vout) {
+                if (txout.scriptPubKey.IsUnspendable() && txout.nAsset.GetAsset() == Params().GetConsensus().pegged_asset && !txout.IsFee()) {
```
tomt1664: Are non-OP_RETURN but IsUnspendable() outputs allowed?

jsarenik:
Currently I run a pruned Elements node on liquidv1 with the default `validatepegin` setting of 1. Reading the ELIP sounds like it will not be possible when the new options get enabled because `getrawtransaction` on my pruned Bitcoin node will return the transaction merely while it is in the mempool (which can happen to be mined into a block before Elements node notices the transaction). IMHO it's asking for a race condition.

I don't like the idea of the main chain Bitcoin transaction fee influencing the amount getting burned on liquidv1 during peg-in (after being enabled, of course, but adding the possibility now into the code just begs for enabling it eventually).

Concept NACK

I may be misunderstanding something of course so please help me understand ELIP 202 if you see anything in what I write.

delta1:
@jsarenik  

> Currently I run a pruned Elements node on liquidv1 with the default validatepegin setting of 1. Reading the ELIP sounds like it will not be possible when the new options get enabled because getrawtransaction on my pruned Bitcoin node will return the transaction merely while it is in the mempool (which can happen to be mined into a block before Elements node notices the transaction). IMHO it's asking for a race condition. 

Do you mean with validatepegin=0? Even so, you can still compose the peg-in transaction with subsidy correctly by passing the parent (bitcoin) transaction feerate to `claimpegin` or `createrawpegin` if/when this change is ever active on liquidv1.

> I don't like the idea of the main chain Bitcoin transaction fee influencing the amount getting burned on liquidv1 during peg-in (after being enabled, of course, but adding the possibility now into the code just begs for enabling it eventually). 

As the Liquid Federation drives the tech roadmap for liquidv1, it's their prerogative should they want this enabled to subsidise or discourage low value peg-ins that they must pay for on the mainchain to sweep or spend. 

> Concept NACK 

Noted.

> I may be misunderstanding something of course so please help me understand ELIP 202 if you see anything in what I write. 

It may be worth reading it again.

PR #1503, opened 2025-11-04
Minor fixes for guix and regular builds

psgreco:
This MR contains 2 fixes
1. Cherry-pick from bitcoin (with modifications) the move from git.savannah.gnu.org to codeberg.org in guix builds
2. Elements-only fix to parse correctly the tag when doing tagged builds


delta1 (approved):
ACK 000227db763a9b04a1a7ab1c8355bb7008b3a53e 

Ran tests locally, have not tested guix yet

PR #1539, opened 2026-03-13
policy: allow exact annex padding for simplicity spends 

delta1:
Implements https://github.com/BlockstreamResearch/simplicity/issues/290 for Elements standardness policy. 

~~Something concerning to be investigated is that [`Cost::get_padding`](https://github.com/BlockstreamResearch/rust-simplicity/blob/d28440bc0c6be333aa84fa441844541c14dbb563/src/analysis.rs#L148) in rust-simplicity is returning a padding 2 bytes bigger than this calculation.. presumably an error in that method?~~ 

Fix for rust-simplicity `get_padding` at https://github.com/BlockstreamResearch/rust-simplicity/pull/356


src/script/interpreter.cpp
```diff
@@ -3322,7 +3326,23 @@ static bool VerifyWitnessProgram(const CScriptWitness& witness, int witversion,
                 simplicityRawTap.controlBlock = control.data();
                 simplicityRawTap.pathLen = (control.size() - TAPROOT_CONTROL_BASE_SIZE) / TAPROOT_CONTROL_NODE_SIZE;
                 simplicityRawTap.scriptCMR = script_bytes.data();
-                return checker.CheckSimplicity(simplicity_program, simplicity_witness, simplicityRawTap, budget, serror);
+                int64_t minCost = 0;
+                if ((flags & SCRIPT_VERIFY_ANNEX_PADDING) && padding.size() > 0) {
+                    valtype zero_padding(padding.size(), 0);
+                    zero_padding[0] = ANNEX_TAG;
+                    if (padding != zero_padding) {
+                        return set_error(serror, SCRIPT_ERR_SIMPLICITY_PADDING_NONZERO);
```
roconnor-blockstream: Feel free to push back on this idea, but I think it would be marginally better to move this zero-padding check into `policy.cpp`.  That is to say that `IsWitnessStandard` says that an annex is standard only when it is both

1. all zeros and
1. is part of a Simplicity redemption.

The all zero check can be validated without parsing of the Simplicity program, and moving the code there removes the need for an explicit `SCRIPT_ERR_SIMPLICITY_PADDING_NONZERO` error.  It just simply becomes non-standard to have a non-zero annex.

src/script/interpreter.cpp
```diff
@@ -3322,7 +3326,23 @@ static bool VerifyWitnessProgram(const CScriptWitness& witness, int witversion,
...
+                    }
+                    // Compute what the budget would have been without the padding.
+                    // budget includes the padding cost, so subtracting this stack item won't underflow.
+                    minCost = budget - ::GetSerializeSize(padding);
+                    if (!zero_padding.empty()) {
```
roconnor-blockstream: `zero_padding.empty()` can never be true here because `zero_padding.size > 0`.

Preferably we'd add a test case to catch this programming error before fixing it.  It would be something like a program that needs exactly one byte of padding in order to be spendable.  When I was last looking at this construction and asked @apoelstra to use his fuzzer to find such an example.

roconnor-blockstream (changes requested):

src/script/interpreter.cpp
```diff
@@ -3278,11 +3278,15 @@ static bool VerifyWitnessProgram(const CScriptWitness& witness, int witversion,
         // BIP341 Taproot: 32-byte non-P2SH witness v1 program (which encodes a P2C-tweaked pubkey)
         if (!(flags & SCRIPT_VERIFY_TAPROOT)) return set_success(serror);
         if (stack.size() == 0) return set_error(serror, SCRIPT_ERR_WITNESS_PROGRAM_WITNESS_EMPTY);
+        valtype padding;
```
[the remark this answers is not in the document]
roconnor-blockstream: If we are keeping the `zero_padding` in this file, then my cloning comment is no longer relevant.

src/script/interpreter.cpp
roconnor-blockstream: No no, keep the `minCost` addition inside the conditional.

The logic here is as as follows, and maybe we should explicitly document this in the code:

(1) If there is no annex, or we are in consensus checking mode, `minCost` is set to 0 which effectively disable any overweight cost checks.

(2) If there is an annex, we set `minCost` to be what the budget would have been if they user made their annex padding smaller.  This computation proceeds in two stages:

First we set `minCost` to be what the stack size would have been without the annex.  Because the number of stack items is at most 3, this calculation can be done by subtracting the annex stack item size from the budget.

(a) If the annex exists and is empty, then the only way the annex could be smaller is by eliminating it entirely, and thus we use the above computed value for `minCost`.  Note: in this case the `minCost` is 2 WU less than the budget.

(b) If the annex exists and is non-empty, then we add to `minCost` the value of an annex that contains one fewer byte.  Note: If the length of the padding (including the annex prefix) is exactly 253, then `minCost` will be 3 WU less than the budget!  Similarly for padding of length 65536.

src/policy/policy.cpp
```diff
@@ -282,8 +283,16 @@ bool IsWitnessStandard(const CTransaction& tx, const CCoinsViewCache& mapInputs)
...
             Span stack{tx.witness.vtxinwit[i].scriptWitness.stack};
             if (stack.size() >= 2 && !stack.back().empty() && stack.back()[0] == ANNEX_TAG) {
-                // Annexes are nonstandard as long as no semantics are defined for them.
-                return false;
+                SpanPopBack(stack); // drop the annex
+                const auto& control_block = SpanPopBack(stack);
+                if (!control_block.empty() && (control_block[0] & TAPROOT_LEAF_MASK) == TAPROOT_LEAF_TAPSIMPLICITY) {
+                    // Annexes are allowed for Simplicity spends
+                    // checks for zero padding and exact size are done in CheckSimplicity
+                    return true;
```
roconnor-blockstream: Preempting the rest of the policy by returning true here is a little optimistic. :)

delta1: Thanks for catching this. updated in 3d352ed which I've just added as a fixup commit for ease of review, will squash it later

apoelstra:
utACK 3d352ed91ece7a63b8bddeace50a0ad91e78999d

Needs rebase. Sorry for the long delay.

delta1:
Thanks @apoelstra. Rebased.

delta1:
Lint CI job is unrelated to this PR, and resolved by #1572 

"test each commit" job is also unrelated and will eventually be fixed upstream by BlockstreamResearch/simplicity#345

apoelstra:
utACK fb337f2f9d8f7b279631a50574280733d2b38a96

apoelstra:
What needs to be done to land this?

delta1:
@apoelstra I think it's good to go, I'll fix the lint CI job (edit: rebased on master). 

"test each commit" needs a subtree update so I don't see it as a blocker for this PR.

src/policy/policy.cpp
```diff
@@ -315,8 +316,14 @@ bool IsWitnessStandard(const CTransaction& tx, const CCoinsViewCache& mapInputs)
             // Taproot spend (non-P2SH-wrapped, version 1, witness program size 32; see BIP 341)
             Span stack{tx.witness.vtxinwit[i].scriptWitness.stack};
             if (stack.size() >= 2 && !stack.back().empty() && stack.back()[0] == ANNEX_TAG) {
-                // Annexes are nonstandard as long as no semantics are defined for them.
-                return false;
+                SpanPopBack(stack); // drop the annex
+                const auto& control_block = SpanPopBack(stack);
```
apoelstra, line 320: In bfde7b715:

This looks like a double-pop. A few lines below this change there is another if-block, `if (stack.size() >= 2)` (which will trigger if this block triggers) that also has

```
const auto& control_block = SpanPopBack(stack);
```

Found by Qwen 3.8-Flash-Next.

apoelstra: The bot also gave me this unit test file

```c++
// Temporary review probe for Elements PR #1539 (NOT for commit):
// annexed Simplicity spend where the 32-byte CMR starts with 0xc0/0xc1.
// IsWitnessStandard's annex branch pops annex AND control block, then the
// generic script-path branch pops the CMR as "control block"; if cmr[0] & 0xfe
// == 0xc0 (Tapscript leaf), MAX_STANDARD_TAPSCRIPT_STACK_ITEM_SIZE (80) is
// applied to the remaining witness items. Probe compares a Simplicity annexed
// spend with a 100-byte witness across cmr[0] in {0x11, 0xc0, 0xc1, 0xbe}.
#include <test/util/setup_common.h>

#include <coins.h>
#include <policy/policy.h>
#include <primitives/transaction.h>
#include <script/script.h>
#include <arith_uint256.h>
#include <uint256.h>

#include <boost/test/unit_test.hpp>

#include <map>
#include <vector>

namespace {
class MapCoinsView : public CCoinsView
{
    std::map<COutPoint, Coin> m_coins;
public:
    void Add(const COutPoint& out, Coin coin) { m_coins[out] = std::move(coin); }
    std::optional<Coin> GetCoin(const COutPoint& outpoint) const override
    {
        auto it = m_coins.find(outpoint);
        if (it == m_coins.end()) return std::nullopt;
        return it->second;
    }
};
} // namespace

BOOST_FIXTURE_TEST_SUITE(annex_padding_probe, BasicTestingSetup)

static CTransactionRef MakeAnnexedSpend(uint8_t cmr_first, size_t witness_len)
{
    CMutableTransaction tx;
    tx.vin.resize(1);
    tx.vin[0].prevout = COutPoint(Txid::FromUint256(ArithToUint256(1)), 0);
    tx.vout.resize(1);
    tx.vout[0].nValue.SetToAmount(1000);

    CTxInWitness win;
    std::vector<uint8_t> cmr(32, 0x22);
    cmr[0] = cmr_first;
    std::vector<uint8_t> control(33, 0xcc);
    control[0] = 0xbe; // TAPROOT_LEAF_TAPSIMPLICITY, depth 0
    std::vector<uint8_t> annex(12, 0);
    annex[0] = 0x50;   // ANNEX_TAG + zero padding
    win.scriptWitness.stack = {
        std::vector<uint8_t>(witness_len, 0xaa), // "simplicity witness"
        std::vector<uint8_t>(40, 0x11),          // "simplicity program"
        cmr,                                     // script CMR (32 bytes)
        control,                                 // control block
        annex,
    };
    tx.witness.vtxinwit.push_back(win);
    return MakeTransactionRef(std::move(tx));
}

BOOST_AUTO_TEST_CASE(annex_padding_double_pop_probe)
{
    CScript taproot_spk;
    std::vector<uint8_t> prog(32, 0xab);
    taproot_spk << OP_1 << prog;

    MapCoinsView view;
    CTxOut prevout;
    prevout.nValue.SetToAmount(100000);
    prevout.scriptPubKey = taproot_spk;
    view.Add({Txid::FromUint256(ArithToUint256(1)), 0}, Coin{std::move(prevout), 1, false});
    CCoinsViewCache coins(&view);

    const bool a = IsWitnessStandard(*MakeAnnexedSpend(0x11, 100), coins);
    const bool b = IsWitnessStandard(*MakeAnnexedSpend(0xc4, 100), coins);
    const bool c = IsWitnessStandard(*MakeAnnexedSpend(0xc5, 100), coins);
    const bool d = IsWitnessStandard(*MakeAnnexedSpend(0xbe, 100), coins);
    const bool e = IsWitnessStandard(*MakeAnnexedSpend(0xc4, 80), coins); // <= limit
    BOOST_TEST_MESSAGE("100-byte witness: cmr[0]=0x11 -> " << a << "; 0xc4 -> " << b
                       << "; 0xc5 -> " << c << "; 0xbe -> " << d << "   | 80-byte witness, 0xc4 -> " << e);
    BOOST_CHECK(a);
    BOOST_CHECK(d);
    BOOST_CHECK(e);
    BOOST_CHECK_EQUAL(b, a);
    BOOST_CHECK_EQUAL(c, a);
}

BOOST_AUTO_TEST_SUITE_END()
```

PR #1542, opened 2026-04-07
Change test order in feature_discount_ct.py to fix intermittent failure

tomt1664:
Prevents blinded output being selected intermittently and resulting in vszie and fee assertions failing.


delta1 (approved):
ACK 6515f583514022cbcea12e27c0405f36b6c0d4a5; fixes discount_ct test flakiness

PR #1552, opened 2026-05-21
Uncap datacarrier by default

ivanlele:
Cherry picks:

https://github.com/bitcoin/bitcoin/commit/f3bbc746647d1fd23bf5cfe
https://github.com/bitcoin/bitcoin/commit/451ba9ada41f687c0e4bb34d5925374a68a8f8a3

### DIFF
--- a/test/functional/mempool_accept.py
+++ b/test/functional/mempool_accept.py
@@ -330,7 +330,6 @@ class MempoolAcceptanceTest(BitcoinTestFramework):
             result_expected=[{'txid': tx.rehash(), 'allowed': False, 'reject-reason': 'min relay fee not met'}],
             rawtxs=[tx.serialize().hex()],
         )
-
         self.log.info('A timelocked transaction')
         tx = tx_from_hex(raw_tx_reference)
         tx.vin[0].nSequence -= 1  # Should be non-max, so locktime is not ignored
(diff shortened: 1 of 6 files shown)


delta1:
Looks good so far, just needs an update to one of the functional tests (or a new functional test) to check it's working correctly

ivanlele:
> Looks good so far, just needs an update to one of the functional tests (or a new functional test) to check it's working correctly

Pushed the test for this into one of the functional tests. The only thing left is for CI to pass again because of this change.

The previous version had almost passed CI here:
https://github.com/ElementsProject/elements/runs/77381958009

I'll will monitor new run, but I think it can already be reviewed

test/functional/mempool_accept.py
```diff
@@ -349,6 +348,17 @@ def run_test(self):
             maxfeerate=0,
         )
 
+        self.log.info('Multiple OP_RETURN and large data are standard with default datacarriersize (100000)')
+        # Multiple OP_RETURN outputs with data far exceeding the old per-output
+        tx = tx_from_hex(raw_tx_reference)
+        tx.vout.append(CTxOut(nValue=0, scriptPubKey=CScript([OP_RETURN, b'\xff'])))
+        tx.vout.append(CTxOut(nValue=0, scriptPubKey=CScript([OP_RETURN, b'\xff' * 50000])))
```
delta1, line 355: is 50000 the largest we can make this op_return before reaching the transaction size limit?

ivanlele: This is from a cherry-picked commit. The maximum value should be around 99xxx I think. I can try to pick the exact value and update it here instead.

https://github.com/bitcoin/bitcoin/commit/f3bbc746647d1fd23bf5cfe#diff-a99d72c0ed66c256169e92327e04ab223e71f5ef598e14aac0ff44dc2a1194daR343

delta1: Ah okay, if it matches upstream that’s fine

delta1 (approved):
ACK b521a1bf4cebf642c70e1ea58b7210857bc36793; tested locally

PR #1560, opened 2026-06-24
rpc: fix fields description in a man in 'decodepsbt'

RuslanProgrammer:


src/rpc/rawtransaction.cpp
```diff
@@ -1155,9 +1155,7 @@ static RPCHelpMan decodepsbt()
                         {RPCResult::Type::NUM, "output_count", "The number of outputs in this psbt."},
                         {RPCResult::Type::NUM, "inputs_modifiable", /*optional=*/true, "Whether inputs can be modified"},
                         {RPCResult::Type::NUM, "outputs_modifiable", /*optional=*/true, "Whether outputs can be modified"},
```
delta1: These should also be `BOOL`  


```suggestion
{RPCResult::Type::BOOL, "inputs_modifiable", /*optional=*/true, "Whether inputs can be modified"},
{RPCResult::Type::BOOL, "outputs_modifiable", /*optional=*/true, "Whether outputs can be modified"},
```

RuslanProgrammer: Fixed

tomt1664:
ACK 91ae0e92e67cc32c5d9141948f070ec516854b82

tomt1664:
ACK ef1d2673d700061eb7349431c54e59d91da64efc tested locally

PR #1562, opened 2026-06-24
Legacy wallet support for v29

tomt1664:
Enable legacy wallet creation (and set as default in GUI). 
Removal of deprecated rpc check for create_bdb wallets.


delta1 (approved):
ACK ac16e342411a048974391f8898c7489ed12978c4; ran tests locally, did not test gui

PR #1564, opened 2026-06-25
mempool: disable full-RBF, require BIP125 opt-in signaling

delta1:
Bitcoin Core defaults to accepting any fee-bumping replacement regardless of signaling (full-RBF). Revert this for Elements: only replace mempool transactions that explicitly opt in via BIP125 nSequence signaling (or TRUC), and report fullrbf=false in getmempoolinfo.

Update feature_rbf.py accordingly, and add a regression test for CVE-2021-31876 confirming that inherited signaling from an unconfirmed parent does not make a non-signaling child replaceable.


tomt1664:
Tested locally ACK 0e0d01913a881845fa95ed37b012f7e7edd2c65d

`wallet_basic.py --descriptors` failing in [ASan + LSan + UBSan + integer, no depends, USDT](https://github.com/ElementsProject/elements/actions/runs/28173324544/job/83442973886?pr=1564#logs)
Not sure if intermittent.

tomt1664:
Passing all jobs on re-run.

tomt1664 (approved):

Issue #1106, opened 2022-04-15
`initialfreecoins` does not work for descriptor wallets. 


sanket1729:
I am not even sure if this should work. If it worked earlier for me. But it does not work now.

Steps to reproduce:
Set the following in the elements.conf file (regtest mode)
```
initialfreecoins=2100000000000000
anyonecanspendaremine=1
```

Create a descriptor wallet
```
elements-cli -named createwallet wallet_name=descriptors_e1 descriptors=true
```
Wait for funds to mature
```
elements-cli -generate 101 
```
Check the balance
```
elements-cli getbalance
```
This should be equal to `2100000000000000`, but is `0`. The same works for non-descriptor wallets

sanket1729:
Workaround:

In `elements.conf`
```
con_blocksubsidy=5000000000
```

Create funds manually by adding a subsidy on the side of the elements blockchain. This should allow developers to test easily with descriptor wallets without using initial funds. Just like bitcoin, this produces 50 bitcoins per block on elements node that can be used in the wallet after 100 blocks `elements-cli -generate 101`

ErikBorgh1:
Thanks for getting me approved

tiero:
> This should be equal to `2100000000000000`, but is `0`. The same works for non-descriptor wallets


This happen to us for another reason, to fix we need to call to `rescanblockchain` right after generating 101 blocks

dgpv:
elements code tutorial https://elementsproject.org/elements-code-tutorial/working-environment should be updated to create non-descriptor wallet, otherwise novices might be stuck. I couldn't create issue on elementsproject github (issues turned off there?), so I'll just leave a comment here and tag the author of the tutorial

@wintercooled

PR #1565, opened 2026-06-26
Guard against trimmed chain_start in TryLowWorkHeadersSync

tomt1664:
When -trim_headers is enabled, TryLowWorkHeadersSync can receive a chain_start_header pointing to a trimmed block index entry. The `HeadersSyncState` constructor immediately calls `GetBlockHeader()` on it via the m_last_header_received initialiser, which hits `assert_untrimmed()` and aborts the node. 

Conditional untrim_to guard added inside `TryLowWorkHeadersSync`, before `HeadersSyncState`. 

For nodes running -trim_headers, in practice the crash path is not reachable on Elements because `already_validated_work` is always true on a federated chain (nChainWork == nMinimumChainWork == 0), so `TryLowWorkHeadersSync` is not entered. But this ensures consistency with guarded serving paths. 

`TryLowWorkHeadersSync` and `HeadersSyncState` were introduced in Bitcoin v24 (bitcoin/bitcoin#25717) and so this is not applicable to Elements v23.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -2634,6 +2634,15 @@ bool PeerManagerImpl::TryLowWorkHeadersSync(Peer& peer, CNode& pfrom, const CBlo
         // otherwise they don't have more headers after this so no point in
         // trying to sync their too-little-work chain.
         if (headers.size() == m_opts.max_headers_result) {
+            // chain_start_header may refer to a block deep in the chain that
+            // has been trimmed from memory by -trim_headers. The HeadersSyncState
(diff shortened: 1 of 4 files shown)


delta1 (approved):
ACK e151cada58e00f9e12df2fe36a2a99934bbac40f; tested locally

psgreco:
Does this fix https://github.com/ElementsProject/elements/issues/1106 ?

delta1:
> Does this fix #1106 ?

Yes I think so

PR #1571, opened 2026-07-07
Additional RPC documentation fixes

tomt1664:
Additional fixes for RPC documentation. 

| # | File | RPC | Issue |
| --- | --- | --- | --- |
| 1 | blockchain.cpp | getblockheader, getblock | "target" documented but never emitted |
| 2 | blockchain.cpp | getdescriptoractivity | "spend_vout" in docs vs "spend_vin" in code |
| 3 | blockchain.cpp | getblock | Typo "Uninforced" vs "Unenforced" |
| 4 | blockchain.cpp | scantxoutset | "asset" not optional |
| 5 | blockchain.cpp | getsidechaininfo | current_fedpeg_programs/current_fedpegscripts output order inverted |
| 6 | mining.cpp | getmininginfo | "next" not marked optional; absent on all signed-block chains |
| 7 | mining.cpp | getmininginfo | "signet_challenge"/"next" emit order inverted vs docs |
| 8 | mining.cpp | combineblocksigs | witnessScript declared optional but required for dynafed |
| 9 | mining.cpp | getcompactsketch | Return type STR should be STR_HEX |
| 10 | mining.cpp | getblocktemplate | "capabilities" pushed first, documented fourth |
| 11 | mining.cpp | getnewblockhex | Stale "Wallet keypool empty" error message |
| 12 | rawtransaction.cpp | decodepsbt | "fees" documented twice with conflicting types |
| 13 | rawtransaction.cpp | decodepsbt | "input_count"/"output_count" not marked optional |
| 14 | rawtransaction.cpp | getrawtransaction | Verbosity=0 return type STR should be STR_HEX |
| 15 | rawtransaction.cpp | combinerawtransaction | Return type STR should be STR_HEX |
| 16 | rawtransaction.cpp | rawblindrawtransaction | Return type STR should be STR_HEX |
| 17 | rawtransaction.cpp | rawblindrawtransaction | inputAmounts error says "input blind" |
| 18 | rawtransaction.cpp | rawblindrawtransaction | asset validation error says "asset blinds" not "asset ids" |
| 19 | rawtransaction.cpp | signrawtransactionwithkey | privatekey typed STR_HEX but is base58 |

Wallet RPC fixes:

| # | File | RPC | Issue |
| --- | --- | --- | --- |
| 1 | coins.cpp | listunspent | "asset", "amountblinder", "assetblinder" not marked optional |
| 2 | coins.cpp | listunspent | "ancestorfees" typed STR_AMOUNT but code emits raw integer |
| 3 | coins.cpp | getreceivedbyaddress, getreceivedbylabel | include_immature_coinbase examples pass true as assetlabel |
| 4 | spend.cpp | walletcreatefundedpsbt | psbt_version implies configurability but only version 2 is accepted |
| 5 | spend.cpp | signrawtransactionwithwallet | amountcommitment prevtxs arg typed STR instead of STR_HEX |
| 6 | wallet.cpp | getwalletinfo | Balance OBJ fields document a fixed "bitcoin" key; code emits dynamic asset map |
| 7 | elements.cpp | createrawpegin | Result "hex" typed STR instead of STR_HEX |
| 8 | elements.cpp | sendtomainchain | bip32_counter result typed STR; should be NUM to match input arg type |


src/rpc/mining.cpp
```diff
@@ -1310,7 +1310,7 @@ static RPCHelpMan combineblocksigs()
                             },
                         },
                     },
-                    {"witnessScript", RPCArg::Type::STR_HEX, RPCArg::Optional::OMITTED, "The hex-encoded witnessScript for the signblockscript"},
+                    {"witnessScript", RPCArg::Type::STR_HEX, RPCArg::Optional::NO, "The hex-encoded witnessScript for the signblockscript"},
```
delta1: I don't think this is correct: the handler only requires it conditionally:

```cpp
if (is_dynafed) {
    if (request.params[2].isNull()) {
        throw JSONRPCError(RPC_INVALID_PARAMETER, "Signing dynamic blocks requires the witnessScript argument");
    }
    ...
}
```

`RPCHelpMan::IsValidNumArgs()` (src/rpc/util.cpp:766) computes num_required_args from the last non-optional argument. Since witnessScript is the last positional arg, marking it NO makes 3 params the hard minimum for every call. A legacy (non-dynafed) call passing only blockhex + signatures — previously valid and still handled correctly in the body — will now fail argument-count validation and throw the help text before the handler even runs. This silently breaks existing non-dynafed signing workflows.

Suggested fix: keep Optional::OMITTED, and instead improve the description text (e.g. "required for dynafed blocks") rather than changing the optionality tag.

src/rpc/blockchain.cpp
```diff
@@ -2902,7 +2900,7 @@ static RPCHelpMan getdescriptoractivity()
                         {RPCResult::Type::STR_HEX, "spend_txid", "The txid of the spending transaction"},
                         {RPCResult::Type::NUM, "spend_vout", "The vout of the spend"},
                         {RPCResult::Type::STR_HEX, "prevout_txid", "The txid of the prevout"},
-                        {RPCResult::Type::NUM, "prevout_vout", "The vout of the prevout"},
+                        {RPCResult::Type::NUM, "prevout_vin", "The vin of the prevout"},
```
delta1: This isn't right, the code is : 

```cpp
event.pushKV("spend_txid", tx->GetHash().ToString());
event.pushKV("spend_vin", vin);          // doc still says "spend_vout" — unfixed
event.pushKV("prevout_txid", txin.prevout.hash.ToString());
event.pushKV("prevout_vout", txin.prevout.n);   // doc now wrongly says "prevout_vin"
```

Suggested fix: revert the prevout_vout→prevout_vin change; instead rename the spend_vout doc entry to spend_vin (line ~2901).

src/rpc/mining.cpp
delta1: I think we probably need to cover this in a functional test somewhere to avoid any regressions

delta1 (changes requested):
Looks really good, just 2 things to fix up

src/rpc/mining.cpp
tomt1664: Added new function to feature_blocksign.py

src/rpc/blockchain.cpp
tomt1664: Change wrong line! Fixed.

delta1 (approved):
ACK e5d02b9fd50860258749e02a5c7d8946075ab9e5; tested locally

PR #1574, opened 2026-07-24
29.x: Merged backports to 29.4

tomt1664:
Upstream merges from Bitcoin 29.x to 29.4. 

Additional fixes for CI:

## 1. Merge setup & CI infrastructure

- **Merge script sanity check** — confirmed all 42 upstream backport PRs (`e9e6825b8c`..`3fc0865963`) were merge commits with no stray non-merge first-parent commits; `merge-prs.sh` pointed at `elements-29.x`.
- **CI migration, Cirrus CI → GitHub Actions** — resolved Docker/host environment conflicts, macOS path divergence (`~/.elements` vs `~/Library/Application Support/Elements`), subtree lint failures (`secp256k1`, `minisketch`), `test-each-commit` `FETCH_DEPTH` issues, Windows MSVC job config.
- **`ci/test/03_test_script.sh`** — removed dead autotools-era code paths (`./autogen.sh`, `./configure`), rewrote to match upstream's CMake-only flow.
- **doc/man generation** — resolved merge conflicts from upstream's `doc/man/bitcoin-*.1` vs Elements' `elements-*.1` naming; regenerate via `contrib/devtools/gen-manpages.py` at release time instead of merging.
- **`contrib/devtools/check-deps.sh`** — added ~30 missing `SUPPRESS[...]` entries recording Elements-specific cross-module dependencies (confidential_validation, pegins, mainchainrpc, pak, dynafed, block_proof, versionbits) that upstream's dependency-graph checker had no knowledge of.
- **CentOS/depends/GUI job — "No space left on device"** — infrastructure issue (disk/ccache exhaustion on the runner), not a code bug; not something fixable via source changes.
- **`--exclude feature_dbcrash`** — `feature_dbcrash.py` is commented out of Elements' own `test_runner.py` test list (too disk/time-intensive for GHA); excluding an already-absent test is now a hard error in `test_runner.py`, not a no-op. Removed from the exclude list in `ci/test/00_setup_env_native_previous_releases.sh`.
- **Same file — `--extended,feature_fee_estimation,...` argparse corruption**

---

## 2. RPATH — "libsecp256k1.so.6: cannot open shared object file" bug

top-level `CMakeLists.txt` sets `CMAKE_SKIP_BUILD_RPATH TRUE` globally (deliberate upstream policy, tied to Guix reproducible builds). This is invisible under a normal (static-secp256k1) build, but any build enabling `libbitcoinkernel` forces secp256k1 to build as a shared library — and every other executable target then fails at runtime unless it gets the same opt-out.

- `elements-tx`, `elements-util` (`src/CMakeLists.txt`)
- `test_elements` (`src/test/CMakeLists.txt`)
- `bench_bitcoin` (`src/bench/CMakeLists.txt`)
- `elements-qt` (`src/qt/CMakeLists.txt`)
- `test_elements-qt` (`src/qt/test/CMakeLists.txt`)
- `elementsd` (`src/CMakeLists.txt`)
- `elements-cli` (`src/CMakeLists.txt`)

---

## 3. Code/sanitizer bugs

- **`CTxOut` 2-arg vs 3-arg constructor** — ported Bitcoin Core test/IPC code used upstream's `CTxOut(amount, script)`; fixed to Elements' `CTxOut(CAsset(), amount, script)` in `transaction_tests.cpp`, `test/ipc_test.cpp`.
- **`checkqueue.h::Loop()` ownership bug** — failing/skipped-after-failure checks never `delete`d in Elements' pointer-based (`T*`) `CCheckQueue`; real leak reachable from production validation code.
- **`spend.cpp` — `vtxinwit`/`vin` invariant violation** — `vtxinwit.emplace_back()` only ran inside a conditional branch; moved to run unconditionally alongside every `vin.emplace_back()`.
- **`wallet.cpp::SignPSBT`** — missing `tx.witness.vtxinwit.resize(tx.vin.size())` (had the `vtxoutwit` equivalent but not the `vtxinwit` one); found via gdb backtrace.
- **`coins.h::AddFlags`** — `Assume(flags & (DIRTY|FRESH))` didn't account for Elements' third `PEGIN` flag; fixed to `Assume(flags & (DIRTY|FRESH|PEGIN))`.
- **`kernel/chainparams.h::CChainParams`** — missing virtual destructor; `new-delete-type-mismatch` deleting a `CCustomParams` through `unique_ptr<const CChainParams>`.
- **`validation.cpp::CheckInputScripts` synchronous path** — leaked the heap-allocated `CScriptCheck` on every call (missing `delete`).
- **`test/fuzz/deserialize.cpp`** — generic round-trip fuzz check inappropriately applied to `Coin`, which asserts `!IsSpent()` on serialize; fuzzed bytes trivially violate this.
- **`test/fuzz/package_eval.cpp` — duplicate `tx_mut.vin.push_back(in)`** — an earlier fix (adding `vtxinwit.emplace_back()` alongside the `push_back`) was applied *additively* instead of *replacing* the original line, leaving every selected input pushed into `vin` twice. Traced via reproducible fuzz crash + gdb + targeted `fprintf` instrumentation to a `outpoints.insert(in.prevout).second` assertion failure; root-caused to the duplicate push, not (as initially suspected) `OutpointsUpdater` or cross-iteration state.
- **`kernel/chainparams.cpp::CTestNet4Params`** — entire Elements-specific initialization block (`enforce_pak`, `accept_unlimited_issuances`, `multi_data_permitted`, `accept_discount_ct`, `create_discount_ct`, `pegin_subsidy`, `pegin_minimum`, `anyonecanspend_aremine`, `consensus.has_parent_chain`, `consensus.genesis_subsidy`, `consensus.connect_genesis_outputs`, `consensus.subsidy_asset`, `g_signed_blocks`) missing entirely — `testnet4` is a brand-new upstream 29.x chain type, and the new class was adapted from upstream's version (which has no concept of any of these fields) without ever adding Elements' equivalent block. Caught independently by both UBSan (`invalid-bool-load`) and MSan (`use-of-uninitialized-value`) in two separate CI jobs.
- **`validation.cpp::GuessVerificationProgress`** — unguarded division (`pindex->nHeight / (pindex->nHeight + moreBlocksExpected)`) with a denominator that can hit exactly zero when a block's timestamp is far enough ahead of "now" (e.g. during `feature_taproot.py`'s test-vector generation). Added a guard clause returning `1.0` when the extrapolation is degenerate.
- **`wallet/rpc/spend.cpp` — `psbtbumpfee` not-mine-input path** — `FillPSBT(...)` call omitted Elements' `imbalance_ok` parameter (defaults to `false`), causing a legitimate CT-balance-check failure for external/not-owned inputs to trip a `CHECK_NONFATAL(!err)` that assumed the call could never fail. Every other Elements-specific `FillPSBT` call site in the same file already passes `/*imbalance_ok=*/true`; this one didn't. Root-caused by diffing against upstream (confirmed the `CHECK_NONFATAL` pattern itself is unmodified) and against Elements' own `FillPSBT` declaration.
- Fee outputs in USDT tests. 

---

## 4. Type/size mismatches

Several issues traced to `CoinsCachePair`/`Coin`/`CTxOut` being substantially larger than upstream's equivalents (confidential asset/value/nonce commitments), breaking constants and assumptions tuned for upstream's smaller structs:

- **`CTxMemPoolEntry::discountSizeWithAncestors`** — declared `uint64_t` while its three sibling ancestor-state accumulators are `int64_t`; the odd type caused a UBSan `implicit-integer-sign-change` when a negative delta was added during removal. Retyped to `int64_t` to match siblings.
  - **Follow-up compile fix:** `GetDiscountTxSize()` returns `size_t`; brace-init into the now-`int64_t` member needed an explicit `static_cast<int64_t>(...)`.
  - Confirmed the six external call sites (`node/miner.h`, `node/miner.cpp`, `txmempool.h`) did **not** need changes — they only ever read the post-assert, always-non-negative value into `uint64_t`, which is a value-preserving conversion UBSan doesn't flag.
- **`coins.h::CCoinsMapMemoryResource`** — `PoolResource`'s generic default constructor hardcodes a 256 KiB chunk, sized for upstream's smaller `CoinsCachePair`; no longer big enough to hold 1000 of Elements' larger nodes without overflowing into a second chunk. Fixed by passing an explicit `chunk_size_bytes` scaled to `sizeof(CoinsCachePair)`.
  - **Follow-up:** `validation_flush_tests.cpp`'s `getcoinscachesizestate` test hardcoded `MAX_COINS_CACHE_BYTES = 262144 + 512` based on the (now-stale) "PoolResource defaults to 256 KiB" assumption; updated to compute the same way as the fixed default.
- **32-bit ARM `validation_flush_tests.cpp`** — the test's own "unsupported architecture" fallback path hardcoded `1000` coins as "obviously enough" to cross the CRITICAL cache-size threshold; doesn't hold on 32-bit ARM where `COIN_SIZE` is smaller (32-bit pointers). Fixed by computing the coin count from `MAX_COINS_CACHE_BYTES / COIN_SIZE` instead of a fixed constant.

---

## 5. Build configuration / CMake

- **`src/CMakeLists.txt` — simplicity subtree warnings** — `-Wconditional-uninitialized`/`-Wimplicit-fallthrough` (Clang) and `-Wtype-limits` (GCC/mingw) in vendored `dag.c`/`rsort.c`; suppression was gated `if(APPLE)` only, extended to all Clang toolchains plus a GCC-specific suppression.
- **`src/qt/CMakeLists.txt`** — `WITH_MULTIPROCESS` block called `import_plugins(bitcoin-gui)`, never renamed to Elements' `elements-gui` target.
- **`src/kernel/CMakeLists.txt`** — `libbitcoinkernel` missing several source files (`addresstype.cpp`, `key.cpp`, `script/miniscript.cpp`, `common/{config,settings}.cpp`, `chainparamsbase.cpp`, `mainchainrpc.cpp`) and libevent linkage, needed because `pegins.cpp` (pulled into the kernel lib) calls `IsConfirmedBitcoinBlock` unconditionally.
- **Windows cross-compilation** — secp256k1 `__imp_` linker symbols (fixed via `PUBLIC` CMake linkage); `pegins.cpp`/`confidential_validation.cpp` needed to move from `bitcoin_node` to `bitcoin_common` to satisfy the dependency graph; NSIS installer asset staging.

## 6. Toolchain / fuzz harness

- **`test/fuzz/simplicity_tx.cpp`** — missing `SeedRandomStateForTest(SeedRand::ZEROS);` as the first line of the `FUZZ_TARGET` body, required by the `CheckGlobals` reproducibility guard whenever a target touches the global RNG state. Also needed the corresponding `#include <test/util/random.h>`

---

## 7. Python functional-test gaps

- **`test/functional/p2p_headers_sync_with_minchainwork.py`** and **`p2p_headers_presync.py`** — both call `self.nodes[0].createwallet(...)` unconditionally (an Elements-specific addition needed because Elements' `generatetoaddress`-based mining approach requires a wallet address, unlike upstream's `self.generate()`), but neither had the standard `self.skip_if_no_wallet()` guard that every other wallet-touching test in the suite already has, causing `Method not found (-32601)` failures in no-wallet CI builds. Both fixed by adding the guard as the first line of `run_test()`. Flagged for a broader sweep (`grep -rL skip_if_no_wallet` against files that call wallet methods) since both were adapted from the same original source and the same gap may recur elsewhere.

---

## 8. Static analysis / clang-tidy

Batch of ~9 findings from a `-warnings-as-errors` `clang-tidy` run,
- `rpc/util.cpp` — two hand-written trivial default constructors → `= default`.
- `qt/networkstyle.cpp` — legitimate bounded single-level recursion (`instantiate()`, guarded by an `assert` immediately before the recursive call) → `NOLINT(misc-no-recursion)` rather than a rewrite.
- `wallet/coinselection.cpp` — two occurrences of `auto input_set/set = result.GetInputSet();` copy-constructed but only used by reference (`.find()`/`.end()` need to operate on the same instance, so the copy can't simply be inlined away either) → `const auto&`. One occurrence was caught by CI in a second pass after the first fix (§10 recurrence).
- `assetsdir.cpp` — range-for copying each string → `const std::string&`.
- `primitives/bitcoin/merkleblock.cpp` — three inherently-recursive Merkle-tree traversal functions, straight from unmodified upstream → `NOLINT(misc-no-recursion)` rather than rewriting a correctness-sensitive algorithm.
- `validation.cpp` — `emplace_back` in a loop of known size without `reserve()`.
- `bitcoin-tx.cpp` — three occurrences of the same unnecessary-copy pattern as coinselection.cpp.
- `test/blind_tests.cpp` — `push_back` in a fixed 6-iteration loop without `reserve()`.

## 9. Previous releases

- `wallet_migration.py` disabed. 
- `wallet_backwards_compatibility.py` remains disabled (as in 23.x).


delta1 (approved):
ACK 7055c83; tested locally 

I took the liberty of merging elements-29.x and resolving the conflict.

Waiting on CI.

delta1 (approved):
ACK cd384f01a2565cfd0d5c30f71e7f02976a7e3b07; tested locally

PR #1576, opened 2026-08-07
Fix intermittent failure in feature_pegin_subsidy.py

tomt1664:
feature_pegin_subsidy.py has several assertions comparing the burn subsidy against hardcoded constants assumed fixed-length witness (t * 72 bytes per signature), but the actual subsidy is computed by the node from the parent-chain deposit transaction's real fee and vsize. DER-signature lengths on the parent-chain transaction vary by a byte or two causing a rare intermittent failure.

Replaced hardcoded subsidy values with a Python function that replicates `CheckPeginSubsidyAndMinimum`'s computation.

### DIFF
--- a/test/functional/feature_fedpeg.py
+++ b/test/functional/feature_fedpeg.py
@@ -10,6 +10,7 @@ from test_framework.util import (
     get_datadir_path,
     rpc_port,
     p2p_port,
+    tor_port,
     assert_raises_rpc_error,
     assert_equal,
     find_vout_for_address,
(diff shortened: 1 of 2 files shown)


delta1:
Thanks @tomt1664 this looks good. Busy running it in a loop now.

delta1 (approved):
ACK 45fd28fca72b19fcc2db9d5fa013aee16febf75b; tested locally 

Added 2 commits to fix the pegin subsidy and fedpeg tests to work with parent bitcoind correctly

PR #1577, opened 2026-08-09
rpc: expose effective fee asset

Abdullah1738:
## Motivation

Elements accepts `-feeasset=<asset-id>` and uses the resulting `policyAsset` for transaction-fee and relay policy, while `getsidechaininfo` currently exposes only the consensus `pegged_asset`. An RPC client connected to an already-running node therefore cannot determine whether the operator selected a different effective fee asset.

## Change

- add an always-present `fee_asset` field to `getsidechaininfo`
- define it as the effective `policyAsset` after startup argument processing
- keep the existing `pegged_asset` field unchanged
- add a wallet-independent functional test covering both the default and an explicit override

The change is additive and read-only. `policyAsset` is fixed before RPC service starts, so it does not introduce new mutable state or locking.

## Local verification

- `rpc_sidechaininfo.py` passed directly and through the functional test runner
- `rpc_tweakfedpeg.py --legacy-wallet` passed
- `feature_dynafed.py --legacy-wallet` passed
- touched-file, include, Python UTF-8, functional-test-list, mypy, and dead-code checks passed

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -32,6 +32,7 @@
 #include <node/transaction.h>
 #include <node/utxo_snapshot.h>
 #include <node/warnings.h>
+#include <policy/policy.h>
 #include <primitives/transaction.h>
 #include <rpc/server.h>
 #include <rpc/server_util.h>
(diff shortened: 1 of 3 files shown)


delta1 (approved):
ACK 13d830f59fa61b9e339f374004f9d11cfdaa8261; tested locally

tomt1664:
tACK https://github.com/ElementsProject/elements/commit/13d830f59fa61b9e339f374004f9d11cfdaa8261

PR #1580, opened 2026-08-12
build: force the surjectionproof CMake option

Abdullah1738:
The Elements wrapper currently forces `SECP256K1_ENABLE_MODULE_SURJECTION`, but the vendored secp256k1 subtree defines `SECP256K1_ENABLE_MODULE_SURJECTIONPROOF`. This leaves surjection proof support enabled only by the dependency default and lets an external cache value disable it.


tomt1664:
ACK 68affce5cc978655a9dbaaa5ed72ce6b032bf77d tested locally.

tomt1664 (approved):

PR #1581, opened 2026-08-12
Introduce per-txin sighash midstate cache for legacy/p2sh/segwitv0 scripts

tomt1664:
Merge of https://github.com/bitcoin/bitcoin/pull/32473 with additional cache index bit added for sighash combination with SIGHASH_RANGEPROOF.


apoelstra:
utACK 0a0dd9b069771a2706dbe9a2e1df1434f4b8f157

FWIW I think you could change the size of the `cache_entries` table to 14. Probably not worth the extra thinking and risk of OOB access to save a couple entries.

src/script/interpreter.cpp
```diff
@@ -2802,11 +2802,61 @@ bool SignatureHashSchnorr(uint256& hash_out, ScriptExecutionData& execdata, cons
     return true;
 }
 
+int SigHashCache::CacheIndex(int32_t hash_type) const noexcept
+{
+    // Note that we do not distinguish between BASE and WITNESS_V0 to determine the cache index,
+    // because no input can simultaneously use both.
+    // ELEMENTS: SIGHASH_RANGEPROOF changes the preimage (segwit v0 appends hashRangeproofs; the
+    // legacy serializer appends each output's rangeproof and surjectionproof), so it must be a
+    // dimension of the cache key. Removing this term is a consensus split with elements-23.x.
```
apoelstra: In 903be1e538ae16beff79ba15081a84362c3dc5c7:

This comment is a bit confusing. Removing this term would break the caching logic, but it wouldn't be a "consensus split with elements 23" in particular. It'd just be wrong.

src/script/interpreter.h
```diff
@@ -289,8 +289,30 @@ extern const HashWriter HASHER_TAPSIGHASH_ELEMENTS; //!< Hasher with tag "TapSig
...
 
+/** Data structure to cache SHA256 midstates for the ECDSA sighash calculations
+ *  (bare, P2SH, P2WPKH, P2WSH). */
+class SigHashCache
+{
+    /** For each sighash mode (ALL, SINGLE, NONE, ALL|ANYONE, SINGLE|ANYONE, NONE|ANYONE, ALL|RANGEPROOF, SINGLE|RANGEPROOF, NONE|RANGEPROOF, ALL|ANYONE|RANGEPROOF, SINGLE|ANYONE|RANGEPROOF, NONE|ANYONE|RANGEPROOF),
+     *  optionally store a scriptCode which the hash is for, plus a midstate for the SHA256
+     *  computation just before adding the hash_type itself. */
+    // ELEMENTS: the SIGHASH_RANGEPROOF (0x40) bit changes the sighash preimage, so it is part of
+    // the cache key and the table has 16 entries rather than upstream's 6. Do not drop this when
```
apoelstra, line 309: In 903be1e538ae16beff79ba15081a84362c3dc5c7:

Fine to leave as-is, but the old table had 6 elements and we're adding 8 to it, so 14 is the correct number.

tomt1664: Yes, you are right. I will leave as is though, as it is (again incorrectly) 16 in 23.x.

delta1: Isn't it 12? 

3 (sighash types) x 2 (anyonecanpay) x 2 (rangeproof)

test/functional/feature_taproot.py
```diff
@@ -1240,6 +1264,70 @@ def predict_sigops_ratio(n, dummy_size):
...
+                        inputs.append(getter("sign", codesepnum=codeseps, hashtype=op))
+                inputs.reverse()
+                script = CScript(script)
+                add_spender(spenders, "sighashcache/legacy", p2sh=p2sh, witv0=witv0, standard=False, script=script, inputs=inputs, key=eckey1, sigops_weight=12*8*(4-3*witv0), no_fail=True)
+
+    # Sighash caching in tapscript.
+    for _ in range(10):
+        # Construct a script with 700 checksig operations (7 sighash types, each 100 times),
+        # randomly ordered and interleaved with 100 OP_CODESEPARATORS.
+        ops = [0, 1, 2, 3, 0x81, 0x82, 0x83, -1] * 100
```
delta1, line 1303: 1484750d83adfd3cd52667013cc77f277821fa7f test passes even without the code changes in 903be1e538ae16beff79ba15081a84362c3dc5c7, because we don't have any rangeproof sighashes here. 

I tried adding one (0x41), but then it fails even without our code changes in 903be1e538ae16beff79ba15081a84362c3dc5c7, because SIGHASH_RANGEPROOF requires dynafed. 

Enabling Dynafed in this test then starts failing in `gen_test_vectors` with `invalid-dyna-fed` since we're not constructing dynafed blocks.

delta1: I don't know if it's going to be easier (and/or less merge conflicts) to make this test work with dynafed, or to just design a new unit or functional test to make sure this works as required.

src/script/interpreter.h
tomt1664: The index is 8 x rp + 3 x acp + 2 x single + 1 x none. The low part 3 x acp + 2 x single + 1 x none spans 0–5, and the rangeproof term adds 8, so the reachable indices are:

0 1 2 3 4 5        (0x40 clear)
8 9 10 11 12 13    (0x40 set)

Twelve reachable values, but a maximum of 13 — so it must be minimum 14. 6 and 7 are unreachable.

src/script/interpreter.cpp
tomt1664: Updated comment.

test/functional/feature_taproot.py
tomt1664: `prepare_mixed_sighash_multisig_tx` test added to `feature_sighash_rangeproof.py ` to verify the additional bit. Have verified that this test fails without https://github.com/ElementsProject/elements/commit/903be1e538ae16beff79ba15081a84362c3dc5c7

delta1 (approved):
ACK 9631bd33eb38d7bf66968571daf6b1d1ac7a33cf; tested locally

PR #1584, opened 2026-08-19
script: add and default to SIGHASH_ALL_WITH_RANGEPROOF for pre-taproot signing 

delta1:
Pre-Taproot signatures using the historical SIGHASH_ALL default do not commit to output rangeproofs, leaving a witness malleability gap: an attacker can alter a transaction's rangeproofs without invalidating its signatures. This branch closes that gap by making signing default to SIGHASH_ALL | SIGHASH_RANGEPROOF on chains where dynafed is active, while leaving explicit user-supplied sighash types untouched and preserving the legacy default.

Scope:

- script: adds SIGHASH_ALL_WITH_RANGEPROOF and a DefaultSighashType() helper; strips the 0x40 bit for Taproot/Schnorr signing so the constant is a valid universal default.
- node: exposes Chain::isSighashRangeproofActive() for a live tip-based dynafed check.
- chainparams: adds SighashRangeproofActiveByParams() for chainstate-less gating (used by bitcoin-tx).
- wallet + raw RPCs (signrawtransactionwithkey, signrawtransactionwithwallet, walletprocesspsbt, descriptorprocesspsbt) + - bitcoin-tx: default to the rangeproof-committing sighash when dynafed is active.
- Adds unit and functional test coverage plus a release note.

### DIFF
--- a/src/test/validation_tests.cpp
+++ b/src/test/validation_tests.cpp
@@ -2,6 +2,7 @@
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
+#include <bitcoin-build-config.h> // IWYU pragma: keep
 #include <chainparams.h>
 #include <consensus/amount.h>
 #include <consensus/merkle.h>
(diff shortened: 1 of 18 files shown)


apoelstra (approved):
utACK bbb1f5a540cf8480d36af2a7d27030d35d4ca526

src/wallet/wallet.h
```diff
@@ -702,6 +702,7 @@ class CWallet final : public WalletStorage, public interfaces::Chain::Notificati
     OutputType TransactionChangeType(const std::optional<OutputType>& change_type, const std::vector<CRecipient>& vecSend) const;
 
     /** Fetch the inputs and sign with SIGHASH_ALL. */
+    // ELEMENTS: sign with SIGHASH_ALL_WITH_RANGEPROOF for DynaFed chains
```
tomt1664, line 705: Minor nit: this comment seems to contradict the one above

tomt1664 (approved):

src/test/validation_tests.cpp
```diff
@@ -364,4 +364,28 @@ BOOST_AUTO_TEST_CASE(block_malleation)
     }
 }
```
tomt1664, line 367: `#include <bitcoin-build-config.h> // IWYU pragma: keep` at top for linter

tomt1664:
`feature_fedpeg.py --post_transition --legacy-wallet` and `feature_fedpeg.py --pre_transition --legacy-wallet` failing (verified locally). 

signed_pegin comes back complete: True, but fin_psbt comes back complete: False

src/wallet/wallet.h
delta1: Thanks, I thought to leave the original comment and just add an Elements specific one. Less merge conflicts, and the Elements line is only true for DynaFed chains.

tomt1664: Yes, better for future merges.

src/test/validation_tests.cpp
delta1: Thanks!

delta1:
Thanks @tomt1664 , fixed the issue in psbt.cpp

tomt1664:
ACK feb50a3f09fc7a82d0522afde262e48b86dd75d6 tested locally

PR #1590, opened 2026-08-27
Fix interface_zmq.py functional tests

tomt1664:
For the test-each-commit CI job:

Broaden the CMake guard from if(APPLE) to also match Clang on any platform, e.g. if(CMAKE_C_COMPILER_ID MATCHES "Clang") (this covers both Clang and AppleClang) for the elementssimplicity compilation. 

Fix is_valid() in test framework for elements amounts, enabling interface_zmq.py to pass.


delta1 (approved):
ACK 665d873883192dacd67118580ad39c1c2a463c68; tested locally

PR #1593, opened 2026-09-01
Fix RPC return errors for psbt and invalid rangeproofs

tomt1664:
Fixes for a number of issues with RPC errors for invalid PSBTs and amounts/rangeproofs.

### DIFF
--- a/src/blind.cpp
+++ b/src/blind.cpp
@@ -546,7 +546,9 @@ int BlindTransaction(std::vector<uint256 >& input_value_blinding_factors, const 
 
                 // Generate rangeproof, no script committed for issuances
                 bool rangeresult = GenerateRangeproof((nPseudo ? txinwit.vchInflationKeysRangeproof : txinwit.vchIssuanceAmountRangeproof), value_blindptrs, nonce, amount, CScript(), value_commit, asset_gen, asset, asset_blindptrs);
-                assert(rangeresult);
+                if (!rangeresult) {
+                    return -1;
+                }
 
                 // Successfully blinded this issuance
                 num_blinded++;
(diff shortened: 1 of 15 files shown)


src/wallet/wallet.cpp
```diff
@@ -2306,7 +2306,7 @@ bool CWallet::SignTransaction(CMutableTransaction& tx) const
         coins[input.prevout] = Coin(wtx.tx->vout[input.prevout.n], prev_height, wtx.IsCoinBase());
     }
     std::map<int, bilingual_str> input_errors;
-    return SignTransaction(tx, coins, DefaultSighashType(chain().isSighashRangeproofActive()), input_errors); // ELEMENTS
+    return SignTransaction(tx, coins, SIGHASH_DEFAULT, input_errors);
```
delta1: Intentional? This is a regression of #1584

delta1 (commented):
LGTM just 1 comment

src/wallet/wallet.cpp
tomt1664: No, that's my mistake. Have reverted.

src/blind.cpp
```diff
@@ -621,7 +623,9 @@ int BlindTransaction(std::vector<uint256 >& input_value_blinding_factors, const
 
             // Generate rangeproof
             bool rangeresult = GenerateRangeproof(txoutwit.vchRangeproof, value_blindptrs, nonce, amount, out.scriptPubKey, value_commit, asset_gen, asset, asset_blindptrs);
-            assert(rangeresult);
+            if (!rangeresult) {
+                continue;
+            }
```
delta1: should these continues rather return an error with RANGEPROOF_UNABLE ?

tomt1664: This was following the same logic as the `SurjectOutput` check on the line below (631) that continues if it fails (so output is left unblinded and the caller's ignoreblindfail logic will decide what happens). 

If this is the incorrect then I can change both.

tomt1664: Changed to return -1 for this and SurjectOutput failure.

delta1: Sorry I guess the new unit test relied on this behaviour?

delta1 (approved):
ACK a9db3b1c1fce0288f47dac21915f9fceaaeb231a; tested locally

src/blind.cpp
tomt1664: return -1 for the SurjectOutput failure caused existing naive_blinding_test to fail. So I think I was wrong about the SurjectOutput failure continue being a bug. It can mean no suitable input asset to prove against. 

I have reverted this and kept the return -1 for the GenerateRangeproof failure.

PR #1594, opened 2026-09-02
secp: context randomize in a few places 

delta1:
Adds a few instances of `secp256k1_context_randomize` in areas where we are using the secp context


apoelstra:
utACK. Only reviewed the total diff in the Github UI. It's a worthwhile change IMHO. But I assume you weighed this defense-in-depth against causing potential rebase conflicts.

delta1:
Dropped the commit adding aux_rand to schnorr sign since upstream does that intentionally. Also fixed lint CI by not asserting in RPC code

src/pegins.cpp
```diff
@@ -37,6 +38,10 @@ class Secp256k1Ctx
         assert(secp256k1_ctx_validation == nullptr);
         secp256k1_ctx_validation = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY | SECP256K1_CONTEXT_SIGN);
```
apoelstra: In 9cf2c2f54315dcb72071eb9ce19ef0ded250834e:

This is called `ctx_validation` but we create it with the `CONTEXT_SIGN` flag. We don't ever use that flag -- we only use this context on public data. So (a) we should drop the flag, and (b) we don't need to randomize this one. Although it's harmless to do so.

Also upstream has deprecated these flags for a while. See `src/secp256k1/include/secp256k1.h` which advises us to use `SECP256K1_CONTEXT_NONE` everywhere.

src/CMakeLists.txt
```diff
@@ -31,7 +31,18 @@ set(SECP256K1_ENABLE_MODULE_WHITELIST ON CACHE BOOL "" FORCE)
 set(SECP256K1_DISABLE_SHARED ON CACHE BOOL "" FORCE)
 set(SECP256K1_ENABLE_MODULE_ECDH ON CACHE BOOL "" FORCE)
 set(SECP256K1_ENABLE_MODULE_RECOVERY ON CACHE BOOL "" FORCE)
+# Compile out the unused experimental secp256k1-zkp modules. These four default
+# to ON in libsecp256k1's CMake and have zero references in the Elements tree.
+# Building them is dead code: a larger binary and a wider attack surface with
+# no functional or consensus benefit.
+# The modules genuinely used by Elements (ECDH, RECOVERY, EXTRAKEYS,
+# SCHNORRSIG, ELLSWIFT, GENERATOR, RANGEPROOF, SURJECTIONPROOF, WHITELIST)
```
apoelstra: In 97ca39d99918fdb073e275630e617bb56b29c463:

This whole comment has llm tone and should be dropped. Or at least shortened to `# Disable unused on-by-default features from secp256k1-zkp`.

apoelstra:
utACK 0b87696445 -- though I guess the transition to `SECP256K1_CONTEXT_NONE` ought to happen everywhere. I guess upstream will do it and we can pull it in.

delta1:
Apologies for the force push, it was necessary to fix the failing windows CI job. 

It changed `GetRandBytes` for `GetStrongRandBytes`, which is the correct function to use anyway

PR #1600, opened 2026-09-08
sigcache: harden range proof cache keys and add -norangeproofcache option

delta1:
Target branch is elements-23.3.x

Switch the range-proof and surjection-proof cache hashers from raw
CSHA256 concatenation to CHashWriter . This length-prefixes
every field, so distinct argument tuples with byte-identical raw
concatenations no longer collide to the same cache key.

Both caches retain per-process salted midstates with distinct domain
separators ('r' / 's').

Also adds a -norangeproofcache startup option to
disable the range proof cache without recompiling.

Unit tests cover field-boundary collisions, script sensitivity, domain
separation, determinism, and vTags sensitivity. 

Bump version to 23.3.4 and updates manpages.


delta1:
ARM CI task failure is because Debian 11 is EOL 

https://www.debian.org/News/2026/20260831

l0rinc:
A fuzz oracle to test the combinations with and without cache could help catching these problems next time.

src/script/sigcache.cpp
```diff
@@ -72,13 +71,39 @@ class CSignatureCache
...
     }
-    void ComputeEntrySurjectionProof(uint256& entry, const uint256 &hash, const std::vector<unsigned char>& proof, const std::vector<unsigned char>& commitment) {
-        CSHA256 hasher = m_salted_hasher_surjection_proof;
-        hasher.Write(hash.begin(), 32).Write(proof.data(), proof.size()).Write(commitment.data(), commitment.size()).Finalize(entry.begin());
+    void ComputeEntrySurjectionProof(uint256& entry, const uint256 &hash, const std::vector<unsigned char>& proof, const std::vector<unsigned char>& commitment, const std::vector<secp256k1_generator>& vTags) const {
+        CHashWriter hasher = m_salted_hasher_surjection_proof;
+        // We hash all arguments passed to CachingSurjectionProofChecker::VerifySurjectionProof,
+        // to ensure that any change in the way that the verification function is called will
+        // trigger a cache miss and explicit verification. However, we note that the `wtxid`
+        // (hash) commits to all the other data such that we could technically hash only it.
```
apoelstra, line 96: In 94000967f6dc05b1afd435e79b1bbc597e29f816:

I realize it isn't true that just the wtxid and proof are sufficient -- you can imagine the same proof appearing multiple times in the same transaction, such that it's only valid in one place.

Our code correctly avoids this vulnerability by hashing the commitment and tags, it's just the comment saying "y'know..." that's wrong.

src/script/sigcache.cpp
```diff
@@ -72,13 +71,39 @@ class CSignatureCache
...
+        // We retain the other data as a defense against future refactorings.
+        //
+        // Serialize vTags as a flat byte vector (each secp256k1_generator is 64 bytes).
+        std::vector<unsigned char> vTagsBytes;
+        vTagsBytes.reserve(vTags.size() * 64);
+        for (const auto& tag : vTags) {
+            vTagsBytes.insert(vTagsBytes.end(), std::begin(tag.data), std::end(tag.data));
```
apoelstra, line 103: In 94000967f6dc05b1afd435e79b1bbc597e29f816:

Just highlighting that this use of the internal `tag.data` array is fine, although the libsecp256k1-zkp docs say that the internals of `secp256k1_generator` are implementation-defined and nonportable. We do not need portability here or any particular property of the bytes, only that they represent the generator in question.

src/script/sigcache.cpp
```diff
@@ -179,6 +179,10 @@ bool CachingTransactionSignatureChecker::VerifySchnorrSignature(Span<const unsig
 // To be called once in AppInit2/TestingSetup to initialize the rangeproof cache
 void InitRangeproofCache()
 {
+    if (!gArgs.GetBoolArg("-rangeproofcache", true)) {
```
apoelstra, line 182: In 19d704279b3197af16ac52d910a915225505fe61:

I think we should also disable the surjectionproof cache, throughout. Fine to leave the option name as-is.

apoelstra:
Done reviewing 1bfa1eeb45159ce481926febc12c5fd2888dd1a0. Just two nits.

psgreco:
ACK 1bfa1eeb45159ce481926febc12c5fd2888dd1a0

PR #1602, opened 2026-09-08
CI: disable apt valid-until check for bullseye as EOL

tomt1664:
Debian 11 (bullseye) has reached end of LTS and has been pulled from the live mirrors: deb.debian.org 404s on the bullseye pool. 

The images carry commented-out snapshot.debian.org lines pinned to the timestamp the image was built from i.e. the exact repository state the installed packages came from, security suite included. This PR enables those and drops the dead mirrors.


delta1 (approved):
ACK ca8c8d7afb2ff12e16a2367fb5a4ff99b138aea2; all CI passed

PR #1603, opened 2026-09-10
sigcache: prevent rangeproof cache key boundary collisions

HarshRajSinghania:
## Summary

`ComputeEntryRangeProof` on `master` hashes `proof || commitment || asset_commitment || scriptPubKey` without field delimiters. Because `proof` and `scriptPubKey` are variable-length fields at opposite ends of the stream, distinct argument tuples can produce the same hashed byte sequence. A positive cache hit bypasses rangeproof verification.

This PR length-prefixes the four fields before hashing and adds a regression test for the proof/script boundary case. The test primes the cache with a valid proof and then verifies that a re-split, invalid tuple does not hit the same cache entry.

The current `master` implementation was verified at `c7e856fab1b0c4d37005e25c0940184d812a26a0` before this change.

This is the master-branch counterpart of the already-public cache-key hardening work in #1600/#1601; those changes landed on `elements-23.3.x`, while this report specifically identified the same raw-concatenation construction on `master`.

## Testing

- Added `rangeproof_cache_key_field_boundary` regression coverage in `src/test/blind_tests.cpp`.
- The regression is designed to fail on the vulnerable raw-concatenation implementation and pass with length-prefixed keying.


tomt1664:
Thanks - already been resolved in https://github.com/ElementsProject/elements/pull/1600

PR #1604, opened 2026-09-14
[master] sigcache: harden range proof cache keys and add -norangeproofcache option

delta1:
target is **master** branch, cherry-picked commits from #1600 

Switch the range-proof and surjection-proof cache hashers from raw
CSHA256 concatenation to CHashWriter . This length-prefixes
every field, so distinct argument tuples with byte-identical raw
concatenations no longer collide to the same cache key.

Both caches retain per-process salted midstates with distinct domain
separators ('r' / 's').

Also adds a -norangeproofcache startup option to
disable the range proof cache without recompiling.


tomt1664:
ACK 53167b1f452f9cdf96b1ec4af7fed613ba772c95 tested locally

PR #1605, opened 2026-09-16
CI: upgrade debian:bullseye to bookworm for ARM

tomt1664:
Upgrade to debian:bookworm and support for debian:bullseye depreciated. 

GUI tests disabled:
The ARM job cross-compiles for armhf and runs the unit tests under qemu-arm. Its purpose is verifying that the tree still builds and that core/consensus tests pass on a 32-bit target — not validating GUI behaviour. test_elements-qt runs headless under emulation against the minimal platform plugin, and on bookworm it crashes in AppTests::initTestCase() before executing any test body, for the same qemu/glibc thread-initialisation reason that affects the other suites. Qt is built and tested natively on x86_64, where the GUI is actually exercised.


ci/test/04_install.sh
```diff
@@ -120,12 +110,12 @@ if [[ ${USE_MEMORY_SANITIZER} == "true" ]]; then
...
   CI_EXEC "cd ${BASE_SCRATCH_DIR}/msan/build/ && make $MAKEJOBS cxx"
 fi
-
+ 
 if [ -z "$DANGER_RUN_CI_ON_HOST" ]; then
   echo "Create $BASE_ROOT_DIR"
   CI_EXEC rsync -a /ro_base/ "$BASE_ROOT_DIR"
 fi
-
+
```
delta1: Please revert these whitespace changes, they seem unnecessary

delta1 (approved):
ACK 482b0722619633cf70ce8436794a6c69562f2e9d; CI passing

